Issue #277, opened 2012-12-07
Create multiple maps of same type based off parameter [FeatureRequest]


CJM3407:
Is there a way to call CreateMap on multiple times on the same mapping, but make it choose the correct mapping definition based on a parameter.  For example, I am mapping ObjectA to ObjectB.  I want to set a property on ObjectB to the string concatenation of two properties on ObjectA.  But I also in a different part of code want map ObjectA to ObjectB where the same property is only based on one of ObjectA's properties.  

This question stems from a feature in Unity (dependency injection library).  It allows you to name your registration with a string.  And maybe this feature exists in AutoMapper, and I am not looking in the correct location.

PR #283, opened 2013-01-04
Added name mappings support

MichaelUlmann:
Resolves issue #277

This extension allows:
Mapper.CreateMap<Source, Destination>("mapping1")...;
Mapper.CreateMap<Source, Destination>("mapping2")...;

and resolve them:
Mapper.Map<Source, Destination>("mapping1");


jbogard:
So you can take care of this using 4.2.0, that introduces removing of the static interface to instance-based. Now you can partition configuration as much as you want.

PR #305, opened 2013-02-21
Option AllowDuplicateCallsPerMapping

dariusdamalakas:
New option AllowDuplicateCallsPerMapping to track whether a typemap pair has already been registered or not.

### DIFF
--- a/src/AutoMapper/AutoMapper.csproj
+++ b/src/AutoMapper/AutoMapper.csproj
@@ -92,6 +92,7 @@
     <Compile Include="..\CommonAssemblyInfo.cs">
       <Link>Properties\CommonAssemblyInfo.cs</Link>
     </Compile>
+    <Compile Include="DuplicateCallsPerMappingException.cs" />
     <Compile Include="Internal\AliasedMember.cs" />
     <Compile Include="Mappers\ExplicitConversionOperatorMapper.cs" />
     <Compile Include="Mappers\ImplicitConversionOperatorMapper.cs" />
--- a/src/AutoMapper/ConfigurationStore.cs
+++ b/src/AutoMapper/ConfigurationStore.cs
@@ -34,6 +34,7 @@ namespace AutoMapper
 		    _typeMapFactory = typeMapFactory;
 		    _mappers = mappers;
             _globalIgnore = new List<string>();
+            AllowDuplicateCallsPerMapping = true;
 		}
 
 	    public event EventHandler<TypeMapCreatedEventArgs> TypeMapCreated;
@@ -42,6 +43,8 @@ namespace AutoMapper
 	    {
 	        get { return _serviceCtor; }
 	    }
+
+        public bool AllowDuplicateCallsPerMapping { get; set; }
 
 	    public bool AllowNullDestinationValues
 		{
@@ -241,7 +244,12 @@ namespace AutoMapper
 		public TypeMap CreateTypeMap(Type source, Type destination, string profileName, MemberList memberList)
 		{
 			TypeMap typeMap = FindExplicitlyDefinedTypeMap(source, destination);
-				
+
+            var profileConfiguration2 = GetProfile(profileName);
+			if (this.AllowDuplicateCallsPerMapping == false)
+                if (typeMap != null)
+                    throw new DuplicateCallsPerMappingException(string.Format("Mapping was already define before! Mapping from {0} to {1}, profile {2}", 
+                        source.FullName, destination.FullName, profileName));
 			if (typeMap == null)
 			{
 			    var profileConfiguration = GetProfile(profileName);
--- a/src/AutoMapper/ConfigurationStore.cs
+++ b/src/AutoMapper/ConfigurationStore.cs
@@ -15,6 +15,7 @@ using AutoMapper.Mappers;
 
 namespace AutoMapper
 {
+
 	public class ConfigurationStore : IConfigurationProvider, IConfiguration
 	{
 	    private readonly ITypeMapFactory _typeMapFactory;
@@ -26,6 +27,12 @@ namespace AutoMapper
         private readonly ConcurrentDictionary<TypePair, TypeMap> _typeMapCache = new ConcurrentDictionary<TypePair, TypeMap>();
         private readonly ConcurrentDictionary<string, FormatterExpression> _formatterProfiles = new ConcurrentDictionary<string, FormatterExpression>();
 		private Func<Type, object> _serviceCtor = ObjectCreator.CreateObject;
+        /// <summary>
+        /// Stores stack traces where a type map where registered. This is in use only when AllowDuplicateCallsPerMapping
+        /// setting is set to true. This will output previous typeMap registration location, plus current one (which 
+        /// will be evident from exceptions stack trace)
+        /// </summary>
+        private readonly  ConcurrentDictionary<TypeMap, string> _typeMapRegistrationLocations = new ConcurrentDictionary<TypeMap, string>();
 
 	    private readonly List<string> _globalIgnore;
 
@@ -245,11 +252,11 @@ namespace AutoMapper
 		{
 			TypeMap typeMap = FindExplicitlyDefinedTypeMap(source, destination);
 
-            var profileConfiguration2 = GetProfile(profileName);
 			if (this.AllowDuplicateCallsPerMapping == false)
                 if (typeMap != null)
-                    throw new DuplicateCallsPerMappingException(string.Format("Mapping was already define before! Mapping from {0} to {1}, profile {2}", 
-                        source.FullName, destination.FullName, profileName));
+                    throw new DuplicateCallsPerMappingException(string.Format("Mapping was already define before! Mapping from type '{0}' to '{1}'. " +
+                                                                              "Original type map was registered here {2}", 
+                        source.FullName, destination.FullName, _typeMapRegistrationLocations[typeMap]));
 			if (typeMap == null)
 			{
 			    var profileConfiguration = GetProfile(profileName);
@@ -258,6 +265,7 @@ namespace AutoMapper
 
                 typeMap.Profile = profileName;
 			    typeMap.IgnorePropertiesStartingWith = _globalIgnore;
+                
 
 			    IncludeBaseMappings(source, destination, typeMap);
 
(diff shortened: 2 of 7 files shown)


fduman:
We are also having trouble managing mappings in a large project. We are facing really annoying mapping bugs. We have 5 people in our team. Unfortunately, nobody searches whether or not there are the same type mappings. Developers can write the same type mapping in different ways. Ok, some developers may not understand the Automapper usage well, but I don't know what can I do.

I have to inherit the Automapper's Profile class and override new CreateMap functions which check this situation in our profile classes.

@jbogard  I wonder why you didn't add this option? What is your opinion about this matter? Are we using the Automapper in a wrong way (of course we are wrong but we are some miserable developers :) ) or what?

dariusdamalakas:
@fduman , I can rebase my branch to resolve conflicts if this is OK to be accepted. Unfortunately, this is being ignored.

jbogard:
@dariusdamalakas Are you sure this is still an issue with the 4.0/4.1 packages?

fduman:
@jbogard We are still using 3.3.1 in our production environment. I don't think we can replace soon to the Automapper's 4.0+ package. 

I wonder what is the current behavior of the Automapper?

fduman:
@dariusdamalakas we really appreciate, if you can apply changes for 3.3.1

dariusdamalakas:
@fduman , actually, if 4.0 solves this issue, I am happy to drop this PR.    @jbogard, sorry, I have glanced over [release notes](https://github.com/AutoMapper/AutoMapper/releases/tag/v4.0.0), and could not find whether any of the features/bug fixes do address this issue. What is the standard behaviour then when adding duplicate mapping?

jbogard:
@dariusdamalakas What's supposed to happen when you call CreateMap twice is that the first one creates it, the second one uses the existing map, nothing gets overwritten.

dariusdamalakas:
ok, this does solve the issue by ignoring the second call.  This is not ideal, as this would still lead to heavy head-scratch on massive projects. Take this as an example:

````
# file A in assembly A
AutoMapper.Mapper.CreateMap<Book, BookViewModel>()
    .ForMember(dest => dest.Author, opts => opts.MapFrom(src => throw null));

# file B in assembly B
AutoMapper.Mapper.CreateMap<Book, BookViewModel>()
    .ForMember(dest => dest.Author, opts => opts.MapFrom(src => src.Author.Name));

````

Now, depending on the order how the assemblies and auto mapper initialisation modules are scanned and run, this will either work fine, or will blow up with `NullReferenceException` when mapping is used. The proposal (as suggested in this PR) is to throw an exception when a duplicate mapping is about to be added. This way the issue will be spotted immediately after a second mapping is added and AutoMapper is initialised.

jbogard:
I think as long as it defaults to not throw, it should be OK.

fduman:
@dariusdamalakas I think your first solution is better. 

We use Automapper profiles. I  inherit our profile classes from MappingProfile which is something like this:

    public abstract class MappingProfile : AutoMapper.Profile
    {
        public static readonly List<string> MappingErrors = new List<string>();
        protected MappingProfile()
        {

        }

        public new IMappingExpression CreateMap(Type sourceType, Type destinationType)
        {
            CheckIfAlreadyMapped(sourceType, destinationType);
            return base.CreateMap(sourceType, destinationType);
        }

        public new IMappingExpression CreateMap(Type sourceType, Type destinationType, MemberList memberList)
        {
            CheckIfAlreadyMapped(sourceType, destinationType);
            return base.CreateMap(sourceType, destinationType, memberList);
        }

        public new IMappingExpression<TSource, TDestination> CreateMap<TSource, TDestination>()
        {
            CheckIfAlreadyMapped<TSource, TDestination>();
            return base.CreateMap<TSource, TDestination>();
        }

        public new IMappingExpression<TSource, TDestination> CreateMap<TSource, TDestination>(MemberList memberList)
        {
            CheckIfAlreadyMapped<TSource, TDestination>();
            return base.CreateMap<TSource, TDestination>(memberList);
        }

        [Conditional("DEBUG")]
        private void CheckIfAlreadyMapped(Type sourceType, Type destinationType)
        {
            if (Mapper.FindTypeMapFor(sourceType, destinationType) != null)
            {
                MappingErrors.Add(string.Format("Type is already mapped in profile:{0} for source:{1} destination {2}", this.GetType().FullName, sourceType.FullName, destinationType.FullName));
                return;
                //throw new InvalidOperationException(string.Format("Type is already mapped in profile:{0} for source:{1} destination {2}", this.GetType().FullName, sourceType.FullName, destinationType.FullName));
            }
        }

        [Conditional("DEBUG")]
        private void CheckIfAlreadyMapped<TSource, TDestination>()
        {
            if (Mapper.FindTypeMapFor<TSource, TDestination>() != null)
            {
                MappingErrors.Add(string.Format("Type is already mapped in profile:{0} for source:{1} destination {2}", this.GetType().FullName, typeof(TSource).FullName, typeof(TDestination).FullName));
                return;
                //throw new InvalidOperationException(string.Format("Type is already mapped in profile:{0} for source:{1} destination {2}", this.GetType().FullName, typeof(TSource).FullName, typeof(TDestination).FullName));
            }
        }
    }

So Profile classes use this CreateMap methods. I can validate in Application_Start if double mapping issue happens.

@jbogard I don't know why didn't you answer me directly?  Also, I think you couldn't understand what we mean.  It's not important to us that first call creates mappings and seconds do nothing. Important thing is that we don't know which one is called first. Maybe wrong one called first and destroys all system. So, there must be a general rule which is "there should be always one CreateMap call for two different types". Am I correct?

Regards.

jbogard:
@fduman Not really correct, the desired behavior that's been in AutoMapper for a while is if you have 2 CreateMap, the first one creates it and the second one builds on top of it. It doesn't replace it unless you explicitly wire in conflicting mapping configuration.

So in order to not break existing clients, we'd need the default behavior to simply build on top of maps, not replace them.

seanfisher:
@jbogard I think the proposal and the PR was for an opt-in duplicate-checking behavior, not to change the default behavior. 

Let me throw my vote in that this would be an incredibly time-saving feature. We have something like 15 mapper profiles across shared PCLs, Xamarin projects, and web API projects. I just spent a long time tracking down a bug (not the first, and probably not the last) where a developer added a mapping from <Model, Dto> in profile B when the mapping already existed in profile A, and it led to unexpected behavior.

Why shouldn't the developer have just searched for an existing mapping before adding a new one? It's incredibly difficult to accurately search for mappings across profiles to see if `<Model, Dto>` has already been registered, because you're specifying types, which can be namespace-qualified or not. So in this case you could search for `<Model, Dto>` and not find anything because profile A is registered as `<My.Namespace.Model, Dto>` or `<My.Namespace.Model, Namespace.Dto>` or whatever. And combing through 500 references of `Model` to find the one that's used in the mapper profile is time-consuming and most developers don't do it.

Not to mention that once you add in a duplicate mapping the only way you know which mapping is "first" and which is "second" is by examining the profile registrations, which could exist in a completely different assembly, or in our case, in multiple assemblies (one in ASP.NET startup, one in iOS startup, etc). We had a really confusing situation where a mapping was incorrect, and we changed the mapping to be correct but for some reason when running the application it was still doing it incorrectly. After lots of cleaning, rebuilding, pulling our hair out we realized that, oh, we have _two_ mappings, we had changed the second (which wasn't actually influencing anything at all) and the original one was incorrect.

Managing all of that becomes a real pain and source of unexpected behavior. This is an *opt-in* behavior that would really help us out (and others, it sounds like).

@dariusdamalakas @jbogard Could we re-open this?

dariusdamalakas:
@seanfisher what a pain you must have gone through.  We have stopped using automapper as the de-facto standard for writing mapping code, and we are still sitting on old automapper (v3?), i reckon this should be re-evaluated whether this PR still fits.

lbargaoanu:
As a workaround you cam detect that yourself:

		var configuration = (IConfiguration) cfg;
		(
			from profile in configuration.Profiles
			from typeMap in profile.TypeMapConfigs
			from otherProfile in configuration.Profiles 
			from otherTypeMap in otherProfile.TypeMapConfigs 
			where otherProfile != profile && otherTypeMap.Types == typeMap.Types
			select typeMap
		).Dump();

seanfisher:
Thanks @lbargaoanu! Thanks to your code I was able to write a unit test that checks for duplicates. I altered it from your nice compact LINQ statement to something more garish in order to get the data in a structure more conducive to the output that I wanted. I'll paste it here for others' reference, and any suggestions of fancy ways to make it better/more compact would be appreciated:

    [Test]
    public void Automapper_NoDuplicateMappings()
    {
        var configuration = MyConfig; // Get from wherever you configure Automapper
        var mappings = new Dictionary<TypePair, List<IProfileConfiguration>>();
        foreach (var profile in configuration.Profiles)
        {
            foreach (var typeMap in profile.TypeMapConfigs)
            {
                if (mappings.ContainsKey(typeMap.Types))
                {
                    mappings[typeMap.Types].Add(profile);
                }
                else
                {
                    mappings[typeMap.Types] = new List<IProfileConfiguration> { profile };
                }
            }
        }
        var duplicates = mappings.Where(a => a.Value.Count > 1).ToDictionary(a => a.Key, a => a.Value);
        Assert.IsEmpty(duplicates, BuildAutomapperDuplicateMappingMessage(duplicates));
    }

    public string BuildAutomapperDuplicateMappingMessage(Dictionary<TypePair, List<IProfileConfiguration>> duplicateMappings)
    {
        var sb = new StringBuilder();
        sb.AppendLine("There are duplicate mappings added to Automapper config.");
        foreach (var mapping in duplicateMappings)
        {
            sb.AppendFormat("\nMapping ({0} -> {1}) is found in profiles:\n", mapping.Key.SourceType, mapping.Key.DestinationType);
            foreach (var profile in mapping.Value)
            {
                sb.AppendFormat("\t{0}\n", profile.ProfileName);
            }
        }

        return sb.ToString();
    }

Example output:

    There are duplicate mappings added to Automapper config.
    Mapping (My.Namespace.Model -> My.Namespace.Dto) is found in profiles:
        My.Namespace.ProfileName
        My.Namespace.ProfileName // Displays twice if mapping is added twice to same file
        My.Namespace.OtherProfile

Running this unit test also showed me a couple other areas where some duplicate mappings had snuck in, so it's already been helpful.

lbargaoanu:
@jbogard Having a map appear in more profiles seems toxic, we don't need an option for that :)

jbogard:
It might *seem* toxic, but it'd be a breaking change that we've had more than one person ask to make sure that the current behavior is expected. That maps are additive, and don't replace existing config, and that you don't throw or anything like that.

However, if we can find the original issues that brought that situation up, there might be a better way to solve their problem.

jbogard:
Derp. this is the original issue.

lbargaoanu:
I think the people that spoke in this issue want the opposite. And while I see your point about backward compatibility, when the maps are in different profiles, I think that does more harm than good.

jbogard:
So the idea would be this is an opt-in feature? I could get behind that.

jbogard:
(obviously a bit of work to clean up this PR for merging)

lbargaoanu:
Maybe open an issue with a link to this. Starting from scratch seems easier.

PR #395, opened 2013-10-01
Automaticly resolve mappings

JordyLangen:
Hi,

I wrote this little code a while back for a project. It resolves mapping in an assembly using reflection. I thought it would be useful for other people as well.


jbogard:
I like the idea here, but I'm gonna go a slightly different route here.

PR #528, opened 2014-05-25
Add remove items to existing collection mappings

TylerCarlson1:
New mapper that will map IEnumerable to ICollection<> types, IFF both are not null and have equivalency expression assigned to them. Equivalency expressions will be used to Add/Modify/Delete objects within a collection.
Added to all mappers after read only collection mapper as to not attempt on read only collections.

Included Extension method for creating custom comparisons in the CreateMap<TSource,TDest> fluent API. (Unit test added in unit tests)
Ability to add custom equivalency expression conventions that will generate equivalencies based on source and dest type.
Added EntityFramework primary key example in Automapper.EntityFramework project.  Unit test in Integrations test project.
EntityFramwork's mapping are overwritten by custom mappings.

Set up so other ORMs can be used as long as you can ascertain the PK's of each object by their types.

I want to include further conventions on top of PKs like override key mappings, but this shows the default implementation that should work for most cases.


TylerCarlson1:
Moved implementation of base equality comparison to https://github.com/TylerCarlson1/Automapper.Collection.
I still support this being in AutoMapper base, but don't know how long it will be before this will come.

PR #583, opened 2014-09-03
Projecting Enum to Enum in another assembly throws an exception

nholling:


yyjdelete:
Also failed ProjectTo with for Enum -> byte
byte is the underType of the enum, `public enum ModelStatus : byte`
`ProjectUsing` is an workaround, but is there other way to do this?

lbargaoanu:
@yyjdelete Try the MyGet build.

lbargaoanu:
I don't think this is worth fixing. One can work around it by creating a map for the enums and casting in ProjectUsing (or whatever works).

yyjdelete:
@lbargaoanu 
Yes, you are right. But it's strange when using `v5.0.0-ci-1049`, the same Exception(maybe) is thrown without `ProjectUsing`, but an `NullReferenceException` is thrown when access ex.Message, seems another issue?
```
System.NullReferenceException: (locale message...)
   at AutoMapper.AutoMapperMappingException.get_Message()
```

yyjdelete:
AutoMapperMappingException.Message is work well again in `v5.0.0-ci-1052`

lbargaoanu:
@yyjdelete Yes, I saw that and fixed it yesterday.

Issue #258, opened 2012-10-04
Is it possible to create a mapping per type discriminator?


fabrikater:
Something like this:

Mapper.CreateMap\<DTO.Party, Party\>();
Mapper.CreateMap\<DTO.Party, SomeParty\>().Where(source => source.PartyTypeDiscriminatorValue == 1);
Mapper.CreateMap\<DTO.Party, SomeOtherParty\>().Where(source => source.PartyTypeDiscriminatorValue == 2);

var party = new DTO.Party() { PartyTypeDiscriminatorValue = 2 };

var result = Mapper.Map\<DTO.Party, Party\>(party);

Assert.IsTrue(result is SomeOtherParty)

julealgon:
I just posted a question on StackOverflow related to the problems I'm having with mapping based on a property of the original model in a inheritance scenario.

http://stackoverflow.com/questions/30128551/how-do-i-leverage-constructusing-in-a-mapping-with-inherited-types

Jimmy, I'm sure my problem is related to this. Could you perhaps take a look at that and suggest some workaround?

TylerCarlson1:
It is related to this and unfortunately there's nothing you can do with includes to make this work.
Issue is when you include one class the logic will then check every class that inherits from the base and use the highest derived ones for it.  Regardless if you included it or not.

Fix for this issue in code shown here 3ed6d8af5ccc12d0990e7a8bb57d02612006070a as part of #590

PR #590, opened 2014-09-15
Include with conditions on source type

TylerCarlson1:
Addresses issue #258

Added IncludeOnSourceType function which will map to destination if source condition is met.

Had to do a lot of internal re-factoring in multiple commits to show an underlying problem with how includes were handled, and then add the ability to specify conditions, default before which was did the types inherit, and now condition from the source to determine how to map to the destination.

Before it wasn't using the includedDeriveTypes as part of the possible included mappings, only whether one existed or not.  Ergo it assumed that if there were included types it would include them all.
So now it's going on the condition that type map must match types of included TypePair, as well as meet its condition to say the type map is applicable to be mapped over.

This pull request might need a few changes.  Better name for the new Include function call, internal re-ordering of include functionality, ect...  But this will allow #258 to be possible, along with others to make their own extensions off the Include(type, type, condition) for whatever scenario they might want, not just checking source and destination types.

### DIFF
--- a/src/AutoMapper/AutoMapper.csproj
+++ b/src/AutoMapper/AutoMapper.csproj
@@ -47,7 +47,7 @@
     <CodeAnalysisRuleSet>AllRules.ruleset</CodeAnalysisRuleSet>
     <DocumentationFile>bin\Debug\AutoMapper.xml</DocumentationFile>
     <NoWarn>1591,0618</NoWarn>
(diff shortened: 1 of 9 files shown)


jbogard:
I don't think one will jive any more with the move to just do TypePair matching going forward, unfortunately.

TylerCarlson1:
Yea.  This was out of date when you redefined caching and refactored include, on-top of the type pair thing.

Issue #597, opened 2014-09-23
AutoMapper 3.1.0+ Not Working with Array Type


stanwen:
AutoMapper 3.1.0+ seems not working well with non-basic value type of Array. If debugging through the below codes in Visual Studio with MSTest, when putting cursor on anOrder, it shows incorrect Array type EventFlagDTO, which should be EventFlag. And there are no values inside the array.

The same code was working with 3.0.0 and below. If changing Array to Generic List type, it's working fine. Otherwise need to uncomment customized EventTypes mapping to make it working for Array type.

We have a lot of codes using non-basic value type of Array. Is this intended change for 3.1.0+ or something missed in code? This is really a roadblock for upgrading to newer version of AutoMapper.

Thanks!

```
using System.Collections.Generic;
using System.Linq;
using AutoMapper;
using Microsoft.VisualStudio.TestTools.UnitTesting;

namespace MVC.Tests
{ 
    public class Order
    {
        public int ID { get; set; }
        public EventFlag[] EventTypes { get; set; }
    }

    public enum EventFlag
    {
        Disabled = 0,
        LogOnly = 1
    }


    public class OrderDTO
    {
        public int ID { get; set; }
        public EventFlagDTO[] EventTypes { get; set; }
    }

    public enum EventFlagDTO
    {
        Disabled = 0,
        LogOnly = 1
    }


    [TestClass]
    public class UnitTest
    {
        [TestMethod]
        public void TestMethod()
        {
            Mapper.CreateMap<EventFlagDTO, EventFlag>();
            Mapper.CreateMap<OrderDTO, Order>();

            //Mapper.CreateMap<OrderDTO, Order>()
            //  .ForMember(dest => dest.EventTypes,
            //             opts => opts.MapFrom
            //             (
            //               src => src.EventTypes.Select(element => (EventFlag)element).ToArray()
            //             )
            //);

            var anOrderDTO = new OrderDTO
            {
                ID = 10,
                EventTypes = new EventFlagDTO[] { EventFlagDTO.LogOnly, EventFlagDTO.Disabled }
            };

            var anOrder = new Order();
            Mapper.Map(anOrderDTO, anOrder);
        }
    }
}
```

PR #604, opened 2014-10-08
Update AssignableArrayMapper.cs

stanwen:
This is a fix for issue #597. See #597 for details.


jbogard:
Got rid of this class altogether, which also fixed the issue :)

stanwen:
Excellent and thanks!

Issue #445, opened 2014-01-15
Provide context information when mapping


adam-resdiary:
I've got a situation where I want to create a mapping, but the mapping depends on the individual call to Map. For example, say I have the following source and destination objects:

``` csharp
public class Restaurant
{
    ...
    public DbGeography Location { get; set; }
}

public class RestaurantDto
{
    ...
    public double? Distance { get; set; }
}
```

Then I've got a method that does a query based on a certain location, and needs to returns shops ordered by their location, along with their distance from that location. Something like this:

``` csharp
public PageDto<RestaurantDto> GetByLocation(double lat, double long, int page, int pageSize)
{
    Page<Restaurant> restaurants = _repository.GetByLocation(lat, long, page, pageSize);

    return Mapper.Map<PageDto<RestaurantDto>, Page<Restaurant>>(restaurants);
}
```

One solution is to wrap the Page<Restaurant> in another class that contains the current location, then make the mapping something like Mapper.CreateMap&lt;Wrapper&lt;Page&lt;Restaurant>>, PageDto&lt;RestaurantDto>>(). This ends up being quite messy though. What I'd like to be able to do instead is something like this:

``` csharp
public PageDto<RestaurantDto> GetByLocation(double lat, double long, int page, int pageSize)
{
    Page<Restaurant> restaurants = _repository.GetByLocation(lat, long, page, pageSize);

    var currentLocation = DbGeography.FromText(string.Format("Point({0} {1})", long, lat);
    var context = new Dictionary<string, object>{ { "CurrentLocation", currentLocation } };

    return Mapper.Map<PageDto<RestaurantDto>, Page<Restaurant>>(restaurants, context);
}
```

This context would then be passed down through all the layers in the mapper so I could do something like:

``` csharp
Mapper.CreateMap<Restaurant, RestaurantDto>()
    .ForMember(dto => dto.Distance, (o, c) => o.MapFrom(r => r.Location.Distance((DbGeography)c["CurrentLocation"])));
```

Does that seem sensible, or is there already an easier way to do this?

PR #624, opened 2014-11-06
Consistent ResolveUsing and ConvertUsing interfaces

srogovtsev:
Commit 8b3bc93 (described in #445) added an option to do `ResolveUsing` with a delegate that accepts `ResolutionResult`. I further extended this with an option of a delegate that accepts `ResolutionResult, Source`. I also added (in fact, just pushed to public) same overloads for `ConvertUsing`.

So, as a result, following methods are now available:

````c#
void ResolveUsing(Func<TSource, object> resolver);
void ResolveUsing(Func<ResolutionResult, object> resolver);
void ResolveUsing(Func<ResolutionResult, TSource, object> resolver);

void ConvertUsing(Func<TSource, TDestination> mappingFunction);
void ConvertUsing(Func<ResolutionContext, TDestination> mappingFunction);
void ConvertUsing(Func<ResolutionContext, TSource, TDestination> mappingFunction);
````

PS All that started because I needed `ConvertUsing(Func<ResolutionContext, TSource, TDestination> mappingFunction)` in my project.

### DIFF
--- a/src/AutoMapper/IMappingExpression.cs
+++ b/src/AutoMapper/IMappingExpression.cs
@@ -140,6 +140,18 @@ namespace AutoMapper
         void ConvertUsing(Func<TSource, TDestination> mappingFunction);
 
         /// <summary>
+        /// Skip member mapping and use a custom function to convert to the destination type
+        /// </summary>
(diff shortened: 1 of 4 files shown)


jbogard:
Merged!

Issue #648, opened 2014-12-12
Mapping Expressions with ICollection


zhylinsky:
Hello, 
I have a problem with mapping expression with a collection. 
Expression: 
dealers = Find(d => d.Branches.Any(b => branchIds.Contains(b.Id)));

        public IList<TModel> Find(Expression<Func<TModel, bool>> filter, params Expression<Func<TModel, object>>[] includes)
        {
            IQueryable<TEntity> query = SomeDbContext.Set<TEntity>();
            if (filter != null)
            {
                query = query.Where(Mapper.Map<Expression<Func<TEntity, bool>>>(filter));
            }
            foreach (var include in includes)
            {
                query = query.Include(Mapper.Map<Expression<Func<TEntity, object>>>(include));
            }
            return Mapper.Map<IEnumerable<TModel>>(query).ToList();
        }

Exception:
Mapping types:
Expression`1 -> Expression`1
System.Linq.Expressions.Expression`1[[System.Func`2[[Supersite.Models.Dealer],[System.Boolean]]]] -> System.Linq.Expressions.Expression`1[[System.Func`2[[Supersite.Entities.Dealer],[System.Boolean]]]]

Destination path:
Expression`1

Source value:
d => d.Branches.Any(b => value(Supersite.Repositories.EF.DealerRepository+<>c__DisplayClass0).branchIds.Contains(b.Id)) ---> System.NullReferenceException: Object reference not set to an instance of an object.
   at AutoMapper.Mappers.ExpressionMapper.MappingVisitor.VisitMember(MemberExpression node)
   at System.Linq.Expressions.MemberExpression.Accept(ExpressionVisitor visitor)
   at System.Linq.Expressions.ExpressionVisitor.Visit(Expression node)
   at System.Linq.Expressions.ExpressionVisitor.VisitMethodCall(MethodCallExpression node)
   at System.Linq.Expressions.MethodCallExpression.Accept(ExpressionVisitor visitor)
   at System.Linq.Expressions.ExpressionVisitor.Visit(Expression node)
   at System.Linq.Expressions.ExpressionVisitor.VisitLambda[T](Expression`1 node)
   at System.Linq.Expressions.Expression`1.Accept(ExpressionVisitor visitor)
   at System.Linq.Expressions.ExpressionVisitor.Visit(Expression node)
   at System.Linq.Expressions.ExpressionVisitor.VisitArguments(IArgumentProvider nodes)
[...]

PR #649, opened 2014-12-15
Fix ExpressionMapper error for sub property expression conversion

TylerCarlson1:
Fix Issues with Expression Mappings #648 when use properties which types have their own mappings that need to be converted over.

Multiple scenarios have been solved for now including the original problem.
Issue still lies with multiple level same type one to many relationships. (One to many relationships of different types work fine)
Ex: ```Tools.Any(t => t.SubTools.Any(t => t.ID == id))``` will fail to convert over cause Any function will still be assuming ToolDTO instead of Tool as generic parameter type.

### DIFF
--- a/src/AutoMapper/Internal/ReflectionHelper.cs
+++ b/src/AutoMapper/Internal/ReflectionHelper.cs
@@ -83,5 +83,18 @@ namespace AutoMapper.Impl
 
 			return null;
 		}
+
+        public static bool IsStatic(this MemberInfo accessorCandidate)
+        {
+            var fieldInfo = accessorCandidate as FieldInfo;
+            if (fieldInfo != null && fieldInfo.IsStatic)
+                return true;
+
+            var propertyInfo = accessorCandidate as PropertyInfo;
+            if (propertyInfo != null && ((propertyInfo.CanRead && propertyInfo.GetGetMethod().IsStatic) || (propertyInfo.CanWrite && propertyInfo.GetSetMethod().IsStatic)))
+                return true;
+
+            return false;
+        }
 	}
 }
--- a/src/AutoMapper/Mappers/ExpressionMapper.cs
+++ b/src/AutoMapper/Mappers/ExpressionMapper.cs
@@ -255,6 +255,9 @@ namespace AutoMapper.Mappers
 
             private PropertyMap PropertyMap(MemberExpression node)
             {
+                if (node.Member.IsStatic())
+                    return null;
+
                 var memberAccessor = node.Member.ToMemberAccessor();
                 var propertyMap = _typeMap.GetExistingPropertyMapFor(memberAccessor);
                 return propertyMap;
--- a/src/AutoMapper/Mappers/ExpressionMapper.cs
+++ b/src/AutoMapper/Mappers/ExpressionMapper.cs
@@ -67,6 +67,8 @@ namespace AutoMapper.Mappers
 
             private MethodCallExpression GetConvertedMethodCall(MethodCallExpression node)
             {
+                if (!node.Method.IsGenericMethod)
+                    return node;
                 var convertedArguments = Visit(node.Arguments);
                 var convertedMethodArgumentTypes = node.Method.GetGenericArguments().Select(t => GetConvertingTypeIfExists(node.Arguments, t, convertedArguments)).ToArray();
                 var convertedMethodCall = node.Method.GetGenericMethodDefinition().MakeGenericMethod(convertedMethodArgumentTypes);
--- a/src/AutoMapper/Mappers/ExpressionMapper.cs
+++ b/src/AutoMapper/Mappers/ExpressionMapper.cs
@@ -162,7 +162,11 @@ namespace AutoMapper.Mappers
                 var baseExpression = Visit(node.Expression);
                 var propertyMap = FindPropertyMapOfExpression(node.Expression as MemberExpression);
 
-                var typeMap = Mapper.FindTypeMapFor(propertyMap.SourceMember.GetMemberType(), propertyMap.DestinationPropertyType);
+                var sourceType = propertyMap.SourceMember.GetMemberType();
+                var destType = propertyMap.DestinationPropertyType;
+                if (sourceType == destType)
+                    return Expression.MakeMemberAccess(baseExpression, node.Member);
+                var typeMap = Mapper.FindTypeMapFor(sourceType, destType);
                 var memberExpression = new MappingVisitor(typeMap, node.Expression, baseExpression).Visit(node) as MemberExpression;
 
                 return Expression.MakeMemberAccess(baseExpression,memberExpression.Member);
--- a/src/AutoMapper/Mappers/ExpressionMapper.cs
+++ b/src/AutoMapper/Mappers/ExpressionMapper.cs
@@ -67,7 +67,7 @@ namespace AutoMapper.Mappers
 
             private MethodCallExpression GetConvertedMethodCall(MethodCallExpression node)
             {
-                var convertedArguments = node.Arguments.Select(e => base.Visit(e)).ToList();
+                var convertedArguments = Visit(node.Arguments);
                 var convertedMethodArgumentTypes = node.Method.GetGenericArguments().Select(t => GetConvertingTypeIfExists(node.Arguments, t, convertedArguments)).ToArray();
                 var convertedMethodCall = node.Method.GetGenericMethodDefinition().MakeGenericMethod(convertedMethodArgumentTypes);
                 return Expression.Call(convertedMethodCall, convertedArguments);
@@ -107,7 +107,6 @@ namespace AutoMapper.Mappers
 
             private Expression VisitAllParametersExpression<T>(Expression<T> expression)
             {
-                var parameters = expression.Parameters.ToArray();
                 var visitors = new List<ExpressionVisitor>();
                 for (var i = 0; i < expression.Parameters.Count; i++)
                 {
@@ -128,7 +127,6 @@ namespace AutoMapper.Mappers
 
                     var oldParam = expression.Parameters[i];
                     var newParam = Expression.Parameter(typeMap.SourceType, oldParam.Name);
-                    parameters[i] = newParam;
                     visitors.Add(new MappingVisitor(typeMap, oldParam, newParam));
                 }
                 return visitors.Aggregate(expression as Expression, (e, v) => v.Visit(e));
--- a/src/AutoMapper/Mappers/ExpressionMapper.cs
+++ b/src/AutoMapper/Mappers/ExpressionMapper.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Reflection;
+using AutoMapper.Internal;
 using AutoMapper.QueryableExtensions.Impl;
 
 namespace AutoMapper.Mappers
@@ -34,23 +35,25 @@ namespace AutoMapper.Mappers
 
         private class MappingVisitor : ExpressionVisitor
         {
-            private IList<Type> _destSubTypes;
+            private IList<Type> _destSubTypes = new Type[0];
 
             private readonly TypeMap _typeMap;
             private readonly Expression _oldParam;
             private readonly Expression _newParam;
+            private readonly MappingVisitor _parentMappingVisitor;
 
             public MappingVisitor(IList<Type> destSubTypes)
-                : this(null, Expression.Parameter(typeof(Nullable)), Expression.Parameter(typeof(Nullable)) )
+                : this(null, Expression.Parameter(typeof(Nullable)), Expression.Parameter(typeof(Nullable)), null)
             {
                 _destSubTypes = destSubTypes;
             }
 
-            private MappingVisitor(TypeMap typeMap, Expression oldParam, Expression newParam)
+            private MappingVisitor(TypeMap typeMap, Expression oldParam, Expression newParam, MappingVisitor parentMappingVisitor)
             {
                 _typeMap = typeMap;
                 _oldParam = oldParam;
                 _newParam = newParam;
+                _parentMappingVisitor = parentMappingVisitor;
             }
 
             protected override Expression VisitParameter(ParameterExpression node)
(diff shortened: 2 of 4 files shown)


TylerCarlson1:
@jbogard Does Expression mapping HAVE to support null substition expression conversion?

TylerCarlson1:
Fixed all known failing scenarios and updated unit tests to better reflect the scenarios, and be more readable.

TylerCarlson1:
Fixed exception thrown when expression contains static properties or fields.

jbogard:
So between #704, #659, #652 and #649, do we have a single solution here?

TylerCarlson1:
The single solution I have for solving all of the cases presented in #704 #649 and #659 is in https://github.com/TylerCarlson1/AutoMapper/tree/UseAsDataSource which I can submit a PR for.
UseAsDataSource() function is added here to match and improve on the function in #659.

#652 looks like it's a solution to parameter passing problem, which none of the solutions have the option to pass parameters.  They allow IQueryable translation along with fixing issues with Expression translating as it works now.
Currently UseAsDataSource() and its underlying translation can't support parameter passing.  Honestly I didn't even know that was a thing till 2 weeks ago.

jbogard:
PR away!

BlaiseD:
Seems to me UseAsDataSource does not pass the following tests. Without mapping the IQueryables, I don't think you can get away with NOT using projection:

        [TestMethod]
        public void Map_select_method()
        {
            Expression<Func<UserModel, object>> selection = s => s.AccountModel.ThingModels.Select(x => x.BarModel);
            Expression<Func<User, object>> selectionMapped = Mapper.Map<Expression<Func<User, object>>>(selection);
            Assert.IsNotNull(selectionMapped);
        }

        [TestMethod]
        public void Map_orderBy_thenBy_expression()
        {
            Expression<Func<IQueryable<UserModel>, IQueryable<UserModel>>> exp = q => q.OrderBy(s => s.Id).ThenBy(s => s.FullName);
            Expression<Func<IQueryable<User>, IQueryable<User>>> expMapped = Mapper.Map<Expression<Func<IQueryable<User>, IQueryable<User>>>>(exp);
            Assert.IsNotNull(expMapped);
        }

        private static void SetupAutoMapper()
        {
            Mapper.CreateMap<User, UserModel>()
            .ForMember(d => d.Id, opt => opt.MapFrom(s => s.UserId))
            .ForMember(d => d.FullName, opt => opt.MapFrom(s => s.Name))
            .ForMember(d => d.LoggedOn, opt => opt.MapFrom(s => s.IsLoggedOn ? "Y" : "N"))
            .ForMember(d => d.IsOverEighty, opt => opt.MapFrom(s => s.Age > 80))
            .ForMember(d => d.AccountName, opt => opt.MapFrom(s => s.Account == null ? string.Empty : string.Concat(s.Account.FirstName, " ", s.Account.LastName)))
            .ForMember(d => d.AgeInYears, opt => opt.MapFrom(s => s.Age))
            .ForMember(d => d.IsActive, opt => opt.MapFrom(s => s.Active))
            .ForMember(d => d.AccountModel, opt => opt.MapFrom(s => s.Account));

            Mapper.CreateMap<UserModel, User>()
                .ForMember(d => d.UserId, opt => opt.MapFrom(s => s.Id))
                .ForMember(d => d.Name, opt => opt.MapFrom(s => s.FullName))
                .ForMember(d => d.IsLoggedOn, opt => opt.MapFrom(s => s.LoggedOn.ToUpper() == "Y"))
                .ForMember(d => d.Age, opt => opt.MapFrom(s => s.AgeInYears))
                .ForMember(d => d.Active, opt => opt.MapFrom(s => s.IsActive))
                .ForMember(d => d.Account, opt => opt.MapFrom(s => s.AccountModel));

            Mapper.CreateMap<Account, AccountModel>()
                .ForMember(d => d.Bal, opt => opt.MapFrom(s => s.Balance))
                .ForMember(d => d.DateCreated, opt => opt.MapFrom(s => Helpers.TruncateTime(s.CreateDate).Value))
                .ForMember(d => d.ComboName, opt => opt.MapFrom(s => string.Concat(s.FirstName, " ", s.LastName)))
                .ForMember(d => d.ThingModels, opt => opt.MapFrom(s => s.Things));

            Mapper.CreateMap<AccountModel, Account>()
                .ForMember(d => d.Balance, opt => opt.MapFrom(s => s.Bal))
                .ForMember(d => d.Things, opt => opt.MapFrom(s => s.ThingModels));

            Mapper.CreateMap<Thing, ThingModel>()
                .ForMember(d => d.FooModel, opt => opt.MapFrom(s => s.Foo))
                .ForMember(d => d.BarModel, opt => opt.MapFrom(s => s.Bar));

            Mapper.CreateMap<ThingModel, Thing>()
                .ForMember(d => d.Foo, opt => opt.MapFrom(s => s.FooModel))
                .ForMember(d => d.Bar, opt => opt.MapFrom(s => s.BarModel));

            //Mapper.CreateMap<IEnumerable<Thing>, IEnumerable<ThingModel>>();
            //Mapper.CreateMap<IEnumerable<ThingModel>, IEnumerable<Thing>>();
            //Mapper.CreateMap<IEnumerable<User>, IEnumerable<UserModel>>();
            //Mapper.CreateMap<IEnumerable<UserModel>, IEnumerable<User>>();
        }

    }

    public class Account
    {
        public Account()
        {
            Things = new List<Thing>();
        }
        public int Id { get; set; }
        public double Balance { get; set; }
        public string FirstName { get; set; }
        public string LastName { get; set; }
        public DateTime CreateDate { get; set; }
        public ICollection<Thing> Things { get; set; }
    }

    public class AccountModel
    {
        public AccountModel()
        {
            ThingModels = new List<ThingModel>();
        }
        public int Id { get; set; }
        public double Bal { get; set; }
        public string ComboName { get; set; }
        public DateTime DateCreated { get; set; }
        public ICollection<ThingModel> ThingModels { get; set; }
    }

    public class Thing
    {
        public int Foo { get; set; }
    }

    public class ThingModel
    {
        public int FooModel { get; set; }
        public string BarModel { get; set; }
    }

    public class User
    {
        public int UserId { get; set; }
        public bool IsLoggedOn { get; set; }
        public int Age { get; set; }
        public bool Active { get; set; }
        public Account Account { get; set; }
    }

    public class UserModel
    {
        public int Id { get; set; }
        public string FullName { get; set; }
        public string AccountName { get; set; }
        public bool IsOverEighty { get; set; }
        public string LoggedOn { get; set; }
        public int AgeInYears { get; set; }
        public bool IsActive { get; set; }
        public AccountModel AccountModel { get; set; }
    }

TylerCarlson1:
It can support the order by statement and will convert the queryable and because the destination type is the same as the UseDataSource once it has to enumerate after query translation it will use Project().To() so that you get the benifits of selecting directly from EF.
```
var result = new User[0].AsQueryable().UseAsDataSource().For<UserModel>().OrderBy(s => s.Id).ThenBy(s => s.FullName);
```

Your select method on the other hand is not supported right now.
ThingModels specifically is causing problems being IEnumerable.  I think I account for it for the final result to know whether to use projection or not but that might be what's stopping it from being supported.

But in that case it would see the dest type is not the same so it wouldn't bother with projection because there's no need for it.

As for the PR  it might be a while to merge it into development and hope I can get VS 2015 to stop saying System.Object is not defined.

Unless you want to do the merge @jbogard?

TylerCarlson1:
ok I had it wrong why the select scenario wouldn't work.  ThingModels wasn't the problem because it was an IEnumerable.  It was cause it was the second property in the chain before using labmda.  Was looking for property of Account Model instead of ThingModel.

Works when UseDataSource against AccountModel instead of UserModel.

I'll add this scenario to the unit tests and get it working before I submit a PR.

jbogard:
I'll take care of getting it working in VS2015

TylerCarlson1:
Figured out how to fix the scenario.  Also didn't account for when string was returned, but fixed that allready.

Submitted PR #778 with changes

BlaiseD:
I have to disagree with that.

With PR #704, the component that performs the query (whether it is SQL, noSQL or XML) has no knowledge of AutoMapper or the business model objects like AccountModel or ThingModel.

See the AWSample.EF project in the XpressionMapping folder of #704. It has a reference to AWSample.EF.POCO but no references to AWSample.Domain or AutoMapper. The mappings between POCO and model objects and between POCO and model expressions occur at a higher level.  Not mapping the queryable takes that away.

One of the listed purposes of #704 is to provide an alternative to using Project().To().

I think #704 has better separation of concerns, loose coupling and all that good stuff.  Please find an overview in the README.md section here: https://github.com/BlaiseD/XpressionMapper.

Is there something #704 fails to handle?

TylerCarlson1:
My implementation also separates concerns much the same way yours does.  The Query doesn't know of AutoMapper or business Models when it gets called.
Like ```db.Accounts.Project().To<AccountModel>()``` will not pass Account model into the query it will send an automapper generated select statement to EF to translate to SQL.

And for your examples given before (Translated to UseAsDataSource implementation):
```
db.Users.UseAsDataSource().For<UserModel>().OrderBy<UserModel>(s => s.Id).ThenBy<UserModel>(s => s.FullName)
```
Translates
```OrderBy<UserModel>(s => s.Id)``` to ```OrderBy<User>(s => s.Id)```
```ThenBy<UserModel>(s => s.FullName)``` to ```ThenBy<User>(s => s.Account == null ? string.Empty : string.Concat(s.Account.FirstName, " ", s.Account.LastName))```
Will then call ```Projection().To<UserModel>()``` after all the Ordering to get the benefit of having a select statement do the conversion.
This is because return type didn't change from UserModel to anything else.
Equivalent Translation
```
db.Users.OrderBy<User>(s => s.Id).ThenBy<User>(s => s.Account == null ? string.Empty : string.Concat(s.Account.FirstName, " ", s.Account.LastName)).Project().To<UserModel>()
```

Select Example (Translated to UseAsDataSource implementation)
```
db.Users.UseAsDataSource().For<UserModel>().Select<UserModel>(s => (object)s.ThingModels.Select<ThingsModel>(x => x.BarModel))
```
Translates
```Select<UserModel>(s => (object)s.ThingModels.Select<ThingsModel>(x => x.BarModel))``` to ``Select<User>(s => (object)s.Things.Select<Things>(x => x.Bar))```
No Projection is done because return type is now object and not UserModel, so there's no reason to project.
Equivalent Translation
```
db.Users).Select<User>(s => (object)s.Thing.Select<Things>(x => x.Bar))
```

Project().To() is to provide the benifits of it when applicable, but maybe there should be an option to not project and use Map instead?
I see from your DateCreated Mapping it uses External method call, which would definitely not convert from EF to T-SQL and make projection impossible.  Also you have include property maps, which I do not have, because Projection will handle including them in the query to the database.

TylerCarlson1:
As for your readme it seems we are on the same page.  Also seeing your example of IRepository I would like to show you my own idea.

Still learning dotnetfiddle but should show idea. (Will try to add more examples of later)
Don't have UserAsDataSource() in a nuget packet so just made a placeholder.
Also added https://github.com/TylerCarlson1/Automapper.Collection to show how save and delete would work in tandem with a Generic Repository
https://dotnetfiddle.net/7PiKie

The IRepository uses ```IQueryable<TModel>``` which in the implementation will use UseAsDataSource() to convert from T to TModel.
This way you can just query off the queryable and do whatever you need and it will all translate into T-SQL statements with all the benefits of project WHEN applicable.

You can do Count(), Where(), FirstOrDefault(s => s.Id = id), OrderBy(), Select(), GroupBy(), and all the other queryable extensions, Chained one after the other, and the service using the repository will have no idea about what's happening in EntityFramework or Automapper.  No need to define all those functions and make new classes to define each new repository type.


Just have to enumerate the query to get the results and will get the values for you.
 Ex: Count(), First(), FirstOrDefault(),ToList(),ToArray(), ect.

For save and delete the readme for Automapper.Collection will explain how to set up.  It will handle figuring out uniqueness between the model and the EF object to know if to add, update, or delete
Even with One to many child relationships.  Ex: Account.Things if saved from Account.ThingModels it will compare all of them based on Thing's Primary key and update the list accordingly, so EF will persist the changes of the list.  (Add, Remove, Update values)

TylerCarlson1:
For Scenarios #704 that don't work I have a few.
I'll be honest I looked at your code and compared to see if it had equivalent functionality to mine.

Doesn't handle Nullable<TStruct> to TStruct or vice versa. Throws exception, but mine handles the conversion using UnaryExpression.
Lambda expressions only support lambdas with 1 parameter.  Ex: ```Where((entity, index) => index > 4)```
Lambda expressions that use parameters from parent lambdas.  Ex: ```p => p.Children.Where(c => c.Date < p.Date)```
Note:  I'm not entirely sure about this one.
You account for properties, but I don't see anywhere you account for fields.
Note: You SHOULDN'T be using fields, but field are supported, so they should be supported here.
WCF OData support. (Has to do with use case for @Venzhyk in using UseAsDataSource())

Scenarios I don't have
Quote conversion.  I've never seen or run into this before, and I have no clue if my logic works around it or not.  Would be great if you can provide me with an example of an expression that uses this expression type.

Other things of note:
Implementations are different in a few ways.
1) You use sub ExpressionVisitor classes for things like lambdas, where mine has everything in one visitor, aside from parameter re-placer.
2) Seems you have access to all the type maps in infoDictionary from the very beginning.  I only use the type presented at creation and add sub maps if they are needed.
3) You look for the . between properties, where as the original, which I didn't change, map the expression and recursively map the body, which is not guaranteed to be a property.  ```p => p.Children.First(c => c.Id > 10).Id``` might not work because it's a member expression but its body is a method expression.
4) Mine modifies the original expression mapping  All benefits of UseAsDataSource() go into ExpressionMapper as well.  Yours is a brand new expression mapper, meaning either ExpressionMapper goes away, or you have to replace the expression visitor it uses with your own, or maintain 2 sets of ExpressionVisitors.
5)  You made meaningful exception methods on unsupported cases.  Mine has no meaningful exception handling anywhere.

BlaiseD:
First,  the Mapping Helper used in CreatedDate has a DbFunction attribute so it does work very well with EF to SQL.

Nullable to struct is a judgement call.  I throw an exception to tell the user to use an explicit cast.  I thought it was safer than having the mapper handle it.

Fields are not difficult to support.  I’ll have to figure out what the other two entail.

Also, you refer to Project().To() as benefit.  PR 704 does not need it the way PR 704 has been implemented.

It seems like we are clear on what each PR does and does not do.

BlaiseD:
As for Quote Conversion, there's a good explanation of quote expressions here: http://stackoverflow.com/questions/3138133/what-is-the-purpose-of-linqs-expression-quote-method. I always forget after I read it.  As an example right-click Map_orderBy_thenBy_expression in Test Explorer and select "Debug Selected Test" and put a break point after the line "case ExpressionType.Quote:" in ArgumentMapper.  The code will pause when the OrderBy and ThenBy arguments are being mapped.

TylerCarlson1:
For Quote I don't know if I need to support it now, since OrderBy().ThenBy() works for my solution.
But if not it should be easy enough to translate your solution over to mine.

Looking at it again Things of Note number 3 might be an unsupported feature of #704 as well.
All the scenarios I provided that are not supported are not what anyone would consider normal use cases, so it could be personal preference until someone comes along and actually needs that case to work.

It looks like they pretty much do the same thing, minus some scenarios, but those can be fixed using the other PR as an example of how to solve it.

The main difference I see is in API, where #704 you have to map expressions of T and IQueryable where as #778 does all that without the user having to know about expression translation with UseAsDataSource() as well as hide the the Project().To() call. #778 doesn't support IQueryable Expression Translation mappings currently.
If I make my PR support IQueryable Expression mapping would that be sufficient to be able to use for your example @BlaiseD?  You wouldn't have to use UseAsDataSource() to map your filters and order by.

Another option is UseAsDatasource() can have an override to use Map instead of Project so that the user can select how they want the translation from T to TModel to be handled.
If Map was used the include portion would have to be done outside of UseDataSource() as an Include() statement, because it's tied specifically to Entity Framework.

I agree with @BlaiseD that we have both stated our cases for our PRs thoroughly enough.

BlaiseD:
It does support method calls e.g. Select, OrderBy.  There is a bug when the method's arguments are binary expressions.

You can also say with #704 that the user starts with an expression and simply maps it to another expression and does not need to know about UseAsDataSource.

I agree that your functionality is almost complete.  Merging IQuerable mapping with your solution sounds good. I think multiple Expression Visitors can live in the same solutions if we are mostly exposing extension methods.  Thanks for your feedback @TylerCarlson1 .

TylerCarlson1:
With bugs being binary expressions, could that be the argument of a method call is ```Predicate<T>``` instead of ```Func<T,bool>```?  I don't know if I've encountered this yet.

I'll try to get IQueryable mapping to work with ExpressionMapper.  It's proving to not be so easy right now.
I finally see now that your extensions are what handle the lambdas at the top level before being visited, and you have an override of different types.  Before I thought there was an all in one way XPressionMapper visitor handled IQueryable vs T.
It's possible your Extension methods might work with my expression visitor if slightly tweaked.

What I want to do is make an all in one solution if possible to handle all your expression mapping scenarios, and any others that might come about in one mapper.  I think we both have all the scenarios down with our mappers, except for when the arguments of the function of the expression are Generic and their arguments contain the types that we want to map to.  Ergo ```Expression<Func<IQueryable<TModel>,IQueryable<TModel>>>``` to ```Expression<Func<IQueryable<T>,IQueryable<T>>>```

TylerCarlson1:
Submitted change to support mapping of IQueryable to work in #778.

TylerCarlson1:
I also checked OrderBy().ThenBy() and it worked mapping them together.  It handled the quote just fine.  When I tried to visit the quote a different way, all the mapping tests failed.  I think with this #778 should be able to achieve what #704 does.

jbogard:
Merged in the other PR

Issue #646, opened 2014-12-11
Nested Project().To<T>(null, params string[] membersToExpand). How to?


Schlurcher:
Given the following class hierarchy and ExplicitInclude() enabled on each navigation property.
```
Root {
  Nested
  Stuff
}

Nested {
  Stuff
}
```
calling `source.Project().To<Root>(null, "Nested.Stuff")` will not properly expand the nav property `Root.Nested.Stuff`. Curiosly calling `To<Entity>(null, "Nested", "Stuff")` will do so, but will also expand `Root.Stuff`, which is not always desirable. 

I searched the interwebs for a good while on how to use Project().To() and ExplicitInclude() , but since 3.3 is pretty new there is hardly any information to be found. :(

Am I doing something wrong, is this by design or did I stumble upon a bug?

jbogard:
Fixed!

PR #652, opened 2014-12-22
Allow explicit expansion for nested members in projection

csantero:
This PR addresses #646. Given these destination classes:

```
public class Post
{
    public string Name {get; set;}
    public Author Author {get; set;}
}

public class Author
{
    public string Name {get; set;}
    public ICollection<Post> Posts {get; set;}
}
```

the following calls will expand the Author property but not Name (assuming ExplicitExpansion has been set for both mappings).
```
Project().To<Post>(null, p => p.Author, p => p.Author.Name);
Project().To<Post>(null, "Author", "Author.Name");
```

### DIFF
--- a/src/AutoMapper/QueryableExtensions/Extensions.cs
+++ b/src/AutoMapper/QueryableExtensions/Extensions.cs
@@ -164,7 +164,13 @@ namespace AutoMapper.QueryableExtensions
                     continue;
 
                 var propertyTypeMap = mappingEngine.ConfigurationProvider.FindTypeMapFor(result.Type, propertyMap.DestinationPropertyType);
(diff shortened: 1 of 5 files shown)


TylerCarlson1:
Looked at some more and you can't do strait up mappings to queryable like I said.  But there is expression mappings to do OrderDto to Order

csantero:
I tried to rebase this but it seems the project has made some changes since January and I can no longer get the solution to build in either VS2013 or VS2015. I get approximately 15000 errors similar to the following:

```
Predefined type 'System.Object' is not defined or imported	AutoMapper.DNX Core 5.0
```

jbogard:
This was fixed by #688

PR #659, opened 2015-01-06
IQueryable mapper. Make ability to create query with Dto objects and execute query with real objects

Venzhyk:
My idea is a query mapper. 

I propose to make ability to create query with Dto objects and execute query on real objects.

Example:
class Order { Id; Name; }
class OrderDto { OrderId; ProductName; }

We have OData service that expose OrderDto object.
OrderDto object is mapped to data base real entity Order that stored in database.

OData protocol provide ability to apply some filters, select fields and make order statements on OrderDto objects.

You able to create query on client side like this: 
```csharp
ODataClient.OrderDtoCollection
     .Where(o => o.ProductName.StartWith("Book"))
     .OrderBy(o => o.OrderId)
     .Select(o => o.OrderId);
```
This query will be serialized into string (URL), sended to server, deserealized and executed on server side collection of OrderDto objects.

I want to be able to convert that query base element type from OrderDto  to Order and replace all properties in query.
Map query above to:
```csharp
DbContext.OrderSet
     .Where(o => o.Name.StartWith("Book"))
     .OrderBy(o => o.Id)
     .Select(o => o.Id);
// and project results back to dto objects
```

Please look to tests:
AutoMapperSamples/Mappers/Query
AutoMapperSamples/Mappers/QuerySourceInjection

Note: it's only prototype of idea


jbogard:
There is the ability to map expression trees, does that not help in this situation?

TylerCarlson1:
You might be able to make expression from OrderDTO and serialize it and then map it over to expression of Order

http://expressiontree.codeplex.com/ can serialize expressions

Ex:
```
Expression<Func<IQueryable<OrderDTO>,IEnumerable<int>>> orderDtoExpression = 
    orders => orders.Where(o => o.ProductName.StartWith("Book"))
                    .OrderBy(o => o.OrderId)
                    .Select(o => o.OrderId);
var orderExpression = Mapper.Map<Expression<Func<IQueryable<OrderDTO>,IEnumerable<int>>>,Expression<Func<IQueryable<Order>,IEnumerable<int>>>>(orderDtoExpression);
```
orderExpression should be something like
```
orders => orders.Where(o => o.Name.StartWith("Book")).OrderBy(o => o.Id).Select(o => o.Id)
```

In 3.3.0 might not be able to handle every case right now, but I have PR #649 that's trying to fix the expression mapping issues.  I'll be happy to fix failing cases if you can find any.

PS.  OrderBy can be visited and genericly replaced like any other MethodCallExpression I believe, so you don't have to make specific scenarios for every possible LINQ Extension method.

Venzhyk:
I didn't know about ability to map expressions.
Let me look on that.
I think this could partialy resolve my problem.

Venzhyk:
Hi, sorry for long delay.

I've investigated your advice and found that it might be not exactly what I need.
Actually, my main concern is ability to wrap original query by another query.

For e.g. I have DB model with EF context and I want to expose this model with OData.


Simple example:
```csharp
class Order { Id; Name; } -- DB Model
class OrderDto { OrderId; ProductName; } - OData DTO object

Mapping: Id <--> OrderId; .Name <--> ProductName
```
All what I need is to be able to expose ```IQueryable<OrderDto>``` object.
I don't know what exactly queries will be applied on this query object by users, so I cannot predict expression I shoud map.

It could be ```Expression<Func<IQueryable<OrderDTO>,IEnumerable<int>>>```  or ```Expression<Func<IQueryable<OrderDTO>,int>>``` or  even ```Expression<Func<IQueryable<OrderDTO>,IEnumerable<Anonymous_Type>>>``` .

Example below don't work for me because "```Project().To<>()```" methods pair adds "```Select```" expression to query and it force OData to fetch all data from server and execute all next expressions on client site.
```csharp
IQueryable<Order> orders = ...;

orders.Project().To<OrderDto>() ... // 'select' expression was added here
```


Sketch of my proposal:
```csharp
// we have db query from db context
IQueryable<Order> orders = ...; 

// we create dto query that we return to user.
// dto query contains information about db query and can convrert itself to db query.
IQueryable<OrderDto> dtoOrders = orders.UseAsDataSource().For<OrderDto>();
return dtoOrders;

/// -------- Somewhere in client code

// here we create dto query and we don't know anything about db object or db query, we operate only by dtos
var query = dtoOrders.Where(..).Order(..)

// here dto query will be converted to db query, executed and result will be converted back to dto's objects
List<OrderDto> results = query.ToList();
```
Client code should not know anything about Automapper, DB etc.

jbogard:
But client code only needs to know about OrderDto?

Would it be possible to create just a small sample app so I can play around with your idea? I've never used OData so I don't really know where to begin.

TylerCarlson1:
I don't think expression mappings worked like I suggested earlier, but you should be able to map each expression from a where and order by expression into an expression of the DB object from the DTO.

If you want a wrapper where you essentially hide everything I would suggest looking into EnumerableQuery and make a class wrapping around that.
Make the class have `EnumerableQuery<TDB>` as base and have `Automapper<TDTO>` and `CreateQuery<TDTO>` evaluate and map the expression from TDTO to TDB.  Then when you have to evaluate you can  `Project().To<TDTO>()`.

All IQueryable extensions seem to use IQueryProvider.CreateQuery so that should be where the translation of Expressions should happen.

It's not a simple process, but that should get you the hidden Automapper conversion wrapping functionality, and have the client know nothing of the DB objects or Automapper.

Venzhyk:
Please, find example here https://github.com/Venzhyk/AutoMapper.Sketch

Also please note I use my version of AutoMapper libs from https://github.com/Venzhyk/AutoMapper

TylerCarlson1:
Looked at it more and you already did what I suggested.
The problem with the current expression mapping not being applicable for you is that it only does lambda expressions and what you need is to convert an Expression body.  Also visiting constants of generic type, which is a scenario I haven't thought of.

I'll look into making Expressions map from non-Lambda's Expressions and VisitConstant to work with generic type in my Pull Request.
It should be possible then to map expressions of `IQueryable<TDB>` to `IQueryable<TDTO>` and thus translate queries using your wrapper class.

As for predicting expressions gonna have to map.  You would have to generate the destination type expression and pass it into automapper in your wrapper to use the expression mapper.

Venzhyk:
@TylerCarlson1, thanks for your suggestion.

I've tried to use ExpressionMapper in my QueryMapperVisitor:
https://github.com/Venzhyk/AutoMapper/commit/e47ee86cdff63cb5cdc35b138509b6dddd89edb8

However I met several issues/bugs in ExpressionMapper:

1. ExpressionMapper cannot resolve expressions like this:
```
 o => o.ContactName == string.Empty;
```
It stucks on ```string.Empty``` and tries to find map for ```Empty``` property

2. ExpressionMapper cannot resolve nested properties.
```c#
 public class OrderDto
 {
        public string ContactName { get; set; }
 }

    public class Order
    {
        public Customer Customer { get; set; }
    }

    public class Customer
    {
    }

// Plain property to nested works fine
Expression<Func<OrderDto, bool>> dto = o => o.ContactName == "Bob";
var db = Mapper.Map<Expression<Func<Order, bool>>>(dto);

// Nested property to plain throws exception. It cannot find propertyMap for Name property.
// It Seems it tries to search only in Order-> OrderDto map, not in Customer->OrderDto
 Expression<Func<Order, bool>> db = o => o.Customer.Name == "Bob";
 var dto= Mapper.Map<Expression<Func<OrderDto, bool>>>(db);

```

BlaiseD:
Igor,


Consider looking at my pull request #690.  The fork is here: https://github.com/BlaiseD/AutoMapper  .


It handles mapping Expression<Func<IQueryable<UserModel>, IQueryable<UserModel>>> exp; to Expression<Func<IQueryable<User>, IQueryable<User>>> exp;


It also handles the string.Empty case and the nested properties.


I think it provides the abstraction you're looking for where the Where filter, the IQueryable expression, and the include expressions are all defined against the DTO rather than the POCO (EF class).


Also if you're mapping the Where filter, the IQueryable expression, and the include expressions then you shouldn't need to return an IQuerable<T>. You can return a List<T> instead.


One thing Tyler Carlson also mentioned before though: when mapping value types to their nullable counter parts you still need to map to the nullable's Value property.


Hope this helps.

[https://avatars2.githubusercontent.com/u/7595209?v=3&s=400]<https://github.com/BlaiseD/AutoMapper>

BlaiseD/AutoMapper · GitHub
AutoMapper - A convention-based object-object mapper in .NET.
Read more...<https://github.com/BlaiseD/AutoMapper>



________________________________

TylerCarlson1:
What I suggested was not what you did.  I meant to not use QueryMappingExpression and use Expression mapping in its place.  That turns out not to work, and you have to use ExpresisonMapper.MappingVisitor to visit the values instead of QueryMappingExpression.

I worked on it and got a working solution with your UseDataSource() method call.
I even got Select to work and with anonymous types.
If you go to UnitTests.Net4/Bug/ExpressionMapping.cs I added the UseDataSource() method for testing my projections out and they all work.  I got quite a few unique scenarios there.

Solution is located in
https://github.com/TylerCarlson1/AutoMapper/tree/UseAsDataSource

TylerCarlson1:
Oh I didn't check but it still CAN'T do `o => o.ContactName == string.Empty`
It can do `o => o.ContactName == ""` and `o => o.ContactName.Equals("")` though.

Can look through QueryMappingExpression again to see how it handles static constants and apply it to ExpressionMapper.MappingVisitor

TylerCarlson1:
Updated my solution (UseAsDataSource) and I fixed an issue I saw with Execute.
Execute now calls create query or execute from the base type depending on if the type is enumerable or not.
Execute from the provider is meant to return a single value and create query is meant to provide queryable which in this case is an IEnumerable.

Now it should be able to do ToList() and Count(), First() or whatever else linq queries OData supports, along with EF queries using UseAsDataSource method.

TylerCarlson1:
Fixed `o => o.ContactName == string.Empty` not working, and now if result type is an enumeration of the destination type it will properly Project instead of mapping the results.  Otherwise it will map like normal.

dotdot1:
Hi Folks>
Is this PR integrated ? If not is there a nuget / dll that I can try this out?

jbogard:
Merged!

PR #705, opened 2015-04-20
Adding detail to exception thrown by projection

lorenpaulsen:
I hit this exception and was wished it would actually tell me which property couldn't be mapped. This change helped me and maybe it could help someone else.

The exception goes from reading like this 

```
Unable to create a map expression from System.Nullable`1[System.Int32] to System.Int32
``` 

to reading like this 

```
Unable to create a map expression from InvoiceItem.ProductId (System.Nullable`1[System.Int32]) to InvoiceItemModel.ProductId (System.Int32)
```


jbogard:
Brilliant! And nice meeting you at dotnetfringe!

Issue #724, opened 2015-05-27
FindTypeMapFor ignores type parameters when src/dest objects passed


cherrydev:
The offending line is line 345 of ConfigurationStore (at commit c986717, at least)

`return FindTypeMapFor(source?.GetType() ?? sourceType, destination?.GetType() ?? destinationType);`

I'm not sure what was actually intended, so I don't have a PR for you, but obviously discarding `sourceType` and `destinationType` when `source` and `destination` are designated is not what's intended, since it means that when calling `Mapper.Map<TSource,TDest>(srcObject)` TSource is ignored and effectively replaced with `srcObject.GetType()`.  This breaks when, for example, you're using EF dynamic proxies and `typeof(TSource) != srcObject.GetType()`.

jbogard:
Part of the intention is to try and simplify things - the rules for determining what TypeMap to use were confusing as all get out.

What is supposed to happen is srcObject.GetType() might not match, but AutoMapper will traverse the parent types until it finds a match. What does the type hierarchy look like for EF proxies? `Mapper.Map<TSource, TDestination>(srcObject)` still enforces that srcObject is an object that can be cast to TSource via the generic type parameter.  You can't do `Mapper.Map<Duck, DuckDto>(giraffe)`

cherrydev:
Well, as I said, I wasn't quite sure what the intention was, but this change leads to a regression that I can't find a workaround for.  #723 has more information on the specifics of the circumstance.  But now, under some circumstances, if I have:
```csharp
abstract class Animal {
…
}

class Duck extends Animal {
…
}

cfg.CreateMap<Animal, AnimalDto>()
  .Include<Duck, DuckDto>
  …
  .ReverseMap();

…

cfg.CreateMap<Duck, DuckDto>().ReverseMap();
```

As of that commit, `Mapper.Map<Animal,AnimalDto>(aDuck)` fails with `Instances of abstract classes cannot be created.` but (as far as I can tell) only when aDuck is a dynamic proxy subclass of Duck.

I'm creating my mapping exactly as described [here](https://github.com/AutoMapper/AutoMapper/wiki/Mapping-inheritance#mapping-configuration-inheritance-in-automapper-20)

cherrydev:
Okay, I've found the problem.  The algorithm that you're using for searching for an appropriate mapping pair in `FindTypeMapFor(TypePair typePair)` is wrong and the results of it are being cached.

Here's our _typeMaps:

src | dest | note
[...]

PR #725, opened 2015-05-28
Fix for #724

cherrydev:
Used your feedback as a basis for this.

### DIFF
--- a/src/AutoMapper/ConfigurationStore.cs
+++ b/src/AutoMapper/ConfigurationStore.cs
@@ -335,7 +335,7 @@ namespace AutoMapper
         {
             var typePair = new TypePair(sourceType, destinationType);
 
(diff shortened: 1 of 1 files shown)


jbogard:
OK I had to un-merge this, there were hundreds of failing tests D'OH

TylerCarlson1:
@jbogard  You need to keep FindTypeMapFor(TypePair) along with the new lambda in the change in order for all the tests to pass.  The tests fail because you aren't using Mapper.Initilize in Every Single test

cherrydev:
1000 pardons, I hadn't taken the extra time to get the full solution loading in my dev environment and the Unit Tests running (because of all the extra platforms and frameworks I don't have installed).  The proper fix is now at https://github.com/AutoMapper/AutoMapper/pull/727

PR #727, opened 2015-05-28
Fix for #724, tests now all pass again

cherrydev:
As per @TylerCarlson1 's advice, now calls FindTypeMapFor as well and now all tests pass again.


jbogard:
Merged!

PR #736, opened 2015-06-16
Fix Expression`1 -> Expression`1 Error

davidtme:
TypeMap.GetExistingPropertyMapFor can return null so rather than blowing up, return the original expression.


jbogard:
What's the scenario here? Do you have a set of code that shows what this fixes?

Issue #732, opened 2015-06-05
int to nullable decimal fails


JonasSamuelsson:
First mapping works and second throws, but AssertConfigurationIsValid passes.
```cs
class Program
{
    static void Main(string[] args)
    {
        Mapper.DynamicMap<D1>(new I());
        Mapper.DynamicMap<D2>(new I());

        Mapper.CreateMap<I, D2>();
        Mapper.AssertConfigurationIsValid();
    }
}

class I
{
    public int Number { get; set; }
}

class D1
{
    public decimal Number { get; set; }
}

class D2
{
    public decimal? Number { get; set; }
}
```

jbogard:
What version?

JonasSamuelsson:
3.3.1

jbogard:
Welp, this is a bug.

lbargaoanu:
Can I take this? I understand that AssertConfigurationIsValid should report this case as an error.

jbogard:
Sure, go for it!

JonasSamuelsson:
:+1:

lbargaoanu:
[The first try.](https://github.com/AutoMapper/AutoMapper/pull/738)

PR #738, opened 2015-06-16
Int to nullable decimal bug 732

lbargaoanu:
For #732. I ran the tests only for .NET 4. If the NullableMapper does something useful, I can write a test for it. Otherwise I should probably remove it altogether.


jbogard:
Merged!

Issue #720, opened 2015-05-14
Enums unhelpfully converted to strings, in object-valued property, even when explicitly mapped


peterchase:
When the destination property/field is `object`, AutoMapper converts the value to a string, even when an explicit mapping has been put in place that resulted in a specific `enum` value. This is most unhelpful - it should keep the `enum` value explicitly assigned in the mapping expression.

See also the StackOverflow question asked by my manager, to which I contributed only a work-around, not a nice fix: http://stackoverflow.com/questions/30127697/how-can-i-stop-automapper-converting-enum-values-into-strings

```csharp
    using AutoMapper;
    using AutoMapper.Mappers;
    using NUnit.Framework;

    namespace AutoMapperTest4
    {
        [TestFixture]
        public class AutomapperTest
        {
            [Test]
            public void TestAutomapperMappingFieldsOfTypeEnumObject()
            {
                // Configure 
                var configuration = new ConfigurationStore(new TypeMapFactory(), MapperRegistry.Mappers);
                var mapper = new MappingEngine(configuration);
                IMappingExpression<Source, Target> parentMapping = configuration.CreateMap<Source, Target>();
                parentMapping.ForMember(dest => dest.Value, opt => opt.MapFrom(s => ConvertValueToTargetEnumValue(s)));

                var source = new Source { Value = SourceEnumValue.Mule };

                var target = mapper.Map<Target>(source);

                Assert.That(target.Value, Is.TypeOf<TargetEnumValue>()); // Fails.  targetParent.Value is a string "Mule".
            }

            private static TargetEnumValue ConvertValueToTargetEnumValue(Source s)
            {
                return (TargetEnumValue)s.Value;
            }
        }

        public enum SourceEnumValue
        {
            Donkey,
            Mule
        }

        public enum TargetEnumValue
        {
            Donkey,
            Mule
        }

        public class Source
        {
            public object Value { get; set; }
        }

        public class Target
        {
            public object Value { get; set; }
        }
    }
```

jbogard:
Is this something that just started showing up?

peterchase:
We just upgraded from AutoMapper 1.1 to 3.3.1. It didn't happen in 1.1. I am not sure at what point, between those two versions, the behaviour changed.

jbogard:
Wow a little too far for Git bisect ok I'll look into it.

lbargaoanu:
It does seem somewhat arbitrary, but that's all I have for now :)

PR #740, opened 2015-06-18
Enum to string bug 720

lbargaoanu:
#720
It does seem somewhat arbitrary, but that's all I have for now :)


peterchase:
Ad-hoc testing now, by pulling this branch into the 3.3.1-based AutoMapper in our mapping-heavy app.  So far, I'm finding that it's still working. However, the testing so far is with code in our app to work-around the AutoMapper bug that this pull request aims to fix. The next step will be to try to remove that work-around, and see whether our app now keeps working...

jbogard:
Merged!

Issue #733, opened 2015-06-10
Better error message when cannot find property


JakeGinnivan:
I am mapping to a type which has properties with no getters, so I have to use strings to specify the property names.

`.MapFrom("Foo", ...);`

When `Foo` doesn't exist there is a null reference exception, it would be great to catch this and wrap it so I know which property no longer exists.

lbargaoanu:
ArgumentOutOfRangeException seems appropriate.

jbogard:
Fixed!

PR #741, opened 2015-06-18
Non existing property bug 733

lbargaoanu:
#733
Perhaps a more thorough checking of the resulting exception object would be in order.


jbogard:
Merged.

PR #744, opened 2015-06-19
Introduced MapperContext which de-static-izes AutoMapper concerns

mwpowellhtx:
Let me know when you've had a look at it. Coordinating efforts, I will work on the wiki and say a blurb or two about MapperContexts. Thanks so much.


mwpowellhtx:
Closed. Will resubmit with corrections.

PR #745, opened 2015-06-19
Introduced MapperContext which de-static-izes AutoMapper concerns

mwpowellhtx:
Let me know when you've had a look at it. Coordinating efforts, I will work on the wiki and say a blurb or two about MapperContexts. Thanks so much.


mwpowellhtx:
Let me know how I can help move this one along. Thank you...

mwpowellhtx:
Let me know should I need to rebase anything and/or resubmit. Thank you...

jbogard:
So one thing you need to do is revert all the *.proj changes. Those projects use wildcards to include files and are only actually there for the tests. Unfortunately the only way to get the projects to see new files is to close/open the solution.

mwpowellhtx:
Ah, it's good to know. Thanks for the feedback. I'll get on that right away.

mwpowellhtx:
The easiest way to do this is resubmit I believe.

Issue #688, opened 2015-03-11
Nested mapping projections not respecting explicit expansions


pedrodbsa:
I expected that AutoMapper would be able to handle nested explicit expansions.
Don't know if this only happens for projections or not, or if this is even possible.

I've also tried using the string explicit expansion notation but the result is the same.

```
public class FuEntity
{
	public ManEntity Man { get; set; }
}

public class ManEntity
{
	public ChuEntity Chu { get; set; }
}

public class ChuEntity
{
	public int Prop { get; set; }
}

public class Fu
{
	public Man Man { get; set; }
}

public class Man
{
	public Chu Chu { get; set; }
}

public class Chu
{
	public int Prop { get; set; }
}

public class MapperConfiguration : Profile
{
	protected override void Configure()
	{
		Mapper.CreateMap<FuEntity, Fu>()
			.ForMember(dest => dest.Man, opt => opt.MapFrom(src => src.Man))
			.ForMember(dest => dest.Man, opt => opt.ExplicitExpansion());

		Mapper.CreateMap<ManEntity, Man>()
			.ForMember(dest => dest.Chu, opt => opt.MapFrom(src => src.Chu))
			.ForMember(dest => dest.Chu, opt => opt.ExplicitExpansion());

		Mapper.CreateMap<ChuEntity, Chu>()
			.ForMember(dest => dest.Prop, opt => opt.MapFrom(src => src.Prop));
	}
}

public class Test
{
	public Test(IQueryable<FuEntity> test)
	{
		IQueryable<Fu> result1 = test.Project().To<Fu>(new { }, m => m.Man, m => m.Man.Chu);

		// Man object mapped 
		// Chu object not mapped!!!
	}
}
```

jbogard:
Two things. First, you should call CreateMap in your Profile, not Mapper.CreateMap. Second, your "MapFrom" configurations are redundant. The Auto part of AutoMapper is that properties of the same name are auto-mapped.

But no, AutoMapper doesn't support this today. I'll add it to the list for 4.0.

lbargaoanu:
With strings it works with To<Fu>(null, "Man", "Chu").
The PR is for the Expression overload.

lbargaoanu:
I think #652 solved this one.

jbogard:
Fixed.

PR #750, opened 2015-06-24
Nested mapping projections with explicit expanding bug 688

lbargaoanu:
#688
Apparently it works by returning the inner property access.


jbogard:
Merged.

PR #755, opened 2015-06-29
Possibility to define the MapFrom also by a string

jogibear9988:


jbogard:
Would it be possible to add a test on this one?

jogibear9988:
I could not run the Test, because I could not compile the Test assembly...
But it should work!

jbogard:
Merged.

Issue #654, opened 2014-12-29
Support constructor parameter mappings for LINQ expressions


jbogard:
This should work:

```
public class Source {
    public int Value { get; set; }
}
public class SourceDto {
    public SourceDto(int value) {
        _value = value;
    }
    private int _value;
    public int Value {
        get { return _value; }
    }
}

dbContext.Sources.Project().To<SourceDto>().ToList();
```

The expression mapper doesn't respect ConstructorParameterMaps.

lbargaoanu:
I hope it's the right test :)

jbogard:
Fixed.

lbargaoanu:
Maybe it's a little late :) But why is this useful if EF doesn't support it? NHibernate?

PR #759, opened 2015-06-30
The expression mapper doesn't respect ConstructorParameterMaps; #654

lbargaoanu:
I hope it's the right test :)


jbogard:
Merged.

Issue #712, opened 2015-04-29
AutoMapper doesn't support multi-dimensional arrays


jcoburn:

lbargaoanu:
In what way? :) I have a passing test.

    public class FatalError : AutoMapperSpecBase
    {
        To _mapped;
        int[][] _array;
        MappingEngine _factory;

        protected override void Establish_context()
        {
            var store = new ConfigurationStore(new TypeMapFactory(), MapperRegistry.Mappers);
            store.CreateMap<From, To>();
            _factory = new MappingEngine(store);
        }

        protected override void Because_of()
        {
            _array = new int[][] { new[] { 1 }, new[] { 1, 2 }, new[] { 1, 2, 3 } };
            _mapped = _factory.Map<From, To>(new From { Array = _array });
        }

        [Fact]
        public void Should_map_succesfully()
        {
            _mapped.Array.ShouldBeSameAs(_array);
        }

        public class From
        {
            public int[][] Array { get; set; }
        }
        public class To
        {
            public int[][] Array { get; set; }
        }
    }

jbogard:
Verified.

PR #770, opened 2015-07-07
Support multidimensional arrays

lbargaoanu:
#712 Just a simple test. If anybody uses it, we'll know what works :)


jbogard:
Merged.

Issue #769, opened 2015-07-06
Mapping a collection from an abstract base class via "Get" convention fails


dplaskon:
This appears to be a regression in Version 4: When you attempt to map a collection from an abstract base class using the default "Get" convention - the destination fails to be populated.

See below for a failing test case (please excuse the vb.net!):

```vb.net
        Public MustInherit Class SourceBase

            Public Function GetItems() As IEnumerable(Of Integer)

                Return {1, 2, 3}

            End Function

        End Class

        Public Class Source
            Inherits SourceBase

        End Class


        Public Class Destination

            Public Property Items As IEnumerable(Of Integer)

        End Class


        <Test()>
        Public Sub Test_IEnumerable_Mapping()

            Mapper.CreateMap(Of Source, Destination)()

            Dim source As New Source()

            Dim dest = Mapper.Map(Of Destination)(source)

            Assert.NotNull(dest.Items) ' Fails.
            CollectionAssert.AreEqual({1, 2, 3}, dest.Items)

        End Sub
```

lbargaoanu:
I believe it stopped working when removing BindingFlags.

jbogard:
Fixed!

PR #771, opened 2015-07-07
Return also methods from base types when looking for source members

lbargaoanu:
#769
[GetMethods] (https://msdn.microsoft.com/en-us/library/td205ybf%28v=vs.110%29.aspx) returns all the members, including from base types. It seems to be supported on all Windows platforms.


jbogard:
Merged!

Issue #777, opened 2015-07-09
NullReferenceException when mapping nullable int to nullable enum


s-tarasov:
NullReferenceException when mapping nullable int to nullable enum when source object field is null:
``` c#
public enum Values
{
	One = 1,
	Two = 2,
	Three = 3
}

public class A
{
	public int? Field { get; set; }
}

public class B
{
	public Values? Field { get; set; }
}

[Test]
public void Test2()
{
	var a = new A { Field = null };
	Mapper.CreateMap<A, B>();
	var b = Mapper.Map<A, B>(a);
	Assert.AreEqual(null, b.Field);
}
```

Output:
```
AutoMapper.AutoMapperMappingException was unhandled by user code
  HResult=-2146233088
  Message=

Mapping types:
Nullable`1 -> Nullable`1
System.Nullable`1[[System.Int32, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089]] -> System.Nullable`1[[Norm.Services.Tests.Mapping.EmployerMappingTests+Values, Norm.Services.Tests, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null]]

Destination path:
B.Field.Field

Source value:
(null)
  Source=AutoMapper
  StackTrace:
       at Norm.Services.Tests.Mapping.EmployerMappingTests.Test2() in e:\Work\Pronto\Norm\UnitTests\Norm.Services.Tests\Mapping\EmployerMappingTests.cs:line 45
  InnerException: System.NullReferenceException
       HResult=-2147467261
       Message=Object reference not set to an instance of an object.
       Source=AutoMapper
       StackTrace:
            at AutoMapper.Mappers.EnumMapper.Map(ResolutionContext context, IMappingEngineRunner mapper) in c:\dev\AutoMapper\src\AutoMapper\Mappers\EnumMapper.cs:line 69
            at AutoMapper.MappingEngine.AutoMapper.IMappingEngineRunner.Map(ResolutionContext context) in c:\dev\AutoMapper\src\AutoMapper\MappingEngine.cs:line 246
       InnerException:
```

aleha84:
need to fix it as soon as possible.

AlexanderSysoev:
I see the same behavior. Blocking exception for me...

jbogard:
Fixed!

s-tarasov:
Awesome!

PR #780, opened 2015-07-10
Nullable int to nullable enum

lbargaoanu:
#777


lbargaoanu:
@jbogard Did you consider a special NullMapper, so the rest of the mappers don't care, or it's more complicated than that?

jbogard:
Yeah, nullables just suck. No other way to say it. Merged!

Issue #628, opened 2014-11-19
DynamicMap doesn't map arrays


macpak:
Hi,

The test below doesn't pass. I think it should.

                [Test]
               public void  GivenArrayOfFoo_WhenMappingToArrayofBar_ShouldGetTheSameNumberOfItemsInArray()
               {
                               var arrayOfFoo = new[] {new Foo() {Name = "aaa"}};
                               var arrayOfBar = Mapper.DynamicMap<Bar[]>(arrayOfFoo);

                               Assert.AreEqual(arrayOfFoo.Length, arrayOfBar.Length);
               }

                public class Foo
                {
                                public string Name { get; set; }
                }

                public class Bar
                {
                                public string Name { get; set; }
                }

jbogard:
Fixed!

PR #781, opened 2015-07-10
Dynamically map arrays

lbargaoanu:
#628 
Not sure what the best solution is but the reasoning is below.
Creating the map for the arrays in DynamicMap is not helpful, we need only the mapping for the element types. 
After dynamically creating the TypeMap, the cache had null for that TypePair because initially the mapping didn't exist, so clear that entry.


jbogard:
Merged!

Issue #625, opened 2014-11-10
ConstructUsing does not throw exception or gives warning when null returned


TimMahy:
When a function is used to construct the mapper which returns null, a NullReferenceException is throw without any further information. Desired would be to throw an exception with verbose logging stating that a custom ConstructUsing method was used which returned null leading to this exception.

jbogard:
Fixed!

PR #782, opened 2015-07-13
Check if we are able to construct the destination object

lbargaoanu:
#625
Not sure if this is the right place to check, but the tests pass :).


jbogard:
Merged!

Issue #618, opened 2014-10-30
Mapper.CreateMap<Self, Self> ignores nullability rules on collection properties


Serenyrddraig:
This issue arose because my developer reused a Type between two DTOs. Consider the model

        public class Container
        {
            public List<string> Items { get; set; }
        }
        class ContainerDto
        {
            public List<string> Items { get; set; }
        }
        class Holder
        {
            public Container[] Containers { get; set; }
        }
        class HolderDto
        {
            public ContainerDto[] Containers { get; set; }
        }

The rule I must enforce is that if the Items collection in the source is null, it must not be null in the destination. I can do this in a variety of ways, some of which are shown in the following test.

        [Fact]
        public void test_mapping_withNullToInstanceProperty()
        {
            Mapper.AllowNullDestinationValues = false;
            Mapper.CreateMap<Holder, HolderDto>();
            Mapper.CreateMap<Container, ContainerDto>()
                //.ForMember(d =>d.Items, o=>o.NullSubstitute(new List<string>()))
                //.ForMember(d => d.Items, o => o.MapFrom(s => s.Items ?? new List<string>()));
                ;

            var from = new Holder() { Containers = new Container[] { new Container() { Items = null } } };
            var to = Mapper.Map<HolderDto>(from);
            Assert.NotNull(to.Containers.First().Items);
        }

However, the situation that my developer had created effectively is

        [Fact]
        public void test_mapping_withNullToInstanceProperty()
        {
            Mapper.AllowNullDestinationValues = false;
            Mapper.CreateMap<Holder, Holder>();
            Mapper.CreateMap<Container, Container>()
                //.ForMember(d =>d.Items, o=>o.NullSubstitute(new List<string>()))
                //.ForMember(d => d.Items, o => o.MapFrom(s => s.Items ?? new List<string>()));
                ;

            var from = new Holder() { Containers = new Container[] { new Container() { Items = null } } };
            var to = Mapper.Map<Holder>(from);
            Assert.NotNull(to.Containers.First().Items);
        }

No combination of techniques I have tried produce the desired result, i.e. the destination Items collection remains null after mapping. The configuration does not assert as invalid.

Sadly for me this occurred in a very complex object where the outermost types were different but this situation occurred in properties within those objects.

lbargaoanu:
[...]

PR #784, opened 2015-07-13
Only use the AssignableArrayMapper if the elements of the arrays are …

lbargaoanu:
…not explicitly mapped
#618


jbogard:
Merged!

Issue #565, opened 2014-07-30
Guidance on configuration is fractured


breyed:
The various tidbits of documentation regarding configuration leave a muddled message of the best practice that developers should follow.

----

The [Getting Started Guide](https://github.com/AutoMapper/AutoMapper/wiki/Getting-started) only mentions the static `Mapper` method and shows this example:

    Mapper.CreateMap<Order, OrderDto>();

The guide doesn't mention `Mapper.Initialize`; however, the summary for that method says it is "the preferred means to configure AutoMapper."

----

The [Configuration](https://github.com/AutoMapper/AutoMapper/wiki/Configuration) shows this code in the `OrganizationProfile` sample:

    protected override void Configure() 
    {
      //Put your Mapper.CreateMap... Etc.. here
    }

However, the summary for `Profile.Configure` says, "Avoid calling the AutoMapper.Mapper class from this method." So the comment to call `Mapper.CreateMap` in `Configure` looks like bad advice.

----

What is really needed is an API and documentation that is both easy and optimized at the same time. And of course it should be consistent. Hopefully the API is already sufficient; if not, consider adding the necessary shortcuts to make the simple, common case easy but still optimized

For example, consider an `Initialize` overload that takes an unnamed `Profile` directly, so that the quick-start sample code can be like the following, with the result being an AutoMapper configuration that is validated and sealed automatically.

>Call this method when your project starts up.

    Mapper.Initialize(new AppAutoMapperProfile);

>`Initialize` references a profile that you define like this:

    class AppAutoMapperProfile : Profile
    {
        protected override void Configure() 
        {
            CreateMap<Order, OrderDto>();
        }
    }

jbogard:
Done!

PR #788, opened 2015-07-16
Initialize with profile bug 565

lbargaoanu:
#565


jbogard:
Sooooo I don't like the implementation. If you're using a profile you should have more than one, otherwise just use cfg.CreateMap.

lbargaoanu:
Roger :)

jbogard:
I think this is more of a documentation issue to be more explicit. Long term it's more of an API issue.

lbargaoanu:
OK, so better to close it for now?

jbogard:
Yep!

Issue #555, opened 2014-07-10
Cannot map type property resulting from a where


llauck:
After applying a Where condition to a type's property that is a collection type the AutoMapper throws an exception of type "AutoMapper.AutoMapperMappingException" with the inner exception "Unable to cast object of type 'WhereListIterator1[System.Int32]' to type 'System.Collections.Generic.IList'".

The following code reproduces the issue:

    class Program
    {
        static void Main(string[] args)
        {
            Type1 type = new Type1()
            {
                Id = 1,
                ListProperty = new List<int>() { 1, 2, 3 }
            };

            Type2 type2 = new Type2()
            {
                Id = 2,
                ListProperty = new List<int>() { 4, 5, 6 }
            };

            type2.ListProperty = type2.ListProperty.Where(a => true);

            Mapper.Configuration.CreateMap<Type1, Type2>();
            Mapper.Map<Type1, Type2>(type, type2);

        }
    }

    public class Type1
    {
        public int Id { get; set; }

        public IEnumerable<int> ListProperty { get; set; }
    }

    public class Type2
    {
        public int Id { get; set; }

        public IEnumerable<int> ListProperty { get; set; }
    }

We are using a workaround in our internal wrapper to the AutoMapper, the solution is as follows:

    public static TDestino Mapear(TOrigem modelo, TDestino destino)
    {
	    var propsRO = destino.GetType().GetProperties()
            .Where(p => typeof(IEnumerator).IsInstanceOfType(p.GetValue(destino)));

	    foreach (var prop in propsRO)
	    {
            var tipo = prop.PropertyType.GetGenericArguments()[0];
            var tipoDaLista = typeof(List<>).MakeGenericType(tipo);
            var lista = Activator.CreateInstance(tipoDaLista);
            var listaOriginal = prop.GetValue(destino);

            while(((IEnumerator)listaOriginal).MoveNext())
            {
                ((IList)lista).Add(((IEnumerator)listaOriginal).Current);
            }

		    prop.SetValue(destino, lista);
	    }

	    return Mapper.Map(modelo, destino);
    }
Before this issue we are using the AutoMapper 2.1.267 and this behavior starts to happen after we updated to the AutoMapper 3.2.1.

jbogard:
Fixed!

PR #789, opened 2015-07-16
Only apply EnumerableMapper to IList existing destination

lbargaoanu:
if we already have a destination, than it must be IList
#555


jbogard:
Merged!

Issue #542, opened 2014-06-25
Related to issue #91, Source has no property named Type, Target does have property name Type


greyalien007:
If your source object does not have a property named "Type" but your target does have a property named "Type" TypeMap.GetUnmappedPropertyNames does not return this as unmapped which can cause weirdness.  If your target Type property happens to be of string type you end up with the Type.ToString() (guess) filling the property during mapping.  If it's not a string and there's no valid conversion, it blows up.  I added the following code to my project to avoid the issue, it may add clarity if this paragraph doesn't make sense.

            if (sourceType.GetProperty("Type") == null && destinationType.GetProperty("Type") != null)
            {
                expression.ForMember("Type", opt => opt.Ignore());
            }

This is not blocking me, no rush.

jbogard:
Fixed!

PR #792, opened 2015-07-17
Exclude object methods

lbargaoanu:
#542


lbargaoanu:
Not ready yet :)

jbogard:
So just that we're clear here, this means that you can't map GetType any more, right?

jbogard:
Merged!

lbargaoanu:
I think so :). I know there was some related issue, I'll try to find it and write another test.

Issue #670, opened 2015-01-28
Does AutoMapper support conventions?


Sebazzz:
In our domain model we may have a class like this:
```C#
class Question {
     // ...
     public virtual ICollection<QuestionLocalization> Localizations {get;set;}
     // ...
}
class QuestionLocalization {
    public string Text {get;set;}
}
```
And a DTO like this:
```C#
class Question {
   // ... 
   public string Text {get;set;}
```

And we map this using AutoMapper like:
```C#
.ForMember(d => d.Text, a => a.MapFrom(e => e.Localizations.First(loc => loc.Culture == CultureContext.CurrentCulture).Text)
```

Which works great!

Is there any way to plug in a convention in AutoMapper that allows automatic conversion like this to happen, while keeping the possibility to keep the Expression Tree like in this example?

jbogard:
I wish it did. This is my big 4.0 push, to add conventions.

Sebazzz:
Ah, right. I'd be happy to contribute if you need any help.

jbogard:
Any more examples of what you'd like to do? That would help in figuring out what needs to be supported.

Sebazzz:
Not entirely the same problem as this one, but something like this as well: Automapping many types to a default DTO. For example, for a drop down list, one might create a 'IdNameProjection' class:
```C#
public class IdNameProjection {
    public int Id {get;set;}
    public string Name {get;set;}
}
```
Sorry for the vague explanation, I hope it makes sense.

TylerCarlson1:
@Sebazzz I submitted #983 with an example to make a convention that covers your example in using Localization info.

PR #793, opened 2015-07-19
Add Profile Fluent API Convention

TylerCarlson1:
Keeps old API functionality tack, except fixed Name comparisons to always be case sensitive by default, except when using constructor parameters.
If use nothing but old API it will all work the same.

New Fluent API now uses ```AddMemberConfiguration()``` which defaults to just checking single names of properties vs source Properties, Fields, Methods, and Extension Methods.

Old API default uses
```AddMemberConfiguration().AddMember<NameSplitMember>().AddName<PrePostfixName>(_ => _.AddStrings(p => p.Prefixes, "Get")).SetMemberInfo<AllMemberInfo>()```
Also if you don't set any configs it will default to this.

```AddMember<IChildMemberConfiguration>()``` has DefaultMember and NameSplitMember classes.
Can input action against config for parameter.
DefaultMember just compares the source and destination, without any modification.  Is added by default
NameSplitMember Splits names based on INamingConvention for source and destination using Pascal format as default.  NOTE:  INamingConvention has ```string ReplaceValue(Match match)``` in order to change case when re-assembling string, so that case sensitive name compares work.  When it finds matches it calls the parent.MapDestinationPropertyToSource() which will recursively call every NameMapper with the second part of name snippet.

```AddName<ISourceToDestinationNameMapper>()``` has DefaultName, PrefixPostFix, ReplaceName classes.
Can input action against config for parameter.
DefaultName compares name by Case, except when constructor parameters are passed then it's case insensitive.  Is added by default.
PrefixPostFix same as functionality before except string compare is case sensitive.
ReplaceName does both Alias and ReplaceMember functionality, but string comparison is case sensitive

```SetMemberInfo<IGetTypeInfoMembers>()``` has AllMemberInfo, FieldPropertyMemberInfo, and MethodsMemberInfo classes
AllMemberInfo gets public fields, properties, methods, and extension methods.  Is set to default.
FieldPropertyMemberInfo is self explanatory.
MethodsMemberInfo is just regular methods, not extension methods.
Note: Setting this sets what source member info you get to compare against the source for all Members and Name mappers for the entire configuration.

You can create you own classes at each level to inject whatever kind of functionality you want when making conventions.  And you can exclude any functionality you don't want by not adding it.

If you don't use the old API and use ```AddMemberConfiguration()``` the default won't take effect and you can use whatever convention you want.  You can add as many conventions as you want and they will all be ran until one hits.

Profile also has ```AddConditionalObjectMapper()``` which makes a conditional object map that when hit will get member mappings from the profile that created it.  Conditions are against the Resolution context, but there's an extension method called Where that takes comparison against the source and destination type for easy convention setting.

Example: (In Pull Request)
```
public class ConvensionTest : AutoMapperSpecBase
    {
        public class Client
        {
            public int ID { get; set; }
            public string Transval { get; set; }
        }

        public class ClientDto
        {
            public int ID { get; set; }
            public string ValueTransfer { get; set; }
            public string val { get; set; }
        }

        public class ToDTO : Profile
        {
            {
                AddMemberConfiguration().AddName<PrePostfixName>(
                        _ => _.AddStrings(p => p.Postfixes, "Transfer")
                            .AddStrings(p => p.DestinationPrefixes, "Trans"))
                            .SetMemberInfo<FieldPropertyMemberInfo>();
                AddConditionalObjectMapper().Where((s, d) => s.Name == d.Name + "Dto");
            }
        }
        public class FromDTO : Profile
        {
            {
                AddMemberConfiguration().AddName<PrePostfixName>(
                        _ => _.AddStrings(p => p.DestinationPostfixes, "Transfer")
                            .AddStrings(p => p.Prefixes, "Trans"))
                            .SetMemberInfo<FieldPropertyMemberInfo>();
                AddConditionalObjectMapper().Where((s, d) => d.Name == s.Name + "Dto");
            }
        }

        [Fact]
        public void Fact3()
        {
            Mapper.Initialize(cfg =>
            {
                cfg.AddProfile<ToDTO>();
                cfg.AddProfile<FromDTO>();
            });

            var a2 = Mapper.Map<ClientDto>(new Client() { Value = "Test", Transval = "test" });
            a2.ValueTransfer.ShouldEqual("Test");
            a2.val.ShouldEqual("test");

            var a = Mapper.Map<Client>(new ClientDto() { ValueTransfer = "TestTransfer", val = "testTransfer" });
            a.Value.ShouldEqual("TestTransfer");
            a.Transval.ShouldEqual("testTransfer");
        }
    }
```
Conditional Object mapper will use Profile generated, so the prefix and postfix are separated by if they are mapping to or from the Dto.

I'm not 100% satisfied with the Fluent API's naming, but the point is you can now add conventions and you can add or remove any functionality you want at any level of generating the property maps.
Also there's type map conditions could be better, it's just there's no set way to do them now, so I'm using ```Func<ResolutionContext,bool>``` as a base for comparison.

Also dynamic map might not be needed anymore since you can just do an object mapper convention to map any type to any type and you can specify special Property Map conventions as well on top of that.


jbogard:
OK I think this one will need to wait until I get this initial 4.0 release out the door - I want to put in something bigger too to get rid of Configure and just have the constructor mapping. The second thing is to have profiles really just "gather" configuration, then have them applied later.

TylerCarlson1:
So like gathering from things like ```IMapFrom<T>, IHaveCustomMappings``` interfaces?
Also not sure what you mean by have the constructor mapping.

jbogard:
Just that you have to override a protected method. I'd rather have configuration done in the constructor of a Profile, not in some method.

TylerCarlson1:
Oh that's not as bad as I thought it might be.
That and type map scanning should be fairly easy by comparison.

TylerCarlson1:
Added development changes and changed how MemberInfo selection works to use conventions more attune to Fixie.
Oh and updated to newer version of Fixie.  (I can finally click the tests and they will re-direct me to the code using MSTest now)

TylerCarlson1:
Added `MapToAttribute` to support #805

Instead of using a dictionary I use base class `SourceToDestinationMapperAttribute` that I pass all the type data that you could possibly use to determine a match, so that users can create their own and put whatever magical logic they want to use.
Also it's easier to create more custom type logic like #670 because you will know source and destination type and members information.

It's like injecting custom functions into the array to look for, except in easy to use Attribute form.

Also this is only for sub class Attributes.  No Attributes on the class level will be used.  Only Properties, Fields, Constructors, ect.

Can add class attribute but that will have a different way of working.

jbogard:
OK so I want to put this in - but slight problem, all the file changes are in the wrong folder (AutoMapper.Net4). Can you revert the changes to the .NET4 csproj file, and just copy all those files you changed to the AutoMapper folder, overwriting the files there?

TylerCarlson1:
Ok I'll try to move the files over, later today hopefully.
I think I did something just to get the thing to compile and run tests.

Be nice to have some form of feedback other than, hey the files are in the wrong place lol.

jbogard:
Well the problem is all the files look new, so I can't actually see what changed (check out the diff tab to see)

TylerCarlson1:
That makes perfect sense.  
Main one that I'm concerned about is Naming Conventions.  I added a new field that's a function call to convert to that type.  It's used so that some_property will only find a match with SomeProperty instead of SomeProperty or someproperty, depending on which one came first.  It's like a translate to capitalize the first letters of each match in RegEx.

TylerCarlson1:
The files in question have no conflicts originals but removed them and updated .NET4.csproj.  Also merged to development stream
I actually have one failing test when it comes to naming conventions.
Will fix it soon.

TylerCarlson1:
Got test fixed, but last change shows my concerns with this.  Conventions follow more specific rules and case matters A LOT with things like lower case to and from Pascal Case.

This should now be done for review and merging.

TylerCarlson1:
Merged to dev.  Thought you might ask me too

jbogard:
Yep. This is pretty much the last thing on my plate now that I've finally figured out all that TypeMap resolution stuff I think in #941

TylerCarlson1:
If you have any questions feel free to ask.  Hopefully remember what all I did.

Issue #815, opened 2015-08-06
Strange issue with Automapper 4 and overloaded constructors


dplaskon:
Since moving to Automapper 4, I've seen some strange behavior in mapped objects that have overloaded constructors.

Here is a (somewhat) simple repro to show the issue:

```vb
        <Test()>
        Public Sub Can_Reproduce_Automapper_Constructor_Issue()

            Mapper.CreateMap(Of Tester, Tester2)()

            Dim t As New Tester()

            Dim dest = Mapper.Map(Of Tester2)(t)

        End Sub

        <Serializable()>
        Public Class Tester

            Public Sub New()

            End Sub

            Public Property UserId As Integer


        End Class

        <Serializable()>
        Public Class Tester2

            Public Sub New()

            End Sub

            Public Sub New(ByVal userId As Integer,
                           ByVal active As Boolean)
                Me.UserId = userId
                Me.Active = active
            End Sub

            Public Property UserId As Integer

            Public Property Active As Boolean

        End Class
```

The following exception is thrown:

> AutoMapper.AutoMapperMappingException : Missing type map configuration or unsupported mapping.

> Mapping types:
> Tester -> Boolean
> MyProject.IntegrationTests.TasksTests.TempTests+Tester -> System.Boolean

> Destination path:
> Tester2

> Source value:
> MyProject.IntegrationTests.TasksTests.TempTests+Tester

Removing the overloaded constructor results in successful mapping.

Any ideas?

TylerCarlson1:
Submitted a PR to fix this in #816.

Issue is now have can resolve for constructor parameters, but they aren't accounted for correctly.

dplaskon:
Thanks :)

PR #816, opened 2015-08-06
Fix for #815

TylerCarlson1:
-CanResolve needs to be checked before all instances of Ctor Mapping
-CtorParamOveride needs to set CanResolve to true


jbogard:
Merged!

arturoribes:
Is this a fix that AutoMapper developers need to implement or is there a new version to update to?

Issue #840, opened 2015-08-19
Mapping and Projecting enum strange behavior


brenovieira:
It's quite a question father than an issue, because I don't know the right behavior. I'm using AutoMapper 4.0.4.

I'm trying to map and project (ProjectTo<>) some enum properties and the only working configuration I found was something like:
```
Mapper.CreateMap<Source, Destination>();
var m = Mapper.CreateMap<SourceType, DestinationType>();
m.ProjectUsing(s => (DestinationType)s);
m.ConvertUsing(s => (DestinationType)s);

//Source and Destination are classes
//SourceType and DestinationType are enums and properties of Source and Destination
```

- Using ```Mapper.Initialize(cfg => ...)``` or ```Mapper.AddProfile<Profile>()``` have the same behavior

Is there another way (simpler) to achieve that (mapping and projecting enums)?
I'm asking this because I have several enums and if I just map them, there is no need of creating enum mapping, but projecting them doesn't work "automatically".

Please see a gist I created here: https://gist.github.com/brenovieira/106b904f3899713e090d

jbogard:
Well, to use projection, the LINQ provider needs to understand that ProjectUsing piece. The trick is to find the correct expression that EF I'm assuming will understand.

What behavior are you seeing?

brenovieira:
The behavior I didn't except is based on 2 points:

1. To map an enum ```SourceType``` to another enum ```DestinationType```, I don't need to call ```CreateMap<SourceType, DestinationType>()```. AutoMapper "knows" this mapping by default.
You said that long time ago at [StackOverflow](http://stackoverflow.com/questions/11265829/how-can-i-map-between-two-enums-using-automapper/11268711#11268711), so I didn't created it at first.
In case of mapping a class ```Source```to another class ```Destination```, we must call ```CreateMap<Source, Destination>()```.
But to use projection the behavior is the opposite:
To enums, I must call ```CreateMap<SourceType, DestinationType>().ProjectUsing(s => ...)```, but to project classes I don't need to call ```.ProjectUsing(s => ...)```, only ```CreateMap<Source, Destination>()``` works.

2. To just map, I can configure AutoMapper like this:
````
Mapper.CreateMap<Source, Destination>();
````
And both mappings work: ```Mapper.Map<Source, Destination>(source);```and ```Mapper.Map<SourceType, DestinationType>(sourceType);```.

To just project, as I said in 1., I can configure it like this:
```
[...]

PR #851, opened 2015-08-22
Set projection and conversion with same expression

jasonholloway:
As per #840 (in part!), when using QueryableExtensions it would be nice if declared projection rules were picked up for normal mapping too. 

The API raises expectations that this should happen already, as ProjectUsing returns void and doesn't allow any run-on clauses: it's final, and seems like it declares everything that needs to be declared. When Mapper.Map is then inadvertently used alongside projections, and it throws an exception, the innocent user may blame AutoMapper and convince himself that something is broken, rather than suspecting his own misunderstanding of the API.

Therefore I've changed the comment on ProjectUsing to be a little more explicit, and also exposed a new ProjectAndConvertUsing method that does as you'd expect.

### DIFF
--- a/src/AutoMapper/IMappingExpression.cs
+++ b/src/AutoMapper/IMappingExpression.cs
@@ -130,16 +130,16 @@ namespace AutoMapper
         IMappingExpression<TSource, TDestination> WithProfile(string profileName);
 
         /// <summary>
-        /// Skip member mapping and use a custom expression during LINQ projection
+        /// Skip member mapping and use a custom expression for LINQ projection only
         /// </summary>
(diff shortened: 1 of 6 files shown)


jbogard:
So the way I think I'm going to handle this is for the ProjectUsing, I'll just also compile that for ConvertUsing, like MapFrom does.

jbogard:
This is what I wound up doing, having the same behavior for MapFrom, ConstructProjectionUsing etc

jasonholloway:
Looks good! Slightly more straight forward than the whole setter-class thing :)

PR #852, opened 2015-08-23
Small changes to expression-binding behaviour

jasonholloway:
Custom projections weren't being applied to root objects, only to children via the binding mechanism. I've moved the code for rendering the custom projection into a clause in MappingEngine.CreateMapExpression, so it can apply itself either through binding (via MappedTypeExpressionBinder) or up front, at the beginning of expression mapping.

Also lowered the priority of AssignableExpressionMapper, which wasn't giving custom projections chance to kick in under certain circumstances.


jbogard:
It looks like after rebasing this has a failing test:

```

Test 'AutoMapper.UnitTests.Query.SourceInjectedQuery.Shoud_support_IEnumerable_result' failed: System.ArgumentNullException
Value cannot be null.
Parameter name: source
   at System.Linq.Enumerable.ToArray[TSource](IEnumerable`1 source)
   at lambda_method(Closure , Source )
   at System.Linq.Enumerable.WhereSelectArrayIterator`2.MoveNext()
   at System.Collections.Generic.List`1..ctor(IEnumerable`1 collection)
   at System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
   at AutoMapper.UnitTests.Query.SourceInjectedQuery.Shoud_support_IEnumerable_result() in C:\dev\automapper\src\UnitTests\Query\SourceInjectedQuery.cs:line 81
```

jbogard:
Ping

jasonholloway:
Sorry for the delay, am looking into this now...

jasonholloway:
Right: the `IExpressionBinder` classes are inconsistent in binding with null guards. When `EnumerableBinder` is made more prominent, its deficiency is exposed. Nothing to do with `SourceInjectedQuery`, which just happened to test an aspect previously uncovered.

As things are, each `IExpressionBinder` opaquely delivers a `MemberAssignmentExpression`, ready to be centrally bound. This makes it impossible for the central binding function to apply null guards in general, which would seem to be the best behaviour here.

So, I'm going to change the `IExpressionBinder` interface to return a population expression instead. Slightly worried I'm opening a big tin of worms however...

jasonholloway:
Well, done as promised. Null-guard generation is now done centrally in `MappingEngine.CreateMemberBindings()`, and the `IExpressionBinder` classes have become `IPropertyProjector` ones, as they no longer do the bindings themselves.

The problem in the end was that the original failing test didn't set the `AllowNullCollections` option, but some mysterious malfunction in the expression binders was letting it get away with it.

jbogard:
Well nuts, can you fix the conflicts?

jasonholloway:
Done! There was a bit of overlap on MappingEngine.CreateMapExpression with #933, which was basically doing the same thing with centralising custom projections. I've tidied that area up a bit to make what it does a bit clearer.

jasonholloway:
Also should mention that Fixie 1.0.0.29 always crashes the VS test runner on my laptop - don't know if that's just me. I've updated to 1.0.0.3 on the two test projects and it works fine (for me).

jasonholloway:
And... there's a new failing test, though I don't think it's entirely my doing. SourceInjectedQuery.Should_support_enumerable_return_type now fails, and passed before, but a bit of investigation shows it shouldn't have passed in the first place. 

As things are (that is, on develop) if you change the test's result variable to materialize to an array (i.e. stick ToArray() on the end in place of First()), then it no longer passes, though I'm pretty sure that's still in the spirit of the test.

jasonholloway:
And... I seem to have committed changes to every file in the project... somehow...

jasonholloway:
Basically I need to separate out the merge with develop to a separate commit, but I'll do that tomorrow... too late now.

jbogard:
No worries, I pushed out 4.1 just now and marked this for 4.2. I want to get this and all that expression mapping stuff in for that release.

jasonholloway:
Hi again, when do you think you'll be pushing out 4.2? I'll make sure this gets sorted out in time.

jasonholloway:
Actually, I'm a bit baffled. What I wanted was to make the positive changes of my last commit more obvious by separating them from the merging in of develop, which obviously adds lots of noise and obscures them. But - because these positive changes were made as part of the merge, to solve its conflicts, they can't be separated out, can they? Hmmm.

jasonholloway:
Redone this to be bit more to the point - see #1130

PR #871, opened 2015-09-04
Lazily mapped queries

gentledepp:
Hi!

I am pretty excited about this.
AutoMapper already supports QueryableExtensions.
using Map<Dto,Entity().ProjectTo<Dto>() you can map an IQueryable<TEntity> to an IQueryable<TDto>
However as this mapping is made eager, applying any additioanl filters to the resulting IQueryable<TDto> is not possible.
I managed to create an MappingQueryProvider that does this lazily.
Please have a look at the unittest "AutoMapperSamples.EF.EffortQueryTests.Effort_FilteByMappedQuery()" to see the difference.

This lazy evaluation also allows to expose the IQueryable<TDto> in an OData enabled WebApi Controller. I made kind of a proof-of-concept in the project AutoMapperSamples.OData.
There you can see that I can run OData queries on the Dto-Type, and this Expression is then mapped by your ingeious AutoMapper to an EF-compatible query. I love that!

I know that it still is kind of a raw diamond, but I think it has a lot of potential!
Until now exposing an OData Endpoint that allows to execute the queries on the database directly always required you to not use DTOs at all and we both know this is horrible.

So what do you think!

One thing I have to mention.
In order to support OData, I had to add a small hack to AutoMappers MemberAccessQueryMapperVisitor. The reason is, that the WebApi.OData wraps all constant expressions into an internal type "System.Web.Http.OData.Query.Expressions.LinqParameterContainer+TypedLinqParameterContainer" as that improves EF query performance.
AuthMapper did not find a PropertyMap for it and threw an Exception. Do you have an idea how I/we could solve this?

Also this feature is not EF dependant. It should work with NHibernate as well. You'd just need to add another nuget package "Pathoschild.WebApi.NHibernate-OData" as WebApi.OData creates kind'a complex expressions that NHibernate cannot handle.
This is described in: https://github.com/Pathoschild/webapi.nhibernate-odata 

So what do you think?! How should I proceed? :)


jbogard:
Adding @lbargaoanu and @TylerCarlson1

TylerCarlson1:
I would suggest looking into UseAsDataSource()
https://github.com/AutoMapper/AutoMapper/wiki/Expression-Translation-(UseAsDataSource)

It's pretty much exactly what you are doing and it does support OData to my knowledge, although it doesn't say it explicitly.  Someone else did the OData support and the expression visitor handles OData wrapper object.  I just appended it onto my expression visitor to convert things and have never verified it myself.

This implementation instead of calling Mapper.Map<> it uses the Expression visitor the object mapper uses directly do convert the expressions.

gentledepp:
ok - I'll extend my testsuite with "UseAsDataSource" and will ping back to you.
However - there are two features that I am especially keen about:
1) being able to access the IEnumerable<TDto> that resunts from the query BEFORE it is sent to the client. This allows me to apply additional filtering or setting properties that I could not using AutoMapper
2) being able to get the exceptions that may come up because I forgot to map something. In the worst case, these would just be swallowed by WebApi resulting in an unspecific "500 internal server error"
I'll ping back to you...

gentledepp:
By the way:
Creating a Map like this is insufficient, as AutoMapper - for some reason - does not reverse-map from FullName to Name
    
                    cfg.CreateMap<OrderDto, Order>()
                        .ForMember(d => d.Name, opt => opt.MapFrom(s => s.FullName))
                        .ReverseMap(); // reverse map added
I have to explicitly create the reverse mapping by calling:
    
                cfg.CreateMap<OrderDto, Order>()
                    .ForMember(d => d.Name, opt => opt.MapFrom(s => s.FullName));
                cfg.CreateMap<Order, OrderDto>()
                    .ForMember(d => d.FullName, opt => opt.MapFrom(s => s.Name));

Is that by intention, or is it a bug?

gentledepp:
also - as soon as I add the property "public virtual CustomerDto Customer" to my "OrderDto", I get an exception like:
The specified cast from a materialized 'System.Data.Entity.DynamicProxies.Order_4CFE0FDCADE638F08621960EF8AFFD423ACD6D4137C98BEB0CFC6AC0FE831FB1' type to the 'System.Linq.IQueryable`1[AutoMapperSamples.EF.Model.Order]' type is not valid.
What am I doing wrong here? I just don't get it. :-|

gentledepp:
So I added some changes that you can best see and understand in the "OrdersController"
1) allowing to add an optional exception handler in order to
2) allowing to add additional ExpressionVisitors. In my case this is required to support OData with EF as EF does not allow a call to Queryable.LongCount. That however is issued as soon as you request the $inlinecount in OData, which in turn is very important for framerworks like Kendo UI which use that information to support data paging.
3) another feature is allowing to intercept the IEnuemrator<T> that results of the expression. This allows you to even further filter the results and apply whatever changes you'd like.

What I have still problems with is the following:
a) I would like to support OData "Select" queries. In that case however, AutoMapper blows up in "SourceInjectedQueryProvider.InvokeSourceQuery": 
    
    Additional information: The specified cast from a materialized     'System.Web.Http.OData.Query.Expressions.SelectExpandBinder+SelectSome`1[AutoMapperSamples    .EF.Dtos.OrderDto]' type to the     'System.Linq.IQueryable`1[System.Web.Http.OData.Query.Expressions.SelectExpandBinder+SelectSo    me`1[AutoMapperSamples.EF.Dtos.OrderDto]]' type is not valid.


b) Also I would like to use the DynamicProxy feature of EF. However as soon as I change
<pre>
public class Order
    {

        [Key]
        [Required]
        [DatabaseGenerated(DatabaseGeneratedOption.None)]
        public Guid Id { get; set; }
        public DateTime Ordered { get; set; }
        public double Price { get; set; }
    }
</pre>
to
<pre>
public class Order
    {
        public virtual string Name { get; set; }

        [Key]
        [Required]
        [DatabaseGenerated(DatabaseGeneratedOption.None)]
        public virtual Guid Id { get; set; }
        public virtual DateTime Ordered { get; set; }
        public virtual double Price { get; set; }
    }
</pre>
I get the exception 
<pre>
    The specified cast from a materialized 'System.Data.Entity.DynamicProxies.Order_4CFE0FDCADE638F08621960EF8AFFD423ACD6D4137C98BEB0CFC6AC0FE831FB1' type to the 'System.Linq.IQueryable`1[AutoMapperSamples.EF.Model.Order]' type is not valid.
</pre>

c) the last missing thing then would be "$expand" and honestly I do not know whether I would like to support that. I would love to hear your opinion about that.

So what do you think?

TylerCarlson1:
Looks like you added a lot of versatility to UseAsDataSource().  It was pretty bare bones to begin with.

Few questions
1) The enumerable for additional filtering.  Will that effect Skip and Take?  Like if I take 5 and after filter only returns 4 of the 5.  Should it be more like pipe where it's an function on each object setting properties and such?
Also this is a feature requested for Project To functionality.  I don't know if @jbogard might want to encorperate this into project to instead of UseAsDataSource.
2) Exception handling looks like a great idea.  I'm not too familiar with web myself, so don't take my word on it.

3) Expression visitors for before and after map look like a good idea as well.  It kind of follows what 1) does, but if you can get a conversion from using an expressionvisitor to using an expression to replace, you might get a BeforeProjection and AfterProjection equivilent that can be used in Project To.

4) Select statement not working.  You might have to troubleshoot it in MappingVisitor.VisitExtension and write a case for it to work.  That's where I put the OData "hack" for UseAsDataSource().  probably have to convert whatever that is into a Select statement in order for it to work.

5)public virtual CustomerDto Customer, CustomerDto shouldn't have a property referencing back to Orders, cause that's going to cause problems with an infinite loop.  That might part of the problem it doesn't work.

6)For Dyanamic Proxies I think the system can't translate it over, but I don't know why.  And why would you want that anyways?  It's completely disconnected from the data context at that point, so what's the point of keeping it as an object that has a link back to the context?

7) As for expand I think that needs to be extracted from the expression into its own field and use as part of ProjectTo's expand funcitionality.

Overall looks great.  I think most of the features you added, baring the exception handling, should be put into TypeMap.  There have been requests for before map and after map functionality for ProjectTo, so I think they should be implemented at the type map level.

I'm very curious to know what @jbogard thinks of moving these features to TypeMap.  As well as the other features that don't tie directly into OData support.

Oh BTW, thanks for making unit tests for OData scenarios.

gentledepp:
First of all thanks for responding so quickly!
1) You are right, you could affect "Skip" and "Take" with that. I mean you have all objects in memory by then and you can do with it what you want.
e.g. if you skip 50 customers and take 5, you will have 5 in the enumerator.
Then you can of course skip two of them and return only the remaining 3.

As for piping - I do not exactly get the idea of what you mean.
The reason why I hand over the whole IEnumerator<T> is twofold:
a) you could get all of the DTOs, and use their IDs to query additional data from the database with only one query. So calling an "OnEnumerated<T>(T obj) would not work for me
b) you have all of the control over the enumerated elements. So you could - if you are freakin crazy - even create a completely new List<T> and fill it with all stuff you want - thereby ignoring the original enimerator completely - and return that. That gives you a great degree of freedom.

2) If you want to know how it "feels like to be blind", just remove the "OnException" call, comment out the Mapper configuration and see if you can identify the problem by the information the HttpClient.Get() call gives you. Good luck ;-)

3) I dont know where and how to exactly add this. I would be greatful if you could point me towards the right direction.

4) ok. I will dig into that.

5) Seems I was wong. There must have been a bug in EF 6.0.0.0 that is added by Effort. I compiled the current source (6.1.3) and then it suddenly worked. So I updated the nuget package references and now I get my OrderDtos along with all the Customers down to the client side. Awesome! :)

6) This problem has also vanished with 6.1.3 - maybe I did something wrong. Will look into that too.

7) So you mean we should add support for $expand? But what if I do not call $expand now for my OrderDto.Customer reference? In that case I'd like it to be null. Doesn't AutoMapper access that property nevertheless and cause the Lazyloading to kick in?

TylerCarlson1:
1) Pipeing would be replacing Func<IEnumerable<object>, IEnumerable<object>> with Action<object>.  Just expose each object individually and apply some kind of method on it.  Like a foreach statement, just as an extension method so you can fit it in one line.  Can't create a new object, change order, add or remove items.
The problem with having so much power is that it makes it so easy to mess up and not understand why. Also there's no practical use for it, other than to cause weird bugs and do things outside the scope of what AutoMapper does.

3)Maybe this scenario I'll look into.  Still want to hear @jbogard's feedback.  An example would be to look through MemberVisitor I believe.  It has an old and new expression and replaces one with the other.  This one I am less sure about because I'm not sure of how this would be used outside your given scenario.

7) Expand as far as ProjectTo is concerned is an array of strings.  If nothing is expanded array will be empty by default.  If not will have expanded parameters.  Either way will pass into ProjectTo's function that supports expansion and should work without problem.
I'm not sure what all expand does in Projection, since I would assume the conversion to the select statement would expose any property you want to expand anyways.  But it is a case that project has and UseAsDataSource doesn't have, so there might eventually need to be a way to pass the expansions to the Projection statement.  But I don't know if reading from OData is one, because expand is handled by Projection already.
There's no valid benefits to including this feature right now, so I think it's not worth doing at the moment.

gentledepp:
1) I see. So I can understand that this would be too powerful, however I still need to get the whole collection and work on it (not just a single item) in order to do what I need: Fetch additional data from the database in one single additional roundtrip. I will see what I can do..
7) I was working with BreezeJS and I am loving it. It kind of gives you a DBContext on the JavaScript client side. The major problem with it is that it requires you (more or less) to send your entities (and not dtos) to the client as you would loose the powerful querying feature if you didn't.
So why am I telling you all this? Breeze as well supports expansion, but in our project we disabled it (can be one with an additional controller attribute of BreezeJs) as it causes too much trouble on the serverside, like filtering the expanded collections (which is not perfectly possible with EF or NHibernate until today). So I also think it is no so important.
However if it is simply passing strings to another function, I will try it out - thanks anyway.

jbogard:
What are the open questions here? Extending to ProjectTo would be *nice* but perhaps that could wait for V4.2?

src/AutoMapper.Net4/AutoMapper.Net4.csproj
```diff
@@ -49,13 +49,487 @@
...
+      <Link>QueryableExtensions\Impl\ISourceInjectedQueryable.cs</Link>
+    </Compile>
     <Compile Include="..\CommonAssemblyInfo.cs">
       <Link>CommonAssemblyInfo.cs</Link>
     </Compile>
   </ItemGroup>
   <ItemGroup>
-    <Compile Include="..\AutoMapper\**\*.cs">
-      <Link>%(RecursiveDir)%(Filename)%(Extension)</Link>
+    <Compile Include="..\AutoMapper\AliasedMember.cs">
```
jbogard: Can you revert these below changes? The above takes care of "add all the files"

src/AutoMapper.Net4/AutoMapper.Net4.csproj
```diff
@@ -49,13 +49,487 @@
     <Reference Include="System.Core, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089, processorArchitecture=MSIL" />
   </ItemGroup>
   <ItemGroup>
+    <Compile Include="..\AutoMapper\QueryableExtensions\Impl\ISourceInjectedQueryable.cs">
```
jbogard: Revert this one too (see note below)

src/AutoMapper/QueryableExtensions/Impl/QueryDataSourceInjection.cs
```diff
@@ -1,16 +1,74 @@
...
+        /// you do not call "ToList" yourself and therefore cannot catch exceptions
+        /// </summary>
+        /// <param name="exceptionHandler">The exception handler.</param>
+        /// <returns></returns>
+        IQueryDataSourceInjection<TSource> OnError(Action<Exception> exceptionHandler);
     }
 
+    //
+    // Summary:
+    //     Stellt Funktionen zur Auswertung von Abfragen für eine spezifische Datenquelle
```
jbogard: image

src/AutoMapper/QueryableExtensions/Impl/SourceInjectedQuery.cs
```diff
@@ -1,4 +1,6 @@
-﻿namespace AutoMapper.QueryableExtensions.Impl
+﻿using System.Security.Cryptography.X509Certificates;
```
jbogard: What is this used for?

src/AutoMapper/QueryableExtensions/Impl/SourceInjectedQueryProvider.cs
```diff
@@ -1,3 +1,6 @@
+using System.Collections.Generic;
+using System.ComponentModel;
```
jbogard: And this one, what ComponentModel things are getting used?

src/AutoMapperSamples.EF/AutoMapperSamples.EF.csproj
```diff
@@ -0,0 +1,141 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
jbogard: There's an AutoMapper.EF6 project, would any of that belong in there?

src/AutoMapper/QueryableExtensions/Impl/SourceInjectedQuery.cs
TylerCarlson1: I believe that's for WCF Encryption Certificate.  Not sure why that needs to be in solution though.

gentledepp:
Thanks Jimmy for taking the time to investigate.
What is open:

0) Incorporating your feedback and cleaning up my branch... obviously...
1) Question: where should I put the AutoMapperSamples.OData.EntityFrameworkCompatibilityVisitor which is required to convert "Queryable.CountLong" to "Queryable.Count" calls to support OData-EF scenarios?
2) Refactor the "OnEnumerated" method to provide a ReadOnlyCollection<T> to the caller and not allowing to add/remove items returned to the client as this would be too powerfull and therefore error-risky (sorry for my english...)
3) Bugfix: When mapping an expression with a property-path I get an exception. I have a
<pre>
private class Master
        {
            public Master()
            {
                Details = new HashSet<Detail>();
            }
            public Guid Id { get; set; }
            public ICollection<Detail> Details { get; set; }
        }

        private class Detail
        {
            public Guid Id { get; set; }
            public Master Master { get; set; }
        }
</pre>
When I call 
<pre>
 var detailDtoQuery = detailQuery.UseAsDataSource()
                .OnError(ex => exceptions.Add(ex))
                .For<DetailDto>()
                .Where(d => d.Master.Name == "Harry Marry");
</pre>
the fact that both classes have "Name" property seems to confuse the ExpressionMapper.
if I rename the "Master.Name" property to e.g. "FullName" it works. Interesting...  I'm on it.

4) Support projection-expressions with cyclic object graphs. Like the DTO example above. 
Do you have any idea how I could achieve that? I mean mapping an Expression to an DTO-object-tree where the object-tree is cyclic.
This is actually an important feature for me using BreezeJS. :)

5) Nice-to-have:  Support "$extend" for OData and BreezeJS
6) Nice-to-have:  Support "$select" but in order to use this and the "OnEnumerated" feature (see OrdersController) I would have to 
            1) remove the "Select" Expression
            2) create DTOs as mapped with AutoMapper
            3) run the "OnEnumerated" call on the ReadOnlyCollection<DTO>
            4) apply the "Select" expression to the ReadonlyCollection<DTO>
   otherwise I would call the provided "OnEnumerated" action with some anonymous type and it would fail deliberately..
Does my approach sound good to you?

7) Nice-to-have: I am also planning to map-back my DTOs to EF entities using AutoMapper and I mean whole object-graphs. Do you have any hints for me to start?

TylerCarlson1:
3) That's a scenario I should probably take care of.
4) Currently there is no support for this, and for the most part the answer has been, don't make your DTO's circular.  It causes A LOT of problems.
6) That solution might have some problems, if people use it with using Select function call instead of "$select".  If select is the final statement it won't do projection and just map result to result, which will set it to itself.  There's no need to project if you select a subset of properties.
I knew there would be some hiccups with Select and OnEnumerate, but after you explained the steps I realized that's not gonna work the way you want.  The OnEnumerate might have to work on the final return type.  In which case you might be handling anonymous types.

7) I made something for that already. https://github.com/TylerCarlson1/Automapper.Collection
It saves parent and its children, and its children's children, and so on and so forth using the Entity Framework's primary keys as a means to determine uniqueness for updating and inserting Entity Objects.  This is another thing that will not support circular Properties in your DTO object.
I need to update the nuget to support Automapper 4.0.4, but you can fork and update automapper to try it out.

src/AutoMapper/QueryableExtensions/Impl/QueryDataSourceInjection.cs
gentledepp: sry - I'm german :+1:

gentledepp:
Hi Tyler!
@3) the fix is adding the following line to ExpressionMapper:
<pre>

            private PropertyMap PropertyMap(MemberExpression node)
            {
                if (node.Member.IsStatic())
                    return null;

                var memberAccessor = node.Member.ToMemberAccessor();

                // === ADD THIS
                // in case of a propertypath, the MemberAcessors type and the SourceType may be different
                if (memberAccessor.MemberInfo.DeclaringType != _typeMap.DestinationType)
                    return null;

                var propertyMap = _typeMap.GetExistingPropertyMapFor(memberAccessor);
                return propertyMap;
            }
</pre>

TylerCarlson1:
Oh cool.  Got to check that out.  Saves me a lot of time meddling through ExpressionMapper.

gentledepp:
@TylerCarlson1 : Thank you for your hint towards @7
I had the intention do create just the same thing: adding the possibility to inject customer equality-expressions to automapper. A W E S O M E
Only I find it sad that there is so much great stuff which hardly anyone knows about.
There should be blogs full of this to spread the word. Could make a lot of lifes easyer..

gentledepp:
To be a little cleaner, I created an issue for  @3.
https://github.com/AutoMapper/AutoMapper/issues/882
fixed it, added a unittest and made a pull request
https://github.com/AutoMapper/AutoMapper/pull/883

cheers

src/AutoMapper.Net4/AutoMapper.Net4.csproj
gentledepp: Oh man... can't remember having linked anything of that manually to the csproj...

src/AutoMapper/QueryableExtensions/Impl/SourceInjectedQueryProvider.cs
gentledepp: Nothing - removed

src/AutoMapperSamples.EF/AutoMapperSamples.EF.csproj
gentledepp: Hi!
Basically this is only a sample project showing how to use EF with AutoMapper.
The only thing I really added in order to support EF6 ist the
AutoMapperSamples.OData.EntityFrameworkCompatibilityVisitor 
which maps "Queryable.LongCount" methodcallexpressions to "Queryable.Count" expressions as the former are not supported by EF. Should I add the Visitor to AutoMapper.EF?

gentledepp:
So... cleaned things up a bit. I hope you are satisfied.
Kind regards from Vienna, Austria

jbogard:
@gentledepp I'm still looking at having AutoMapper.Collections as part of AutoMapper core. Maybe 4.2?

Also, funnily enough, I'll be in Vienna in a couple of weeks, ha!

gentledepp:
Cool - Why's that? If you need a beer drop me a line on skype ("gentledepp" aka Alexander Marek)

Sorry for having to ask this question: What is "AutoMapper.Collections"? It seems as if I missed something here :)

TylerCarlson1:
Automapper.Collections is Number 7 on my suggestions.  It was originally #528, but since it's over a year old I decided to move it to Automapper.Collections repository to see if I could and to learn GH & nuget.  What @jbogard means by this is the Collections part and custom equality compare part into Automapper, and leave the EF part as a separate dll.
I allways intended it to be a part of Automapper core,  Thought it would be in there by now seeing as how it was a highly requested feature waiting for a PR.  There should be nothing it's conflicting with to put it into 4.2 release.  Might want to submit a fresh PR though for just Automapper.Collections part.

jbogard:
@TylerCarlson1 yep, that way we can isolate that one piece. Hope you had fun learning NuGet :hankey:

TylerCarlson1:
NuGet was at least less of a hassle than learning how to compile AutoMapper in VS2015.  With all the fecal matter I had to go through getting Automapper to work with Fixie tests and such, I can now finally break code up into their own solutions and upgrade to C# 6 compiler features and use Fixie.  Fixie is a godsend if you want to use non-simple class type parameterized tests.

gentledepp:
Ok so what does that mean now for this pull request? :)
Is it in, or is it out? Is something missing?

As for the "Cyclic expression projection" I made a separate pull-request of a proof-of-concept that I'd like you to look at, because it does not seem to be so complicated to me. But maybe I am missing something?

https://github.com/AutoMapper/AutoMapper/pull/884

TylerCarlson1:
I think it's still up in the air right now.  There's been another idea in #880 that shares some similarities to this.  Said he would look over this PR for more similarities.

gentledepp:
Hi folks,
as I promised, I worked on the parameterization and automatic expansion feature on the weekend.
Basically, all I did was adding overrides for the UseAsDataSource() fluent api to support adding parameters and expansion paths and route them to the ProjectionExpression in the SourceInjectedQueryProvider.

With expansion, this seems to work without issues, but I have problems with parameterization.

The reason is, that for all queries that have a (I do not know how you call it, but I call it "final linq operator" as it does not return an IQueryable anymore) "final linq operators" like "First(), Last(), GroupBy()" etc. the "sourceResultType" is not assignable to IQueryable anymore.
<pre>
public TResult Execute<TResult>(Expression expression)
        {
            try
            {
                var resultType = typeof (TResult);
                Inspector.StartQueryExecuteInterceptor(resultType, expression);

                var sourceExpression = ConvertDestinationExpressionToSourceExpression(expression);

                var destResultType = typeof(TResult);
                var sourceResultType = CreateSourceResultType(destResultType);
</pre>

.. which is why "IsProjection<TDestination>(resultType)" returns false and the ProjectionExpression is never created:
<pre>
if (IsProjection<TDestination>(resultType))
                {
                    if (_membersToExpand != null) destResult = new ProjectionExpression(sourceResult as IQueryable<TSource>, _mappingEngine).To<TDestination>(_parameters, _membersToExpand);
                    else destResult = new ProjectionExpression(sourceResult as IQueryable<TSource>, _mappingEngine).To<TDestination>(_parameters, _membersExpressionsToExpand);

                }
</pre>
To clarify this, I added testcases in class 
ParameterizedQueriesTests_with_filter_AsDataSource (as you might guess, I just copied the tests for ProjectTo and changed them to call UseAsDataSource instead)

The following testcase fails, as the call to "FirstOrDefault" leads to a "non-iqueryable" sourceResultType and therefore the "ProjectionExpression" is not created/called
<pre>
        [Fact]
        public void Should_only_replace_outer_parameters_Fails()
        {
            var db = new DB();

            var user = db.Users.UseAsDataSource().For<UserViewModel2>(new {db}).FirstOrDefault(a => a.Id == 2);

            user.position.ShouldEqual(1);
        }
</pre>

if you, however, change the call to the following (such as replace "FirstOrDefault(x)" with "Where(x).ToArray().FirstOrDefault()" it works fine
<pre>

        [Fact]
        public void Should_only_replace_outer_parameters_Works()
        {
            var db = new DB();

            var user = db.Users.UseAsDataSource().For<UserViewModel2>(new { db }).Where(a => a.Id == 2).ToArray().FirstOrDefault();
            
            user.position.ShouldEqual(1);
        }
</pre>


Now I am not absolute sure how I can best overcome this issue.

My first attempt was to actually create an ExpressionVisitor that relaces calls to all "final linq operators" to "Where(x).ToArray().<theOriginalFinalLinqOperator>", but that does not work.

My second attempt was to mimic the internal calls of ProjectionExpression.To<TResult>() in case the resultType was no IQueryable:
<pre>
               object destResult;
                if (IsProjection<TDestination>(resultType))
                {
                    if (_membersToExpand != null) destResult = new ProjectionExpression(sourceResult as IQueryable<TSource>, _mappingEngine).To<TDestination>(_parameters, _membersToExpand);
                    else destResult = new ProjectionExpression(sourceResult as IQueryable<TSource>, _mappingEngine).To<TDestination>(_parameters, _membersExpressionsToExpand);

                }
                else
                {
                    // removed Map() call
                    //destResult = _mappingEngine.Map(sourceResult, sourceResultType, destResultType); 

                    // and try to mimic the internal ProjectionExpression.To<TResult>() calls
                   // does not work as CreateMapExpression requires an IQueryable type as well
                    var mapExpr = _mappingEngine.CreateMapExpression(sourceResultType, typeof(TResult), _parameters);

                    destResult = CreateQuery<TResult>(
                        Expression.Call(
                            null,
                            QueryableSelectMethod.MakeGenericMethod(sourceResultType, typeof(TResult)),
                            new[] { sourceExpression, Expression.Quote(mapExpr) }
                            )
                        );

                }

</pre>

Do you have any time to investigate this? 
Can you give me any feedback what I should try next?

jasonholloway:
Hey gentledepp, glad you're still working on this. Still reading through your post but thought I'd butt in with this link that I recently found: https://msdn.microsoft.com/en-us/library/bb397896.aspx

Basically, it gives names to all the various LINQ operator categories, which is kind of nice. So First(), Last() etc would apparently be 'element operations', in that they return single elements rather than a queryable.

Anyway!

jasonholloway:
And I'm wrong anyway... many different operators don't return queryable, so your name if 'final linq operator' is about perfect. I'll shut up now :)

gentledepp:
No problem, anyways I would appreciate, if @jbogard could possibly give me a hint how he solved this issue with his ProjectTo implementation.
I know an easy answer would be "dude, look at the code!", but it is not such a trivial implementation - at least to me ;-) (fingers crossed....)

jbogard:
@gentledepp how I achieved which part? And it might not have been *me* solving that problem, the LINQ stuff has received a ton of PRs from other people over the years.

gentledepp:
@jbogard  - sorry, let me clarify
My problem at the moment is, that ProjectionExpression needs an IQueryable as first parameter.
So the following statement already works with UseAsDataSource():
<pre>
var user = db.Users.UseAsDataSource().For(new { db }).Where(a => a.Id == 2).ToArray().FirstOrDefault();
</pre>
but this one does not:

<pre>
var user = db.Users.UseAsDataSource().For(new {db}).FirstOrDefault(a => a.Id == 2);
</pre>

this is because, the result-type of the sourceExpression is not an IQueryable, but of type "User" (an element type)

That is basially what I tried to explain in my above comment. :)
Have you any hints that you can give me? We've holiday tomorrow in Austria so I might find some time to work on this. We definetly need that funcionality in our current project.

TylerCarlson1:
How about
```
destResult = new ProjectionExpression(new[] { (TSource)sourceResult }.AsQueryable(), _mappingEngine).To<TDestination>(_parameters, _membersExpressionsToExpand).First();
```

You have a single instance, so you put it into a list and then make it queryable, then you can Project off of it.

Query Provider won't be EF, so it won't expand like with where statement.  You might lose project's ability to retrieve everything in one query call, but it might work.  I don't know if members to expand will work either.

Currently I don't have a solution to project at the last point when it's a single item, as when you enumerate to the end you have a single object and projection is no longer viable.

`private object InvokeSourceQuery(Type sourceResultType, Expression sourceExpression)` is where you would have to edit to fix this though.  What you would have to do is on execute, pass the projection logic in there, so that it gets applied before enumerating the object.  At that point when you check projection for dest type you set to source type, since it's already been translated on enumeration.

gentledepp:
I kind of solved the problem. At least the unittests all work now except of two, but I count myself for only one.

The first one I am not responsible for is AutoMapperSamples.Mappers.QuerySourceInjection.Select_source_items_from_destination
<pre>
[Test]
            public void Select_source_items_from_destination()
            {
                Mapper.Initialize(cfg =>
                {
                    cfg.CreateMap<Destination, Source>()
                        .ForMember(d => d.SrcValue, opt => opt.MapFrom(s => s.DestValue))
                        .ReverseMap();
                });

                IQueryable<Destination> result = _source.AsQueryable()
                    .UseAsDataSource().For<Destination>()
                    .Where(s => s.DestValue > 6);

                result.Count().ShouldEqual(1);
                result.First().GetType().ShouldEqual(typeof(Destination));
            }
</pre>

It says it cannot map from Destination.DestValue to Source.SrcValue.
The problem seems to be in the ".ReverseMap()" implementation: To fix the UT, you simply have to change it to:
<pre>
                    cfg.CreateMap<Destination, Source>()
                        .ForMember(d => d.SrcValue, opt => opt.MapFrom(s => s.DestValue));
                    cfg.CreateMap<Source, Destination>()
                        .ForMember(d => d.DestValue, opt => opt.MapFrom(s => s.SrcValue));

</pre>
The last version of AutoMapper that was released also has this bug which is why I currently do not use "ReverseMap" in our project.
But as I do not quite know the reason why you dropped this feature, or even if that was intended, I do not plan to fix this UT.

The other one i
AutoMapper.UnitTests.Query.SourceInjectedQuery.Shoud_support_const_result().
It leads to a call to 
<pre>
Mapper.CreateMap<Boolean, Boolean>
</pre>
which in turn throws an InvalidOperationException.


Additionally I would like to add a few mor tests. Am I still in time for the next release? We would really like to use those features in our project and I am not keen to maintain a separate AutoMapper package... :)

gentledepp:
Ok guys, solved the explicit expansion also for element results 
<pre>
.UseAsDataSource().For<Dto>(dto => dto.Child1).FirstOrDefault(x => ...);
</pre>

However, several testcases break because of this.
The easiest way to get the build to work is/was to default to the old solution in case no parameters and no  memberexpansions were provideded
<pre>
destResult = _mappingEngine.Map(sourceResult, sourceResultType, destResultType);
</pre>

the three lines of code that provide this fallback are in SourceInjectedQueryProvider
<pre>
if (_parameters != null || _membersToExpand.Length != 0)
                    {
                        sourceExpression = rmv.Visit(sourceExpression);
                    }

</pre>

One complete testclass fails, because 
    MappingEngine.Map
behaves differently than 
    ProjectTionExpression
when it comes to nullable fields.
in
    AutoMapper.UnitTests.Bug.GenericTestExtensionMethods
Child and ChildDto have pairwise properties which are nullable on one end but not nullable on the other.
That is where MappingEngine.CreateMapExpression kind of gives up at the moment.

The other six UTs that fail are in 
    AutoMapper.UnitTests.Query.SourceInjectedQuery
fixture that contain several corner cases such as selecting a "system.object" (i.e. downcasting)
which we do not yet support.

Would that solution (fallback to Mapper.Map<Src,Dest>()) in case no parameters and/or members are provided?

gentledepp:
I also added a new related issue
https://github.com/AutoMapper/AutoMapper/issues/996

TylerCarlson1:
I might be able to move the Null to Non-Null expression visitor logic from Expression Translation and apply it to Project To.  Or if you want to try that you can.  It's the one that uses the Utiary expressions.
Also could you update the branch to devolpment?  Then it can build and run unit tests for the PR and we can see which tests fail from GH.  You don't have to explain them in detail at that point.

gentledepp:
„Update the branch to development“ - I guess you mean point my pull requrst to development insteadof master? I'll try that.:-)

jbogard:
That, and merge develop into your branch.

gentledepp:
created new pull request pointing to develop branch:
#997

PR #877, opened 2015-09-09
Non generic ConstructProjectionUsing bug

lbargaoanu:
It's a small change in MappingExpression. I think GitHub compare gets confused by smth.


jbogard:
What was the bug here?

lbargaoanu:
It needs a Func<SourceType, DestinationType>, not Func<object, object>. It crashed when building the project expression. It replaced the parameter with the dto of SourceType and it passed it to Func<object, object>. I've tried Convert and other things, but to no avail. This seems to work.

PR #884, opened 2015-09-11
Update breeze ef

gentledepp:
DO NOT MERGE THAT IN - it is basically a proof-of-concept I wanted to share with you guys.

@TylerCarlson1 you mentioned that there are a lot of problems when mapping expressions that have a cyclic type-graph as source:
"4) Currently there is no support for this, and for the most part the answer has been, don't make your DTO's circular. It causes A LOT of problems."

Can you, or you @jbogard , give me some more insights on this?

My approach would be the following:
1) detecting cyclic expression-graphs and avoid them
2) in "OnEnumerating" manually fixing up the object-tree.

@1) Basically I was able to achieve just that by adding all previous ExpressionRequests to each created ExpressionRequest and then validating whether a new ExpressionRequest has already been mapped.
You can see that in my "AutomapperSamples.EF.PrerequisiteTests".

@2) Doing this manually does not seem to be the biggest problem to me.
However, when running my tests where I query a <DetailDto> that references its <MasterDto> which in turn contains a collection of <DetailDto>, I found that the detail was already instanciated and added into the <MasterDto> collection. The only drawback for now is that two instances of the <DetailDto> are created: 
<pre>
var detailDto1 = query.ToList().Single();
var detailDto2 = detailDto1.Master.Details.Single();
// now detailDto1 and detailDto2 are completely different object instances, but have the same ID.
</pre>

So could you please be more precise on the problems you think I will encounter?


jbogard:
On the cyclic graphs piece - the SQL queries that come out of EF when you do this are *insane*. Ideally, if you went straight up SQL, you'd use something like CTEs to accomplish this

gentledepp:
One moment - this is not what I am trying to achieve.

I only want a query that is like
<pre>
  context.MasterSet.Include("Details").UseAsDataSource();
</pre>
where this causes a StackoverflowException on ".ToList()".
This type of query is rather basic ans must definetly be supported by EF - so I don't see the problem here ;-/

TylerCarlson1:
I was thinking it would be something like
```
context.MasterSet.UseAsDataSource().For<MasterDto>("Details");
```
as an override, and that can save includes in the query provider and when it gets to project it uses them in the call as expanded variables.

Issue #866, opened 2015-09-01
Mapping Dynamic Object to an Entity Framework DataObject


tsramanathan:
I am trying to perform an update action in which are using partial updates.  The Incoming object is a dynamic object with lesser properties compared to the EF Data object, when we try to use the DynamicMap, it nullfies the properties that are not being passed from dynamic request 

Here is the below example

dynamic  = {"prop1": "value1", 
                     "prop2":"value2"}
Account = {prop1,
                 prop2,
                 prop3,
                 prop4}

Dynamic dynamic = {incoming request} 
Account account = Get(id); (It has the original values)
Mapper.DynamicMap(dynamic, account); (Expected only to change the couple of properties)

In the above it nullifies prop3 & prop4. Is my usage of Dynamic Map correct?  or Is this not the expected functionality of DynamicMap?  

Thanks in Advance!

lbargaoanu:
No, it's just similar names :) But sounds reasonable as a request for Mapper.Map.

lbargaoanu:
@jbogard A [silly gist](https://gist.github.com/lbargaoanu/8cc2a4a7dd0e35b8ef21#file-dynamictoobject-cs). But I've heard the idea before. I think something like this could be useful. Perhaps a built in mapper.

tsramanathan:
Thank you for your reply!  Is this something you think we could add perhaps to Automapper functionality? We have a great requirement in our projects.  

Thanks Again!

jbogard:
@lbargaoanu Yeah, a custom IObjectMapper is likely here, to map from `IDictionary<string, object>` to object I think.

lbargaoanu:
@jbogard Do we want to map only properties that exist on the dynamic source object?

tsramanathan:
Yes that is correct.  So the workflow is:
1. Retrieve the existing data from DB
2. Map the source dynamic object with only incoming properties (Partial Update)
3. Then save the modified data back to DB

Thanks again! Ram

jbogard:
@lbargaoanu Yeah I agree - that's why I wonder if this could be solved with simply some mapper that goes from `IDictionary<string, object>` -> `object`, pretty much how the MVC model binder works.

tsramanathan:
[...]

PR #891, opened 2015-09-15
Dynamic mappers bug 866

lbargaoanu:
#866


lbargaoanu:
@jbogard There are no tests for ExpandoObject, because they would pass with either mapper. The dictionary tests should apply.

lbargaoanu:
In my version the TypeInfo was cached, now it isn't. I just wanted to make sure that's what you wanted.

Issue #893, opened 2015-09-15
Incorrect mapping chosen when immediate source type is not mapped.


MadillJ:
This issue requires a fairly specific set of circumstances for it to occur.

First the source object needs to be an instance of a class that derives the type that is mapped, in my case it is an Active Record Proxy.

Second there needs to be two possible destination types one of which derives from the other.

Finally the source needs to be first mapped to the base destination type, the issue then occurs when mapping to the derived destination type, the type map that is retrieved is the one that was cached on the first mapping which results in an invalid cast.

The problem appears to be in ConfigurationStore.cs between the results of GetRelatedTypePairs() and the select that is performed in ResolveTypeMap(TypePair typePair).  My first guess at the fix is to not include the base class of the destination type as a possible value as I cannot think of a case where that would be valid.

SO question that lead to this issue: http://stackoverflow.com/questions/32574749/auto-mapper-is-failing-to-map-a-castle-active-record-proxy-to-multiple-destinati

lbargaoanu:
Probably duplicate of #842.

MadillJ:
I'm not sure it is as I was unable to resolve it with any variation of Include.

TylerCarlson1:
It's a duplicate of #842 and #887.  It was resolved in 3.3.1 and with 4.0.4 it broke.

Issue #887, opened 2015-09-14
Subtype mappings declares using As() or Include() are ignored when mapping subtypes of the subtype. 


martinlong1978:
Automapper seems to be ignoring subtypes of subtypes which have been declared in subtype mappings using As() or Include(). This is a problem when using EF, for example, as DynamicProxies don't get recognised as the subtype they are a proxy of. 

Using 4.1.0-ci1007 from MyGet

Here are my classes: 
```
 public class Container
    {
        public BaseA Item { get; set; }
    }

    public class Container2
    {
        public BaseB Item { get; set; }
    }
    public class BaseA
    {
        public string Name { get; set; }
    }

    public class BaseB
    {
        public string Name { get; set; }
    }

    public class ProxyOfSubA : SubA
    {
    }
    public class SubA : BaseA
    {
        public string Description { get; set; }
    }

    public class SubB : BaseB
    {
        public string Description { get; set; }
    }
```
And here are my 2 failing tests (on to demonstrate for each of As() and Include() ): 

```
        [Fact]
        public void TestInitialiserProxyOfSub()
        {
            Mapper.Initialize(cfg =>
            {
                cfg.CreateMap<SubA, SubB>();
                cfg.CreateMap<SubA, BaseB>().As<SubB>();
                cfg.CreateMap<Container, Container2>();
            });

            var mapped = Mapper.Map<Container, Container2>(new Container() { Item = new ProxyOfSubA() { Name = "Martin", Description = "Hello" } });
            Assert.IsType<SubB>(mapped.Item);

        }

        [Fact]
        public void TestInitialiserProxyOfSubInclude()
        {
            Mapper.Initialize(cfg =>
            {
                cfg.CreateMap<BaseA, BaseB>().Include<SubA, SubB>();
                cfg.CreateMap<Container, Container2>();
            });

            var mapped = Mapper.Map<Container, Container2>(new Container() { Item = new ProxyOfSubA() { Name = "Martin", Description = "Hello" } });
            Assert.IsType<SubB>(mapped.Item);

        }
```

lbargaoanu:
The include example works if you add

    cfg.CreateMap<SubA, SubB>();

TylerCarlson1:
@lbargaoanu I noticed that too when writing the fix for this issue.  Include doesn't create a mapping of the type for you.

PR #894, opened 2015-09-16
Fixed Issues with property map inheritance

### LINKED ISSUE #842: InvalidCastException when using mapping inheritance
~~~
public class A {};

public class AModel {}
public class BModel : AModel {}

Mapper.CreateMap<A, AModel>()
            .Include<A, BModel>();
Mapper.CreateMap<A, BModel>();
~~~

~~~
Mapper.Map<A, BModel>(anInstanceOfA);
~~~ 

Resulting in:
~~~
System.InvalidCastException: Unable to cast object of type 'AModel' to type 'BModel'. at AutoMapper.MappingEngine.Map[TSource,TDestination](TSource source)
~~~

TylerCarlson1:
Fixes #842, #887, and #893

@jbogard I knew there was a reason GetDerivedTypeFor(Type sourceType) existed.

### DIFF
--- a/src/AutoMapper/ConfigurationStore.cs
+++ b/src/AutoMapper/ConfigurationStore.cs
@@ -417,9 +417,9 @@ namespace AutoMapper
 
         private IEnumerable<TypePair> GetRelatedTypePairs(TypePair root)
         {
-            var includeOverrideTypePairs =
+            var includeOverrideTypePairs = 
                 GetAllTypeMaps()
-                    .Where(tm => tm.SourceType.IsAssignableFrom(root.SourceType) && (tm.DestinationTypeOverride ?? tm.DestinationType) != root.DestinationType && tm.DestinationType.IsAssignableFrom(root.DestinationType))
+                    .Where(tm => tm.HasDerivedTypesToInclude() && tm.SourceType.IsAssignableFrom(root.SourceType) && (tm.DestinationTypeOverride ?? tm.DestinationType) != root.DestinationType && tm.DestinationType.IsAssignableFrom(root.DestinationType))
                     .Select(tm => new TypePair(tm.SourceType,tm.DestinationTypeOverride ?? tm.GetDerivedTypeFor(root.SourceType))).ToList();
             var subTypePairs =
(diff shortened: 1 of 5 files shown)


TylerCarlson1:
Sorry I didn't notice the differences of #842.  It might still fix the problem, but I don't have a unit test for it.

src/AutoMapper/ConfigurationStore.cs
```diff
@@ -417,10 +417,15 @@ public bool HasOpenGenericTypeMapDefined(ResolutionContext context)
 
         private IEnumerable<TypePair> GetRelatedTypePairs(TypePair root)
         {
-            return
+            var includeOverrideTypePairs =
+                GetAllTypeMaps()
+                    .Where(tm => tm.SourceType.IsAssignableFrom(root.SourceType) && (tm.DestinationTypeOverride ?? tm.DestinationType) != root.DestinationType && tm.DestinationType.IsAssignableFrom(root.DestinationType))
+                    .Select(tm => new TypePair(tm.SourceType,tm.DestinationTypeOverride ?? tm.GetDerivedTypeFor(root.SourceType))).ToList();
```
jbogard: Sigh. Yeah I was thinking about this the other day. Good catch.

TylerCarlson1: It should hopefully fix #842 as well.  I'll add the example unit test to see if it passes.
I think both are same problem, it's just #842 gets invalid cast exception because it expects inherited type, while other one does not expect it.

TylerCarlson1: Actually there's some problems with this in inheritance and #842.  It's quite difficult to determine what is inherited and what is not.

TylerCarlson1:
Fixed issues with #842.  Should be ready to submit now.

And that's why I had HasDerivedTypesToInclude() as well.

Glad you didn't delete them with refactor.  Would of been much harder to remember what I did to fix this problem in 3.3.0

martinlong1978:
Would it be possible to either get this merged, or to have a build in MyGet? Thanks.

Issue #900, opened 2015-09-18
AutoMapper unexpectedly mapping Name property to COMPANY_Name property


GauravKGoyal:
Hi,
Mapping PersonViewModel to Person populates Person.COMPANY_Name property with the value of PersonViewModel.Name which i would NOT expect. 
Note if i change uppercase COMPANY to pascal case then mapping does not happen.

    public class PersonViewModel
    {
        public string Name { get; set; }
    }

    public class Person
    {
        public string Name { get; set; }
        public string COMPANY_Name { get; set; }

        public override string ToString()
        {
            return string.Format("Name:{0}, COMPANY_Name:{1}", Name, COMPANY_Name);
        }
    }

    class Program
    {
        static void Main(string[] args)
        {
            AutoMapper.Mapper.CreateMap<PersonViewModel, Person>();
            var person = AutoMapper.Mapper.Map<Person>(new PersonViewModel() {Name = "Bob"});
            Console.WriteLine(person.ToString());
            Console.ReadKey();
        }
    }

Output -> Name:Bob, COMPANY_Name:Bob
Expected Output -> Name:Bob, COMPANY_Name:

Any thoughts would be appreciated. 
Thanks

TylerCarlson1:
The regular expression for pascal is `(\p{Lu}+(?=$|\p{Lu}[\p{Ll}0-9])|\p{Lu}?[\p{Ll}0-9]+)` which breaks up strings pascal format.  Result with `COMPANY_Name` returns `Name`, so matches Name and gets mapping.

GauravKGoyal:
Thanks for the info Tyler. Will do a workaround until there is a new update

jbogard:
@GauravKGoyal so the existing regex matcher isn't perfect. You'd need to replace it with something that matches more specific for you. Here's the current one that allows for C# member naming possibilities and culture-sensitive casing:

`(\p{Lu}+(?=$|\p{Lu}[\p{Ll}0-9])|\p{Lu}?[\p{Ll}0-9]+)`

lbargaoanu:
You can use no naming convention. Or the default one, Pascal. Or any other. But if you don't respect the convention, weird things will happen. Seems fair to me :) I don't think changing the default one for corner cases is the way to go.

GauravKGoyal:
Thanks guys. 
I have resolved it by choosing to ignore the mapping for those properties as they are not many. And i do understand that we can not cater for all the cases.

jbogard:
[...]

PR #922, opened 2015-10-08
Allow disabling the default naming convention

lbargaoanu:
#900

### DIFF
--- a/src/AutoMapper/Internal/IMappingOptionsMemberMapper.cs
+++ b/src/AutoMapper/Internal/IMappingOptionsMemberMapper.cs
@@ -284,8 +284,21 @@ namespace AutoMapper
     }
     public class NameSplitMember : IChildMemberConfiguration
     {
+        INamingConvention _destinationMemberNamingConvention;
+
         public INamingConvention SourceMemberNamingConvention { get; set; }
-        public INamingConvention DestinationMemberNamingConvention { get; set; }
+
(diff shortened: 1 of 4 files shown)


jbogard:
Soooooo I don't think this is the right fix. Probably fixing the existing regex, but in #900 it should probably just use a more appropriate naming convention. A regex can't detect every possibility I think.

lbargaoanu:
I think you should be able to disable the naming convention no matter what :)

jbogard:
What would that do, though? You could always replace it, but there has to
be something.

lbargaoanu:
Well, nothing :) If the splitting expression doesn't really split anything, it's like having no naming convention whatsoever.

lbargaoanu:
Close or rebase? :)

lbargaoanu:
@jbogard I've pointed people to this PR more than a few times. Perhaps reconsider? :)

jbogard:
Maybe this is an "ExactMatchNamingConvention"?

lbargaoanu:
I don't know. But usually people are surprised that the default configuration can lead to these situations, hence the original intent.

jbogard:
I would be OK with having a naming convention that's exact match that you have to set, versus removing it and defaulting it to an exact match.

lbargaoanu:
OK, I guess that would work too. It's a convenience thing after all.

Issue #917, opened 2015-10-07
DynamicMap throws exception when using two derrived classes and an existing map between their base classes


tsebring:
See failing unit test below. Is this a bug or?

```c#
using AutoMapper;
using Microsoft.VisualStudio.TestTools.UnitTesting;

namespace UnitTestProject1
{
    /// <summary>
    /// .NET Framework 4.5
    ///  Using <package id="AutoMapper" version="4.0.4" targetFramework="net45" />
    /// </summary>
    [TestClass]
    public class UnitTest1
    {
        [TestCleanup]
        public void Cleanup()
        {
            Mapper.Reset();
        }

        [TestMethod]
        public void FailingTest()
        {
            Mapper.CreateMap<OneBase, TwoBase>();
            Mapper.DynamicMap<Two>(new One());
        }
        
        class OneBase { }
        class TwoBase { }
        class One : OneBase { }
        class Two : TwoBase { }
    }
}
```

jbogard:
I don't think this is a valid scenario - DynamicMap is only intended for cases where you don't know the type at config time. Things like anonymous types. You're not really supposed to mix the two. If you called CreateMap on One -> Two, this would work.

lbargaoanu:
@jbogard It fails without DynamicMap. Maybe it's missing the map One->Two, but it doesn't say so :)

     Mapper.Map<Two>(new One());

tsebring:
The CreateMap just illustrate the problem. It was never explicitly created in our code rather AutoMapper created it through other calls to DynamicMap.

In our case DynamicMap was used to map two identical objects in different namespaces where the types were not known at compile time.

tsebring:
Example:
```csharp
            Mapper.DynamicMap<TwoBase>(new OneBase());
            Mapper.DynamicMap<Two>(new One());
```

The objects are WCF generated proxys.

TylerCarlson1:
Might want to look at adding a convention in AutoMapper 4.1 as explained in #793.
Maybe something like
```
public class ToWCFProxyProfile : Profile
        {
            protected override void Configure()
            {
                AddConditionalObjectMapper().Where((s, d) => d.Name.StartsWith(s.Name) && s.Namespace == "WCFSource" && d.Namespace == "WCFProxyNameSpace");
            }
        }
```
Shouldn't need AddMemberConfiguration() because identical names, but if so you can make rules for that as well.

jbogard:
[...]

PR #923, opened 2015-10-08
Missing type map with inheritance

lbargaoanu:
#917 
I would say that this code is too complicated for its own good.


jbogard:
Do we need this one anymore with that other As/Include PR in?

lbargaoanu:
I think so. Not really related I would say.

lbargaoanu:
Unless I'm missing something, the original test passes now. But what happens with CreateMap only is that the base type map is returned and the cast fails. It should report the missing map.

Issue #926, opened 2015-10-11
Mapping to proxy instead of configured concrete type


texavery77:
Hello,
I noticed an unexpected mapping using nested properties of interface type with two different implementing types that need to be mapped. Automapper 4.0.4 shows different behavior than Automapper 3.3.1, which behaves as expected and according to documentation.

See following comment as example:

texavery77:
Example program, which fails (leaving out actual class/interface definitions, which are rather obvious):

static void Main(string[] args)
        {
            Mapper.CreateMap<Customer, CustomerData>();          
            Mapper.CreateMap<Address, IAddress>().As<AddressData>();
            Mapper.AssertConfigurationIsValid();

            var customer = new Customer()
            {
                Name = "John Doe",
                Address = new Address()
                {
                    City = "Hamburg",
                    Street = "Fishstreet 999",
                    Zipcode = 20409
                },

                Addresses = new List<IAddress>()
                {
                    new Address()
                {
                    City = "Hamburg",
                    Street = "Fishstreet 999",
                    Zipcode = 20409
                },
                new Address()
                {
                    City = "Hamburg",
                    Street = "Fishstreet 999",
                    Zipcode = 20409
                },
                
                }
            };

            var c2 = Mapper.Map<CustomerData>(customer);

            // Both fail in AutoMapper 4.0.4, but not on 3.x
            Debug.Assert(c2.Address is AddressData);
            Debug.Assert(c2.Addresses.First() is AddressData);

lbargaoanu:
Please make a gist with a complete example.

TylerCarlson1:
MyGet has latest build and should fix this inheritance issue.
https://www.myget.org/F/automapperdev/api/v2

lbargaoanu:
Actually it still fails.

lbargaoanu:
[...]

PR #928, opened 2015-10-13
Map to interface with as bug 926

lbargaoanu:
#926


jbogard:
Is this one good to go?

Issue #925, opened 2015-10-08
NotSuportedException when mapping entities with complex types using AutoMapper and Entity Framework


mcetkovsky:
I would like to map an entity with a complex type to another entity. However, on the last line in the `Main` method a 

> System.NotSupportedException: Cannot compare elements of type 'ConsoleApplication2.ComplexType'. Only primitive types, enumeration types and entity types are supported.

exception is raised.

How to fix it?

I am aware I can flatten the complex type properties to the entity. However, it involves lots of code duplication so I strongly prefer a solution which does not need me to flatten the complex type properties to the entity. Complex types may themselves contain other complex types, so the solution should allow nesting complex types to complex types.

Entity Framework 6.1.3, AutoMapper 3.3.1 and 4.0.4, SQL Server 2014 SP1 Express, VS 2015, .NET 4.5.2 and 4.6.

It works if `.SingleOrDefault(x => x.Id == 1)` or `.ProjectTo<MappedEntity>()` is removed.

If still does not work if the `.SingleOrDefault(x => x.Id == 1)` is rewritten as `.Where(x => x.Id == 1).ToList().SingleOrDefault();`.

    using System.Data.Entity;
    using System.Linq;
    using AutoMapper;
    using AutoMapper.QueryableExtensions;
    
    namespace ConsoleApplication2
    {
        class Program
        {
            static void Main(string[] args)
            {
                Database.SetInitializer(new DropCreateDatabaseAlways<ClientContext>());
    
                Mapper.CreateMap<Entity, MappedEntity>();
                Mapper.CreateMap<ComplexType, MappedComplexType>();
    
                var clientContext = new ClientContext();
                clientContext.Set<Entity>().ProjectTo<MappedEntity>().SingleOrDefault(x => x.Id == 1);
            }
        }
    
        class ClientContext : DbContext
        {
            public virtual DbSet<Entity> Entities { get; set; }
    
            protected override void OnModelCreating(DbModelBuilder modelBuilder)
            {
                base.OnModelCreating(modelBuilder);
    
                modelBuilder.Entity<Entity>().HasKey(x => x.Id);
                modelBuilder.ComplexType<ComplexType>();
            }
        }
    
        class Entity
        {
            public virtual int Id { get; set; }
    
            public virtual ComplexType ComplexType { get; set; }
        }
    
        class ComplexType
        {
            public virtual string Field { get; set; }
        }
    
        class MappedEntity
        {
            public virtual int Id { get; set; }
    
[...]

PR #938, opened 2015-10-20
Complex types bug 925

lbargaoanu:
#925

### DIFF
--- a/src/AutoMapper.Net4/AutoMapper.Net4.csproj
+++ b/src/AutoMapper.Net4/AutoMapper.Net4.csproj
@@ -46,6 +46,7 @@
   <ItemGroup>
     <Reference Include="Microsoft.CSharp, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL" />
     <Reference Include="System, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089, processorArchitecture=MSIL" />
+    <Reference Include="System.ComponentModel.DataAnnotations" />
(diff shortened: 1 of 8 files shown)


jbogard:
Maybe an attribute instead of a marker interface since we have the IgnoreMap attribute already?

lbargaoanu:
Well, yes, but an attribute is more expensive to query and I was tempted to cache it somehow. Decisions, decisions :)

jbogard:
If someone is using EF already, is there an existing ComponentModel attribute they can use? I'm also leaning attributes because it's what EF folks seem to be more OK with than marker interfaces

lbargaoanu:
Spot on :) But it lives in System.ComponentModel.DataAnnotations.dll. Do we want to reference that? :)

jbogard:
Only if it's really available for all platforms. I try not to have compile directives mixed inside a file, and instead scope it to the entire file. This one might be tough to just put in one tiny spot.

lbargaoanu:
New commit.

lbargaoanu:
I can see why no more #if in TypeExtensions, but now what? :) Check by name, a new attribute?

jbogard:
Nah, just remove the alternate. I'm moving to the .NET Platform stuff, which forces me to remove support for older platforms (Silverlight 4 etc), and DataAnnotations are just available. I'm trying to get rid of most of the conditional compilation stuff.

lbargaoanu:
I think ComplexTypeAttribute is only in .NET 4.5.

lbargaoanu:
Or maybe not :)

lbargaoanu:
I'm not sure how useful this is. Complex types are usually configured in code I imagine, not with attributes (that's how I would do it). It doesn't help with #1001. There might be other similar situations.

lbargaoanu:
Maybe reuse RequiredAttribute for #1001?

PR #953, opened 2015-10-24
Make dest property's name type pass over correctly

TylerCarlson1:
Wasn't passing over type of the destination property before.  
Doesn't look like anything used it currently, so no real error, but if people try making their own NameMappers or MemberConfigurations they won't be able to use the destination property type


TylerCarlson1:
Looking at this more I think the interface needs to change to just passing over the Member Info instead of just it's name and type.  The problem is you get the source and the destination property type, but you can't figure out the destination type.

jbogard:
Conflicts :(

TylerCarlson1:
Have to look at this more.  This might be obsolete because of #983 changes the system to be more dynamic.  I'm sure this made the conflict.

TylerCarlson1:
I was wrong in previous assesment.  This needed to be in there as the problem wasn't fixed in #983.
Fixed now

Issue #892, opened 2015-09-15
Strange issue with "doubling up" of destination paths


dplaskon:
I'm having a strange issue with Automapper 4.0.4 that did not occur with Automapper 3.3.1

I'm still trying to isolate a minimal repro (unfortunately this is part of a large/complex solution) to demonstrate the issue, but figured I would post what I have so far, in case anyone has any ideas.

Under certain circumstances, I'm getting exceptions like this:
```
System.Exception: 

Mapping types:
String -> String
System.String -> System.String

Destination path:
CreateCompletedTimelogEntryRequest.RecordedOutputs.RecordedOutputs.RecordedOutputs0[0].OutputValue.OutputValue

Source value:
This is my test value.
   at VolunteerSquared.Core.Extensions.AutoMapperExtensions.MapTo[TDest](Object source) in T:\TeamCity\buildAgent\work\5a9310d77d5a6d38\src\VolunteerSquared.Core\Extensions\AutoMapperExtensions.vb:line 16
   at VolunteerSquared.Web.Areas.Organization.Controllers.HoursController.CreateEntry(CreateEntryModel model) in T:\TeamCity\buildAgent\work\5a9310d77d5a6d38\src\VolunteerSquared.Web\Areas\Organization\Controllers\HoursController.vb:line 275
   at lambda_method(Closure , ControllerBase , Object[] )
   at System.Web.Mvc.ReflectedActionDescriptor.Execute(ControllerContext controllerContext, IDictionary`2 parameters)
   at System.Web.Mvc.ControllerActionInvoker.InvokeActionMethod(ControllerContext controllerContext, ActionDescriptor actionDescriptor, IDictionary`2 parameters)
   at System.Web.Mvc.ControllerActionInvoker.<>c__DisplayClass15.<InvokeActionMethodWithFilters>b__12()
   at System.Web.Mvc.ControllerActionInvoker.InvokeActionMethodFilter(IActionFilter filter, ActionExecutingContext preContext, Func`1 continuation)
   at System.Web.Mvc.ControllerActionInvoker.InvokeActionMethodFilter(IActionFilter filter, ActionExecutingContext preContext, Func`1 continuation)
   at System.Web.Mvc.ControllerActionInvoker.InvokeActionMethodFilter(IActionFilter filter, ActionExecutingContext preContext, Func`1 continuation)
[...]

PR #956, opened 2015-10-26
Fixed duplicated path in the exception message

lbargaoanu:
Reported in #892. But this is just the symptom.


lbargaoanu:
Another issue. Message is a property, but its implementation is not exactly trivial. A field cache would be in order I believe.

Issue #957, opened 2015-10-26
ProjectTo only reading from type map cache and not trying to resolve.


TylerCarlson1:
If use conventions for TypeMaps and don't map them, then ProjectTo will say there's no mapping and to create one.

If you map the types before you call project to it works fine, because the mapping has been cached and can be called by ProjectTo.

Will look into it more and get an example soon.

TylerCarlson1:
Looks like IObjectMapper is a bad place to put convention logic.  ConfigurationStore.ResolveTypeMap is probably a more appropriate place, so it can handle finding type maps.
And possibly OpenGenericMapper should go in here as well.  It looks very out of place where it is now.

jbogard:
Merged.

PR #960, opened 2015-10-28
Fix Conventions and Generic TypeMaps not caching

TylerCarlson1:
Neither would work if you called ProjectTo or try to map an Expression.  The problem is the only way the type maps got cached before was in the IObjectMapper, and because of this you had to map the types before you did ProjectTo in order for the cache to have the TypeMap to make the TypeMapExpression.

This moves the logic for both of them into ResolveTypeMap, saying if Can't find type map and no object mappers can resolve the type maps, then check conventions, then check generics.

Moving the logic here, helps resolve the missing type maps when trying to make a mapping expression without having mapped the objects first.

Generics isn't that big of a deal, cause EF can't support generics right now, but conventions this is a big problem.  It's cool seeing conventions work without explicitly mapping everything out, but it sucks when you go to use ProjectTo and get TypeMap not found, please CreateMap exceptions.


TylerCarlson1:
Oh yea this fixes #957

TylerCarlson1:
Oops. Expression mapper wasn't caching property maps.
Now it's fixed with unit test.

Stupid FindTypeMap vs ResolveTypeMap problem.

jbogard:
Should we fix that stupid problem?

TylerCarlson1:
https://github.com/AutoMapper/AutoMapper/issues/724#issuecomment-106818712 was our last discussion on this.

I wanted to make FindTypeMap private and have everything else use ResolveTypeMap, or something along those lines, but you wanted to keep both names.

TylerCarlson1:
Good question to ask is if, you have conventions, and you call Mapper.FindTypeMap with a type map that falls under a convention, should it return that created type map or null cause it wasn't explicitly created.
Right now it's null unless you mapped using that typemap before calling FindTypeMap, which makes no sense.

jbogard:
Can you fix the merge conflicts?

TylerCarlson1:
Ok will do.  Hopefully in 2-3 at the latest

TylerCarlson1:
Fixed

src/AutoMapper/ConfigurationStore.cs
```diff
@@ -364,6 +364,17 @@ public TypeMap FindTypeMapFor(TypePair typePair)
             return typeMap;
         }
 
+        private bool CoveredByObjectMap(TypePair typePair)
+        {
+            return GetMappers().FirstOrDefault(m => m.IsMatch(new ResolutionContext(null, null, typePair.SourceType, typePair.DestinationType, new MappingOperationOptions(), new MappingEngine(this)))) != null;
+        }
```
jbogard: This bothers me, is there not a better way to see if something is covered by an object map? Do we need to check _all_ mappers?

TylerCarlson1: I don't see any other way then to check all the object mappers to determined if it's covered by any of them.  You can't rely on certain ones by explicitly using them, because they could be removed from mappers and give type map not found exceptions when trying to do the mappings after the fact.  Also down the road there might be more added, and they need to be able to be checked as well.  My Automapper.Collections dll adds 2 IObjectMappers into the master list, and it needs to know about them for when I call Persist, and need to map `Mapper.Map<TDTO,Expression<T,bool>>`.  If it's not all the mapping or is some hidden list, or filter rules, I'm going to have to modify my object mappers I added in in order to make them work with conventions.

If you are concerned about performance, well this should happen only once per TypePair.  Once it's found it will be cahced and this will never have to be called again.  And it's isMatch vs Map which takes less time to compute cause it's usually like 1-5 boolean checks.

jbogard: The really ugly part is newing up the MappingEngine. A while ago we chatted about making object mappers only receive a TypePair. I'm gonna revisit that - perhaps there's something in the middle that makes more sense than having to new up a whole ResolutionContext just to determine if there's a match.

TylerCarlson1: Cheap fix could be to null out the last 2 objects in this function call since they aren't being used.
Also I still agree with using TypePair instead of Resolution Context for IObjectMapper.IsMatch.

jbogard: I'll just merge this and fix it up with getting rid of the ResolutionContext stuff. Be annoying if I pushed a changed and screwed you over at this point.

PR #997, opened 2015-12-10
Lazily mapped queries

gentledepp:
pull request replacement for #871 as it poited to the wrong target branch (automapper/automapper master)

All tests green which are green in targeted development branch.

Remaining issues:

1) AutoMapperSamples.Mappers.QuerySourceInjection.SimpleExample
"Select_source_items_from_destination"
test case fails - error might be due to a wrong testcase configuration i.e. mapping
I added another testcase "Select_source_items_from_destination_with_explicit_mapping" to clarify ths

2) AutoMapper.UnitTests.Query.SourceInjectedQuery tests green, because of kind a hack in SourceInjectedQueryProvider. I decided to go with this hack as it is downwardly compatible, although not correct.

The hack is to default to the old solution in case no parameters and no  memberexpansions were provideded
<pre>
destResult = _mappingEngine.Map(sourceResult, sourceResultType, destResultType);
</pre>

the three lines of code that provide this fallback are in SourceInjectedQueryProvider
<pre>
if (_parameters != null || _membersToExpand.Length != 0)
                    {
                        sourceExpression = rmv.Visit(sourceExpression);
                    }

</pre>

Without hack, the tests fail, because <pre>MappingEngine.Map</pre>
behaves differently than <pre>ProjectTionExpression</pre>
when it comes to nullable fields.
1) in <pre> AutoMapper.UnitTests.Bug.GenericTestExtensionMethods </pre>
Child and ChildDto have pairwise properties which are nullable on one end but not nullable on the other.
That is where MappingEngine.CreateMapExpression kind of gives up at the moment.

2) when you change those fields to be both nullable or non-nullable, the tests still fail because of a "NullReferenceException". This is because the property <pre>Parent.Children</pre> is sometimes <null>
To get rid of it, you'd have to add the following constructor
<pre>
            public Parent()
            {
                Children = new List<Child>();
            }
</pre>

So as nullable propagation is not supported by ProjectionExpression yet, 
a) either the testcases for AutoMapper.UnitTests.Bug.GenericTestExtensionMethods are invalid and should be corrected
b) or nullable type propagation needs to be implemented in the ProjectioExpression and ExpressionMapper

maybe @TylerCarlson1  can do this? :)


gentledepp:
what? the build failed?

TylerCarlson1:
It's trying to build the XProj file not the csproj.  They don't work exactly the same.  I had to change constructor info to use DeclaredType instead of ReflectedType which wasn't available in xproj build solution.

csproj is 4.0 or 4.5.  Easy to test on, but xproj is all the build types that automapper has.

jbogard:
Yep, eventually AutoMapper will ditch all the platform-specific targets. The xproj is what tells me that AutoMapper will run on all the target frameworks of the .NET Standard 1.0 (now done as .NET Platform 5.1 - see this for more https://github.com/dotnet/corefx/blob/master/Documentation/project-docs/standard-platform.md)

.NET 4.5 has a lot of APIs that don't exist cross-platform, so the xproj makes sure you're not using an unsupported API. The tests then only need to test *one* platform, since I would just target one API surface.

Cross platform is complicated.

gentledepp:
Thanks for clarifying  - I already found out :)
I just fixed the build and will commit in a few minutes. I hope that solves the problem...
Why did they have to change so many properties to methods in the new .NET framework?

jbogard:
The biggest change was around reflection - moving the runtime type information out from the type metadata information, that's why you have GetTypeInfo() from a Type object. As to why? I dunno, I just fix red squigglies until the build is green.

gentledepp:
ok... pushed.. hopefully it works now.

jbogard:
There were some merge problems with develop, I still see things like nuspec file changes and project.json changes, those shouldn't be part of it.

src/AutoMapper.sln
```diff
@@ -34,6 +34,20 @@ Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "AutoMapper", "AutoMapper\Au
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "AutoMapper.Net4", "AutoMapper.Net4\AutoMapper.Net4.csproj", "{8C955148-FA96-4F0D-880D-ABB6770F54E9}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "AutoMapper.WinRT", "AutoMapper.WinRT\AutoMapper.WinRT.csproj", "{B80D950D-CC8D-4287-96B1-EEA037825E90}"
```
jbogard: These changes need to be reverted, you shouldn't have any changes to project files, solution files etc.

jbogard:
OK, I think I can fix your merge problems, I'm going to reset back to before the merge and try again.

gentledepp:
hoops - I hope I did not make too many troubles there :-/ But for xproj I had to add compile conditionals unfortunately

jbogard:
I'm going to do my best to NOT have those, though. It's a huge pain to manage all these conditional compilation stuff, and the .NET 4.5 target is really strictly for transitional purposes. I'm expecting that when the .NET platform stuff shakes out, I'll not have so much platform-specific features.

gentledepp:
yeah I can understand that.
so let me if know you're stuck or anything is missing. You can also reach me via skype ("gentledepp")
cheers

gentledepp:
Hi! Any idea when this will be released? (Probably end of this year?)
In that case I can wait, otherwise I'll have to make my own alpha-nuget package in the meantime :)

jbogard:
Milestone date on this one is Jan 19th

gentledepp:
k thanks - so we will go with the private nuget package :)

gentledepp:
Well... did not change a lot and I am confused why this does not build...
Also, I cannot access the appveyor build to see why it failed :-(

jbogard:
I think you need to fix your merge conflicts first?

gentledepp:
Oh maaan... what a crap.
I am getting the following build error from appveyor, however, running psake.bat locally works fine.
<pre>
At C:\projects\automapper\tools\psake\psake.psm1:147 char:17 +                 throw ("Exec: " + $errorMessage) +                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ [<<==>>] Exception: Exec: Error executing command  msbuild /t:Clean /t:Build /p:Configuration=$config /v:q /p:NoWarn=1591 /nologo $source_dir\AutoMapper.sln .
ERROR: 1
</pre>
Is there any way I can find out what "ERROR 1" is? :-|
Command exited with code 1

jbogard:
Scroll up, you'll see the errors there.

gentledepp:
Ok.. I see it now. Thanks.
Seems appVeyor build does not resolve any nuget packages. 
But why is that?
I only see it resolving the dependencies of the project.json and I doubt that I should place Effort and/or NUnit there... :-|
Did you define your appveyor build this way by purpose or did you just forget to add nuget-package resolving?
I really would prefer not to check in any libraries that are available on nuget anyway...

gentledepp:
It seems you're busy and I understand that. But please help me get that build to work!
The PR unitttests work and there are several fixes and improvements in here which are not only interesting for me. At least please have a look at it before you continue to work on your develop branch :)

jbogard:
You shouldn't need NUnit, the project already has unit testing stuff in place. Why do you need that package? Not sure what's going on with NuGet on the server tho

gentledepp:
I have already removed the nuget-package for NUnit and linked the nuget dll from the "lib" folder, but there are other nuget packages that I need.
For example "Effort" to do the entityframework integrationtests without needing a physical database. (muuuch faster) and others.
Unfortunately I cannot look at your appveyor build definition, but according to 
https://www.appveyor.com/blog/2014/03/18/about-nuget-package-restore
nuget restore has to be configured explicitly, but should be quite simple:
"nuget restore" needs to be added to the pre-build commands.
Could you please add that?

Basically, if you just pull it (without committing the merge), you'll see that everything works just fine. All green UTs from the develop branch are green in mine as well. And I have added other UTs for parameterization in the ProjectionQueryProvider (or whatever it is called - the lazy one ;-))

jbogard:
Oh whoops I thought I put my yml into the repo. The command is just
"psake.bat ci".

jbogard:
And this Effort package, I'd never heard of it. VERY INTERESTING.

gentledepp:
It is cool. Basically an EF in-memory dataprovider. Used very often.

gentledepp:
Ok, so how shall we proceed? Are you waiting for something from me?

gentledepp:
Any news on this?

gentledepp:
Any news here? I am still waiting for those changes to pull in.
Sorry, but I have spent way too many hours on it just to abandon it :-/

TylerCarlson1:
Well the PR has merge conflicts and doesn't build.  Those need to happen first before Jimmy can accept it.

gentledepp:
Please read the comments above:
Several weeks ago, when I made the PR, there were no conflicts at all.
Also, the build and the tests ran fine on my machine.
The only problem that remained was that my branch had nuget-packages referenced (Effort, Moq,...) and the appveyor build was (still is??) misconfigured so that it does not resolve the nuget packages.

Please don't get me wrong, but I have actually merged twice, and every time the PR has been stalled until it conflichted with the dev branch again... that is not exactly what keeps me motivated in contributing... :'(

TylerCarlson1:
I understand having to re-merge things.  Have had to do it on quite a few occasions.  Maybe @jbogard can PR your PR to fix the nuget problems, else you might need to fix it yourself so that it builds on the build server somehow.  It looks like to me you can edit psake.bat and all the other files that it calls to make the build work by referencing the nuget package source when resolving.  Maybe you can fix the nuget restore or add data source in psake.bat, or go to psake power shell and add the restore function there.

It should be editable and fixable on your side I believe

jbogard:
The AppVeyor build merely calls out to the command-line build, there's no difference between local and CI build. I'll try and PR the PR.

jbogard:
The hardest part of this merge is all the whitespace changes. For future reference, spaces not tabs, and if you change whitespace, make it a commit that just does that.

gentledepp:
Sorry for that - I will be more careful in the future.
Did you already try it?
If not, I may have some time this weekend. I just need to know how I get the appveyor build to restore the nuget packages and then I can try to merge and fix everything.

So did you already include your "yml" in the dev branch? (that's kind of the appveyor build definition, right?)

jbogard:
I fixed the AppVeyor stuff with the PR I sent you, it was a fix in the actual build script. I could repro locally by clearing out my local packages folder (with VS closed) and trying to build

gentledepp:
Ah ok, now I see: I shall merge your PR with mine, then merge with dev, and then make another PR to Automapper, right?

TylerCarlson1:
If you merge his PR I would think it would just update this PR and try to merge and build against development automaticly.

jbogard:
It would, any commits added to an open PR just update the PR. I do this quite a lot on my projects, titling the PRs with "WIP".

gentledepp:
Added optional null-propagation in order to get some of the tests to work.
Null-propagation is off by default and can be enabled by setting the following option to "true"
<pre>
cfg.EnableNullPropagationForQueryMapping = true;
</pre>
I made this configurable, as EntityFrameworks expressionvisitor complains about the "Expression:Default()" replacements that are used. But we can also change this to be on by default, so EF users can simply switch it off.
Also re-applied the fix for cyclic expression graphs.

gentledepp:
And it BUILDS! I cannot believe it! 💃

gentledepp:
Please don't make me merge again! :-(

jbogard:
I'll PR your PR I think.

gentledepp:
HAHAHA - whatere that means. Go for it. :dancers:

PR #1004, opened 2015-12-14
Add an AllowNull configuration

### LINKED ISSUE #1001: Expression.NotEqual causing an exception when using queryable projections
[Here is the repro repo](https://github.com/mrahhal/AutoMapperEFRepro).

First of all I'm not sure if this problem is ef (or even ef7) specific, but here goes.

We have two simple models with a one-to-many relationship between them, `Article` and `Category`. We also have two DTOs related to them. So we configure the mappings:
```c#
Mapper.Initialize(config =>
    {
        config.CreateMap<Category, CategoryDto>();
        config.CreateMap<Article, ArticleDto>();
    });
```

A simple query like the following directly throws an exception:
```c#
_context.Articles.ProjectTo<ArticleDto>().ToList()
```
```
InvalidOperationException: The binary operator NotEqual is not defined for the types 'System.Int32' and 'System.Object'.
```

I followed the problem to its source, the `MappedTypeExpressionBinder` is generating a condition on `Category` to "handle null source property so it will not create an object with possible non-nullable properties which would result in an exception". You can see the logic [here](https://github.com/AutoMapper/AutoMapper/blob/develop/src/AutoMapper/QueryableExtensions/Impl/MappedTypeExpressionBinder.cs#L31-L37).

This works for optional relationships but for a required one (like the one in my repro) this actually throws an exception because of an incorrect generated comparison logic between int (the table key) and object (null). Probably because ef is seeing that we're checking for null over the navigation property, so it figures that the relationship is "nullable".

One workaround is to set `Mapper.Configuration.AllowNullDestinationValues` to `false` but this is far from a solution because we might actually get into the situation that is described as a reason for the very same generated condition.

This might be a bug in ef7 itself, so I'm not sure but at least can't we provide a solution for this in AutoMapper?

mrahhal:
This is the simplest thing I could think of, and it works. I know that the naming is not appropriate, I think. Should it really be called `AllowNull`?

Usage:
```c#
config
    .CreateMap<News, Dto.NewsDto>()
    .ForMember(n => n.Category, c => c.AllowNull());
```
Where `Category` is another complex DTO of type `CategoryDto`.
Please check [this repro for more details](https://github.com/mrahhal/AutoMapperEFRepro).

This will prevent generating the condition check for `Category` in `NewsDto`. This is necessary only if the relationship between `News` and `Category` is required.

This is supposed to be a workaround for the current release of EF7. (Not sure if EF7 will ever fix the issue because actually it's logical, why would we write a check of null for a relationship that is required?.)

Resolves #1001

### DIFF
--- a/src/AutoMapper/IMemberConfigurationExpression.cs
+++ b/src/AutoMapper/IMemberConfigurationExpression.cs
@@ -83,6 +83,11 @@ namespace AutoMapper
         void Ignore();
 
         /// <summary>
+        /// Allow this member to be null. This prevents generating a check condition for it.
+        /// </summary>
+        void AllowNull();
+
+        /// <summary>
         /// Supply a custom mapping order instead of what the .NET runtime returns
         /// </summary>
         /// <param name="mappingOrder">Mapping order value</param>
--- a/src/AutoMapper/Internal/MappingExpression.cs
+++ b/src/AutoMapper/Internal/MappingExpression.cs
@@ -453,6 +453,11 @@ $"Source member {sourceMember} is ambiguous on type {TypeMap.SourceType.FullName
             _propertyMap.Ignore();
         }
 
+        public void AllowNull()
+        {
+            _propertyMap.AllowNull = true;
+        }
+
         public void UseDestinationValue()
         {
             _propertyMap.UseDestinationValue = true;
--- a/src/AutoMapper/PropertyMap.cs
+++ b/src/AutoMapper/PropertyMap.cs
@@ -66,6 +66,8 @@ namespace AutoMapper
 
         public bool UseDestinationValue { get; set; }
 
+        public bool AllowNull { get; set; }
+
         internal bool HasCustomValueResolver { get; private set; }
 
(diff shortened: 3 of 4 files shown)


jspuij:
+1 for this solution. It makes the null check configurable for each mapping and solves both #1001 and #925.

jbogard:
This is working now.

mrahhal:
@jbogard in 5.0.0-beta-1? What about the api for it?

jbogard:
No, it was fixed by #1130

jbogard:
@mrahhal Can you reopen this PR? I think this is the right fix. I'll fix your PR to update for 5.0.2 once you reopen it.

mrahhal:
@jbogard I'll try to as soons as possible.

mrahhal:
@jbogard you said a while ago:
> This is working now without any changes. Well, ok then.

Does this mean there are still issues that should be fixed manually?

jbogard:
Nah I was wrong. Your changes will fix it.

Issue #1010, opened 2015-12-22
Intermittent Collection was modified; enumeration operation may not execute. on Mapper.Map call


pculver:
I am getting an intermittent exception when calling Mapper.Map\<OrderExtendedDigestDto>(OrderDto).  All of my configuration is done at startup time of the application via Profile classes so I don't know why this would happen. 

The mapping setup looks like

            // Digest DTO
            CreateMap<OrderDto, OrderDigestDto>()
                .ForMember(digest => digest.OrderTemplateId, conf => conf.MapFrom(full => full.CoreTemplate.Id));

            CreateMap<OrderDto, OrderExtendedDigestDto>()
                .IncludeBase<OrderDto, OrderDigestDto>();

The exception is:

```
Collection was modified; enumeration operation may not execute.

   at System.ThrowHelper.ThrowInvalidOperationException(ExceptionResource resource)
   at System.Collections.Generic.List`1.Enumerator.MoveNextRare()
   at System.Collections.Generic.List`1.Enumerator.MoveNext()
   at System.Linq.Enumerable.<ConcatIterator>d__71`1.MoveNext()
   at System.Linq.Enumerable.SingleOrDefault[TSource](IEnumerable`1 source, Func`2 predicate)
   at AutoMapper.TypeMap.ApplyInheritedTypeMap(TypeMap inheritedTypeMap)
   at AutoMapper.TypeMap.Seal()
   at AutoMapper.Mappers.TypeMapMapper.Map(ResolutionContext context, IMappingEngineRunner mapper)
   at AutoMapper.MappingEngine.AutoMapper.IMappingEngineRunner.Map(ResolutionContext context)
```

Any thoughts?  I didn't get this error in version 3.3.1.  I am now using 4.1.1 from nuget.

pculver:
Just saw this in documentation.  

`Mapper.Configuration.Seal(); // this is REQUIRED in AutoMapper 4.0 to make inheritance work`

which I do not do.  Would this be the root of my issue?

lbargaoanu:
Probably. Seal is part of the configuration. I would guess multiple threads were on that stack trace, hence the issue.

pculver:
There were multiple threads.   Should I use `Mapper.Configuration.Seal()` after making a number of calls to `Mapper.Configuration.AddProfile()` or should I use `Mapper.Initialize()`?

lbargaoanu:
Mapper.Initialize is the new way. Seal is more for backward compatibility.

PR #1017, opened 2016-01-07
Match on type pair alone

jbogard:
Some refactoring:
- Makes the IObjectMapper IsMatch match only on TypePairs
- Moves TypeMap mapping into MappingEngine instead of its own thing
- Removed/refactored any IObjectMappers that used anything but a TypePair
- Removed IMappingEngineRunner


jbogard:
This one is fairly large, @lbargaoanu and @TylerCarlson1 thoughts?

src/AutoMapper/ConfigurationStore.cs
```diff
@@ -394,7 +405,8 @@ public TypeMap ResolveTypeMap(TypePair typePair)
                                 FindTypeMapFor(tp) ??
                                 (!CoveredByObjectMap(typePair)
                                     ? FindConventionTypeMapFor(tp) ??
-                                      FindClosedGenericTypeMapFor(tp)
+                                      FindClosedGenericTypeMapFor(tp) ??
+                                      (CreateMissingTypeMaps ? CreateTypeMap(tp) : null)
```
TylerCarlson1: What purpose is CreateMissingTypeMaps for?  Isn't that what conventions are suppose to handle?

TylerCarlson1: Never mind I see it's for dynamic mapping now.

TylerCarlson1:
CreateMissingTypeMaps.  Can instead of having that logic make setting it make a profile or configuration use a conditional mapper of always true?  That way you don't have all these extra if statements in the code.

Also would you even want to set CreateMissingTypeMaps setting anymore in as a mapping configuration?  And I don't understand it all yet, but how does it revert setting back to what it was?  So if it wasn't set and you set the mapping options to true, does it go back to false when you map again?

src/AutoMapper/MappingEngine.cs
```diff
@@ -212,28 +190,22 @@ public TDestination DynamicMap<TDestination>(object source)
 
         public object DynamicMap(object source, Type sourceType, Type destinationType)
         {
+            ConfigurationProvider.CreateMissingTypeMaps = true;
```
TylerCarlson1: Could you make a cached DynamicMappingProfile, with type map condition of true, instead and map using that?

It wouldn't share any prefix and postfixes of the default mapping, but I'm not sure what is to be expected when dynamic mapping.
I guess you could clone the default's properties and clear out TypeMap conventions and add `(sourceType, destType) => true` to achieve the default's behavior but override the confidence.  Or make a wrapper around the default that did everything except TypeMap conventions and override that.

TylerCarlson1:
@jbogard with CreateMissingTypeMaps is it going to be a permanent alternative to DynamicMap?

I would think having conventions exclusively for mapping missed types would be better to replace DynamicMap in the long run vs CreateMissingTypeMaps.  You would have to write an exclusive blog post on this feature first.  I guess it would be much easier creating a property vs explaining how to work a new feature to do the same thing. lol

src/UnitTests/Bug/ExpressionMapping.cs
```diff
@@ -260,14 +260,14 @@ public void When_Using_Non_TypeMapped_Class_Method()
             _valid = new Parent { DateTime = DateTime.Now };
         }
 
-        [Fact]
+        [Fact(Skip = "Failing test")]
```
TylerCarlson1: Ok this is all my fault.  I accept full responsibility for this failure.

Replace lines 264 and 274 with `_valid = new Parent { DateTime = new DateTime(2015,1,1) };` to have the tests past, for this year and every other year.

src/AutoMapper/ConfigurationStore.cs
lbargaoanu: This looks like a dialog in Spanish :) Unreadable.

src/AutoMapper/ConfigurationStore.cs
```diff
@@ -592,24 +606,25 @@ private void DryRunTypeMap(ICollection<TypeMap> typeMapsChecked, ResolutionConte
...
-            }
-            if (mapperToUse is TypeMapMapper)
-            {
                 CheckPropertyMaps(typeMapsChecked, context);
             }
-            else if (mapperToUse is ArrayMapper || mapperToUse is EnumerableMapper || mapperToUse is CollectionMapper)
+            else
             {
-                CheckElementMaps(typeMapsChecked, context);
+                var mapperToUse = GetMappers().FirstOrDefault(mapper => mapper.IsMatch(context.Types));
```
lbargaoanu: This could use a helper method.

src/AutoMapper/Mappers/DynamicMappers.cs
```diff
@@ -62,9 +62,9 @@ protected void SetDynamically(MemberInfo member, object target, object value)
 
     public class FromDynamicMapper : DynamicMapper
     {
-        public override bool IsMatch(ResolutionContext context)
+        public override bool IsMatch(TypePair context)
         {
-            return context.SourceValue.IsDynamic() && !context.DestinationType.IsDynamic();
+            return context.SourceType.IsDynamic() && !context.DestinationType.IsDynamic();
```
lbargaoanu: I used SourceValue so I wouldn't have to check for null. We should check what happens when you map null.

src/AutoMapper/Mappers/StringDictionaryMapper.cs
```diff
@@ -9,33 +9,33 @@ namespace AutoMapper.Mappers
...
         }
     }
 
     public class FromStringDictionaryMapper : IObjectMapper
     {
-        public bool IsMatch(ResolutionContext context)
+        public bool IsMatch(TypePair context)
         {
-            return context.SourceValue is StringDictionary;
+            return typeof(StringDictionary).IsAssignableFrom(context.SourceType);
```
lbargaoanu: The same as with the DynamicMapper, what happens with null.

src/AutoMapper/MappingEngine.cs
lbargaoanu: I guess this could call the overload below with null for destination.

src/AutoMapper/MappingEngine.cs
```diff
@@ -410,40 +382,34 @@ protected override Expression VisitMember(MemberExpression node)
...
 
 
-        object IMappingEngineRunner.Map(ResolutionContext context)
+        public object Map(ResolutionContext context)
         {
             try
             {
+                if (context.TypeMap != null)
+                {
+                    context.TypeMap.Seal();
```
lbargaoanu: This is not thread safe see #1010 for example. And anyway, Seal should have been called on init, no?

src/AutoMapper/MappingEngine.cs
```diff
@@ -410,40 +382,34 @@ protected override Expression VisitMember(MemberExpression node)
...
-                        }
-                        return ObjectCreator.CreateDefaultValue(context.DestinationType);
-                    }
+                    throw new AutoMapperMappingException(context, "Missing type map configuration or unsupported mapping.");
                 }
 
-                return mapperToUse.Map(context, this);
+                return mapperToUse.Map(context);
             }
             catch (AutoMapperMappingException)
```
lbargaoanu: I think an exception filter would work here.

src/AutoMapper/MappingEngine.cs
```diff
@@ -410,40 +382,34 @@ protected override Expression VisitMember(MemberExpression node)
...
+
                 var contextTypePair = new TypePair(context.SourceType, context.DestinationType);
 
                 Func<TypePair, IObjectMapper> missFunc =
-                    tp => ConfigurationProvider.GetMappers().FirstOrDefault(mapper => mapper.IsMatch(context));
+                    tp => ConfigurationProvider.GetMappers().FirstOrDefault(mapper => mapper.IsMatch(contextTypePair));
 
                 IObjectMapper mapperToUse = _objectMapperCache.GetOrAdd(contextTypePair, missFunc);
-                if (mapperToUse == null || (context.Options.CreateMissingTypeMaps && !mapperToUse.IsMatch(context)))
+                if (mapperToUse == null)
```
lbargaoanu: The extra IsMatch check was for a threading bug. Hopefully it works now :)

src/UnitTests/Tests/InitialTypes.cs
```diff
@@ -1,65 +0,0 @@
-﻿using Xunit;
```
lbargaoanu: Not worth keeping I suppose.

src/AutoMapper/ConfigurationStore.cs
lbargaoanu: And the func passed to GetOrAdd should be thread safe. I don't think it is. CreateMap isn't, see #1018 for example.

src/AutoMapper/IConfigurationProvider.cs
```diff
@@ -98,18 +99,10 @@ public interface IConfigurationProvider : IConfiguration
         IEnumerable<IObjectMapper> GetMappers();
```
lbargaoanu: A property would be a better fit.

src/AutoMapper/IConfigurationProvider.cs
```diff
@@ -98,18 +99,10 @@ public interface IConfigurationProvider : IConfiguration
...
-        /// <returns>Type map configuration</returns>
-        TypeMap CreateTypeMap(Type sourceType, Type destinationType, string profileName = ConfigurationStore.DefaultProfileName);
-
-        /// <summary>
-        /// Fired each time a type map is created
+        /// Get all configured mappers
         /// </summary>
-        event EventHandler<TypeMapCreatedEventArgs> TypeMapCreated;
+        /// <returns>List of mappers</returns>
+        IEnumerable<ITypeMapObjectMapper> GetTypeMapMappers();
```
lbargaoanu: A property would be a better fit.

src/AutoMapper/Mappers/EnumerableMapper.cs
```diff
@@ -29,6 +25,15 @@ protected override void ClearEnumerable(IList enumerable)
             enumerable.Clear();
         }
 
+        protected override object GetOrCreateDestinationObject(ResolutionContext context, Type destElementType,
+            int sourceLength)
+        {
+            if (context.DestinationValue is IList && !(context.DestinationValue is Array))
```
lbargaoanu: IList.IsFixedSize would work instead of the array check.

src/AutoMapper/Mappers/NullableMapper.cs
```diff
@@ -4,12 +4,12 @@ namespace AutoMapper.Mappers
```
lbargaoanu: This is not used anymore.

src/AutoMapper/MappingEngine.cs
```diff
@@ -13,7 +13,7 @@ namespace AutoMapper
     using QueryableExtensions;
     using QueryableExtensions.Impl;
 
-    public class MappingEngine : IMappingEngine, IMappingEngineRunner
+    public class MappingEngine : IMappingEngine
```
lbargaoanu: A default constructor would help.

src/AutoMapper/MappingEngine.cs
lbargaoanu: There is also DynamicMap, but that will never work with inheritance I presume.

src/AutoMapper/MappingEngine.cs
jbogard: I made all the DynamicMap calls obsolete. You should just set that one config property instead of doing Map or DynamicMap. Basically the CreateMissingTypeMaps configuration was added, and it made the DynamicMap calls redundant. So prefer the config over other stuff.

src/AutoMapper/MappingEngine.cs
jbogard: Nah DynamicMap thing is really meant for people that can't be bothered with CreateMap

src/AutoMapper/MappingEngine.cs
jbogard: And I think I might be missing what your suggestion is. Maybe I'll just merge it and you can give it a go? Part of the purpose here is to have those object mappers only match on a TypePair.

TylerCarlson1: My suggestion is instead of having CreateMissingTypeMaps be used as a bunch of if statements all throughout the code, is if you set it to true you add a profile that has a convention that's always true, so everything can map to everything.  And when you set to false it removes said convention from list.

The question I have is should this new convention not have prefix and postfixes or should it wrap around the default profile and share it's prefix and postfixes.

I can submit a PR after merge if you want.

TylerCarlson1: Oh and with DynamicMap being obsolete, do you not bother with this issues?
```
public void Main()
{
     Mapper.DynamicMap<A, NotMappedToA>(new A());
     Mapper.Map<B, NotMappedToB>(new B());
}
```
It's expected to throw a map not found exception, but with this commit it won't. and map B to NotMappedToB, since you don't set CreateMissingTypeMaps back to false after mapping

Issue #1023, opened 2016-01-12
Am I doing maps to constructors wrong?


tlycken:
I have two types somewhat like this:

```
public class Foo
{
    public string Foo { get; }
    public int Bar { get; }

    public Foo(string foo, int bar)
    {
        Foo = foo;
        Bar = bar;
    }
}

public class Baz
{
    public string Foo { get; set; }
    public int Bar { get; set; }
}
```

For these, I create a mapping using a profile, in which I have

```
CreateMap<Baz, Foo>();
```

`Mapper.AssertConfigurationIsValid()` doesn't complain, but when I try to map a `Baz` instance to `Foo` I get an error stating 

> Type needs to have a constructor with 0 args or only optional args
> Parameter name: type.

I was under the impression that AutoMapper would be able to map the properties of `Baz` onto the constructor arguments of `Foo`, but apparently not. Am I doing something wrong or stupid here?

lbargaoanu:
It should work. Did you try the latest from MyGet?

tlycken:
Yes, this is on version 4.1.1.

lbargaoanu:
So MyGet, not NuGet, right? :)

tlycken:
Oh, sorry. No, NuGet. I can try MyGet when I get back to work tomorrow, if it matters.

tlycken:
Back to work now. With version 4.2.0-ci-1038, I get the same behavior.

I'm trying to build a MWE that reproduces the problem, but I'm not getting the error there, so it's likely this is something in my actual mappings that's not working, and not an issue with AutoMapper. I'll keep trying and get back with the results, whether this is a bug or PEBKAC.

tlycken:
OK, I found the culprit. This works:

```
// DTOs

public class Foo
{
    public string Bar { get; set; }
    public int Baz { get; set; }
}

public class Qux
{
    public string Bar { get; }
    public int Baz { get; }

    public Qux(string bar)
            : this(42, bar)
    {
    }

    public Qux(int baz, string bar)
    {
        Bar = bar;
        Baz = baz;
    }
}

// in mapping profile
CreateMap<Foo, Qux>();

// main program
Mapper.AddProfile<MapperProfile>();

var foo = new Foo
{
    Bar = "bar",
    Baz = 17
};
var qux = Mapper.Map<Qux>(foo);
```

[...]

PR #1024, opened 2016-01-13
Faling test for #1023

tlycken:
This adds a failing test for #1023. If a property

```
public int Bar { get; set; }
```

is added to the `SourceFoo` type, and the setup is updated accordingly,
the test passes, but I would like it to work without that property.

### DIFF
--- a/src/AutoMapper/ConstructorParameterMap.cs
+++ b/src/AutoMapper/ConstructorParameterMap.cs
@@ -36,5 +36,10 @@ namespace AutoMapper
         {
             SourceResolvers = resolvers;
         }
+
+        public void ResolveUsing(IValueResolver resolver)
+        {
+            SourceResolvers = new[] { resolver };
+        }
     }
 }
--- a/src/AutoMapper/ICtorParamConfigurationExpression.cs
+++ b/src/AutoMapper/ICtorParamConfigurationExpression.cs
@@ -14,6 +14,12 @@
         /// <typeparam name="TMember">Member type</typeparam>
         /// <param name="sourceMember">Member expression</param>
         void MapFrom<TMember>(Expression<Func<TSource, TMember>> sourceMember);
+
+        /// <summary>
+        /// Map constructor parameter from custom func
+        /// </summary>
+        /// <param name="resolver">Custom func</param>
+        void ResolveUsing(Func<TSource, object> resolver);
     }
 
     public class CtorParamConfigurationExpression<TSource> : ICtorParamConfigurationExpression<TSource>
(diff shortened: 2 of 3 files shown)


jbogard:
Merge this in when #1025 gets merged, to make sure this situation is covered.

lbargaoanu:
This is not covered because is not easy to do and #1025 solves the problem with optional parameters. If you want this too, I'll give it a try.

tlycken:
As soon as I get an opportunity, I'll take a look at our actual use case to make sure that #1025 covers what we need, and then I'll report back here.

tlycken:
Is there a new CI build I can test this against?

lbargaoanu:
http://myget.org/gallery/automapperdev

tlycken:
This doesn't quite seem to work yet, even with optional parameters - it seems the optional constructor argument is favored over an available property with the same name:

    public static class Program
    {
        public static void Main(string[] args)
        {
            var configuration = new MapperConfiguration(config =>
            {
                config.AddProfile<MapperProfile>();
            });
            var mapper = configuration.CreateMapper();

            var foo = new SourceFoo
            {
                Foo = "bar",
                Bar = 17
            };
            var qux = mapper.Map<DestinationFoo>(foo);

            Console.WriteLine($"{qux.Foo}, {qux.Bar}"); // Prints "bar, 42" rather than "bar, 17"
        }

        public class SourceFoo
        {

            public int Bar { get; set; }
        }

        public class DestinationFoo
        {
            public string Foo { get; }
            public int Bar { get; }

            public DestinationFoo(string foo, int bar = 42)
            {
                Foo = foo;
                Bar = bar;
            }
        }

        public class MapperProfile : Profile
        {
            {
                CreateMap<SourceFoo, DestinationFoo>();
            }
        }
    }

The previous problem with missing properties etc seems to be fixed, though.

lbargaoanu:
Bar is read-only so it's mapped with the constructor. Looks good to me :)

tlycken:
But I had expected the `bar` parameter of the constructor to be supplied the value of `sourceFoo.Bar`, in effect calling `DestinationFoo(sourceFoo.Foo, sourceFoo.Bar)` rather than `DestinationFoo(sourceFoo.Foo /* bar = 42 implied */)`.

lbargaoanu:
Obviously :) I'll check the test again.

lbargaoanu:
It seems OK to me (I see 17). The PR was merged. I'm not sure what's wrong. Can you check with a decompiler that the code in the PR is in the DLL you're running?

tlycken:
Yes, it seems so. In the decompiled sources, I see the following lines:

```c#
if (!canResolve && info.HasDefaultValue)
{
    canResolve = true;
}
```

which, except for the name of the variable `info` corresponds to the code added in the PR. With a breakpoint on the first of these lines I can see that `canResolve == true` both times I get there, even before evaluating the `if` clause.

lbargaoanu:
It's a little silly, but you can use a PR to run your test :)

PR #1026, opened 2016-01-15
CreateMissingMappings adds convention

TylerCarlson1:
This is what I was talking about with CreateMissingMappings property in #1017

Removes extra if statement with a type pair convention that automatically maps everything.

Took out the get cause it doesn't look that useful with the new implementation, but I could put it back in no problem.


jbogard:
Looks good.

PR #1029, opened 2016-01-17
Fix ConcurrentDictionary issues

aelij:
The current implementation of `ConfigurationStore` causes `ConcurrentDictionary` to lock on all all buckets. This can be a serious perf hit. When you enable the ETW provider `System.Collections.Concurrent.ConcurrentCollectionsEventSource` (or use the Concurrency Visualizer) you can see the following event:

> System.Collections.Concurrent.ConcurrentCollectionsEventSource ConcurrentDictionary acquiring all locks on 32 bucket(s).

This happens when using some "innocent" looking members of the dictionary, such as the `Values` property. You can read more about it in my [blog post](https://arbel.net/2013/02/03/best-practices-for-using-concurrentdictionary/).

### DIFF
--- a/src/AutoMapper/ConfigurationStore.cs
+++ b/src/AutoMapper/ConfigurationStore.cs
@@ -60,7 +60,7 @@ namespace AutoMapper
 
         internal void ForAllMaps(string profileName, Action<TypeMap, IMappingExpression> configuration)
         {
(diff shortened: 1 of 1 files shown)


lbargaoanu:
Most of these methods make sense only during configuration, on a single thread, so it doesn't really matter, but still worth doing I guess. Good to know anyway.

Issue #1031, opened 2016-01-18
Non-static version of Mapper


jbogard:
Things to do:
- [x] Introduce non-static API
- [x] Obsolete existing static API
- [x] Provide bridge for LINQ #1038 
- [x] Read-only configuration #1039

I want to move to a non-static version of AutoMapper, that would get rid of so many headaches on my part of threading and junk. I'm thinking of something like:

```
var mapper = new Mapper(cfg => cfg.CreateMap<Source, Dest>());

var dest = mapper.Map<Source, Dest>(new Source());
```

I'd have overloads for the ctor:

```
public Mapper(Action<IConfiguration> configure);

public Mapper(IConfigurationProvider configuration);
```

And the configuration provider, if you wanted to initialize that by itself, could have a ctor too:

```
public class Configuration(Action<IConfigurationProvider> configure) {
   configure(this);
}
```

Technically configuration is still mutable after Seal(), since you can still go into TypeMaps and do whatever, grab the instance of the configuration and so on.

# Motivation

The static has been haunting me for years. Threading issues, order-of-operations, I can't fix that until I get rid of the statics. The changes in inheritance in 4.x have opened enough issues that I want to be deliberate about this move.

# Upgrade path

If you're using `Mapper.Initialize`, it's a matter of replacing the outer path with the real thing. I want to use the `Mapper` class and a `Configuration` class. I would obsolete all the static methods, obsolete the MappingEngine class and make the Mapper class inherit from it for 4.2. Then in 5.0, remove all the static.

Users, instead of calling Mapper.Map, would use a mapper instance directly, which they can make static all they like. Or configure in a container, whatever.

# Static compromise

~~Another option is to keep `Mapper.Initialize` and `Mapper.Map` and nothing else. Still include static field for the mapper and the configuration instances. I would still remove the static keyword from Mapper and move the MappingEngine pieces inside of Mapper, with the clear signal that "you might want to use the instance methods".~~ 

I want to move to a class called "Mapper" and in C# a static method hides an instance method of the same name. It's basically static or nothing. I could do something like Mapper.Engine.Map but that seems hokey.

@TylerCarlson1 @lbargaoanu thoughts?

flq:
Would the Mapper instance still be usable by multiple threads?

jbogard:
@flq Yes, I would continue to use concurrent dictionaries to cache mapping plans.

TylerCarlson1:
[...]

PR #1034, opened 2016-01-19
Mapping engine cleanup

jbogard:
Working version of #1031


src/AutoMapper/Internal/IMappingOptionsMemberMapper.cs
```diff
@@ -289,7 +289,7 @@ public class NameSplitMember : IChildMemberConfiguration
         public INamingConvention SourceMemberNamingConvention { get; set; }
         public INamingConvention DestinationMemberNamingConvention { get; set; }
 
-        public IEnumerable<MethodInfo> SourceExtensionMethods => (Mapper.Configuration as ConfigurationStore).SourceExtensionMethods;
+        public IEnumerable<MethodInfo> SourceExtensionMethods => (Mapper.Configuration as MapperConfiguration).SourceExtensionMethods;
```
lbargaoanu: I think a cast would work better here. Let it crash if it's the wrong type.

src/AutoMapper/IMappingEngine.cs
```diff
@@ -172,11 +160,20 @@ public interface IMappingEngine
         /// <param name="destination"></param>
         /// <param name="sourceType">Source type to use</param>
         /// <param name="destinationType">Destination type to use</param>
-        [Obsolete("Set the CreateMissingTypeMaps property on Mapper.ConfigurationProvider or your Profile instead")]
         void DynamicMap(object source, object destination, Type sourceType, Type destinationType);
 
+        /// <summary>
+        /// Configuration provider for performaing maps
+        /// </summary>
+        IConfigurationProvider ConfigurationProvider { get; }
```
lbargaoanu: Why expose this? I think the Mapper should be read-only after CreateMapper.

TylerCarlson1: ^

jbogard: It is, the IConfigurationProvider interface is now complete segregated from the mutable side.

TylerCarlson1:
https://github.com/AutoMapper/AutoMapper/blob/develop/src/AutoMapper/Mappers/ExpressionMapper.cs#L211 & https://github.com/AutoMapper/AutoMapper/blob/develop/src/AutoMapper/Mappers/ExpressionMapper.cs#L272 is problem for query tests failing.  Need to pass IMapper or IConfigurationProvider into constructor to fix.

lbargaoanu:
See the comments for [this post](http://blogs.msdn.com/b/pfxteam/archive/2012/02/04/10264111.aspx) about GetOrAdd.

jbogard:
@TylerCarlson1 that did it. The build is passing. One weirdness is the LINQ extension methods will require you to pass in the configuration objects:

```
var config = new MapperConfiguration(cfg => cfg.CreateMap<Source, Dest>());
var builder = config.CreateExpressionBuilder();
_dest = values.ProjectTo<Dest>(builder).ToArray();
```

That builder caches expressions, so I didn't want to combine it with the IMapper/IMappingEngine interface as those might have different lifecycles. IMapper you could actually do one per scoped context, but the expression builder you can cache longer, statically if you want.

I kinda expect people to build their own overloads, and I didn't want to provide a built-in way of doing so I don't think?

I dunno, should I provide a way to expose a setter for an expression builder to not force it to be supplied?

TylerCarlson1:
If you want this in 4.2 or the 4.x series I think it needs to work as is without having to change code as much as possible.  That's what I tried to do with conventions making both the old system work exactly how you expect them to.  Obviously my case sensitivity change was something I didn't make backwards compatible and that had maybe 5-6 issues pop up from that.
That's my limited knowledge on numbering versions.  5.x is when you remove static completely and code using old API intentionally breaks and has to be rewritten.  This probably won't be the case as something always breaks with big changes, but something to strive for.

As for making passing the builder into extensions, I would suggest making `Mapper.Initialize ` set this function for the setter build property.  We are pushing `Mapper.Initialize` to fix build problems, so it should set the expression builder so you don't have to pass it.  I don't know if you should expose it publicly or not.  Maybe the function where you don't pass it in is supported only supported from the Static API, and they have to make their own extension methods if they want to use an instance builder.  The extension where you don't pass the expression builder should also be obsolete and tell the user to use instance

I don't see any good way to satisfy everyone.  I would say static based extensions still work as intended, though Obsolete, and the new extensions you have to pass in the expression builder.

I have to think about it more and look at the code.

Another thing.  I think expression builder should come off of mapper instead of configuration provider.  This is because you only generate one immutable object off of the configuration, where if there was two objects you could possibly make them at two different times.

This is more of a personal preference, I think, and I haven't put as much thought into how this will effect things, as far as refactoring code goes.

jbogard:
btw I just merged this because I wanted to have a chance to give the API a try.

I started down the path of exposing all of those static funcs directly, but I can certainly keep them and mark them as obsolete.

In fact, I'll do that.

jbogard:
And as for the two objects, mapper and expression builder, the reason I did that was because those I thought might have different lifecycles. You can use an IMapper instance with a nested/child container, but with the LINQ stuff you don't really. I imagined IConfiguration/IMapper the same as ISessionFactory/Configuration/ISession with NHibernate. The LINQ stuff I thought could even live off of Configuration, because it only really depends on the configuration anyway. Maybe that's the answer? It would be the difference between `IConfigurationProvider.CreateExpressionBuilder()` and `ProjectTo<Foo>(config)`

But it is a little awkward for sure.

TylerCarlson1:
Oh I thought IConfigurationProvider was mutable, but since that's not the case my point is invalid.

jbogard:
The instance still is mutable underneath the covers. There's just no API exposed to do so. You can only ask it for type maps.

TylerCarlson1:
I don't have a problem with it being under IConfigurationProvider, that's just a matter of personal preference at this point.  I have a problem with the configuration being mutable.

PR #1037, opened 2016-01-20
Added sorting of interfaces

### LINKED ISSUE #1036: Incorrect mapping when source implement multiple interfaces
When source class implement several interfaces, it is possible that least-derived interface would be used for mapping.
Here is an example:
```
using AutoMapper;
using NUnit.Framework;

namespace AutoMapperSamples
{
    namespace Interfaces
    {
        [TestFixture]
        public class MappingToInterfaces2
        {
            [Test]
            public void TestMultipleInterfaceMapping()
            {
                var from = new MapFrom
                {
                    Element = new MapFromElementDerived { Prop = "qwerty", Prop2 = "asdf" }
                };
                Mapper.Initialize(
                    cfg =>
                    {
                        cfg.CreateMap<IMapFrom, MapTo>();
                        cfg.CreateMap<IMapFromElement, IMapToElementWritable>()
                        .Include<IMapFromElementDerived, IMapToElementWritableDerived>();

                        cfg.CreateMap<IMapFromElementDerived, IMapToElementWritableDerived>()
                            .ConstructUsing((ResolutionContext item) => new MapToElementDerived());

                        cfg.Seal();
                    });

                var result = Mapper.Map<MapTo>(from);
                var element = (IMapToElementWritableDerived)result.Element;

                Assert.AreEqual(element.Prop2, "asdf");
            }
        }

        public interface IMapFrom
        {
            IMapFromElement Element { get; }
        }

        public class MapFrom : IMapFrom
        {
            public MapFromElement Element { get; set; }
            IMapFromElement IMapFrom.Element => Element;
        }


        public interface IMapFromElement
        {
            string Prop { get; }
        }

        public interface IMapFromElementDerived : IMapFromElement
        {
            string Prop2 { get; }
        }

        public interface IMapFromElement2 : IMapFromElement
        {
        }

        public interface IMapFromElementDerived2 : IMapFromElementDerived, IMapFromElement2
        {
        }

        public class MapFromElement : IMapFromElement
        {
            public string Prop { get; set; }
        }


        public class MapFromElementDerived : MapFromElement, IMapFromElementDerived2
        {
            public string Prop2 { get; set; }
        }


        public class MapTo
        {
            public IMapToElementWritable Element { get; set; }
        }

        public interface IMapToElementWritable : IMapFromElement
        {
            new string Prop { get; set; }
        }

        public interface IMapToElementWritableDerived : IMapFromElementDerived, IMapToElementWritable
        {
            new string Prop2 { get; set; }
        }

        public abstract class MapToElement : IMapToElementWritable
        {
            public string Prop { get; set; }
        }

        public class MapToElementDerived : MapToElement, IMapToElementWritableDerived
        {
            public string Prop2 { get; set; }
        }
    }
}
```

Here, instead of creation MapToElementDerived mapper create IMapToElementWritable proxy.

iskiselev:
Added sorting of interfaces implemented by type to resolve #1036.

### DIFF
--- a/src/AutoMapper/MapperConfiguration.cs
+++ b/src/AutoMapper/MapperConfiguration.cs
@@ -483,7 +483,9 @@ namespace AutoMapper
                 baseType = baseType.BaseType();
             }
 
-            foreach (var interfaceType in type.GetTypeInfo().ImplementedInterfaces)
+            var allInterfaces = type.GetTypeInfo().ImplementedInterfaces.OrderByDescending(t => t, new InterfaceComparer());
+
+            foreach (var interfaceType in allInterfaces)
             {
                 yield return interfaceType;
             }
@@ -652,5 +654,22 @@ namespace AutoMapper
             }
         }
 
+        private class InterfaceComparer : IComparer<Type>
+        {
+            public int Compare(Type x, Type y)
+            {
+                var left = x.IsAssignableFrom(y);
+                var right = y.IsAssignableFrom(x);
+                if (left & !right)
+                {
+                    return -1;
+                }
+                if (!left & right)
+                {
+                    return 1;
+                }
+                return 0;
+            }
+        }
     }
 }


jbogard:
Can you address the merge conflicts?

lbargaoanu:
You would have a similar problem even if the interfaces don't inherit each other. You might expect interfaces you implement to come before interfaces you inherit.

iskiselev:
@lbargaoanu, I can do this fix more complicated that will also compare, on which type in inheritance hierarchy we implemented interface first, but I can't found any example where it will matter.

iskiselev:
Rebased on current.

lbargaoanu:
Also happens when interfaces [don't inherit each other] (https://gist.github.com/lbargaoanu/6b6b6f115b4fa51900a5#file-noninheritinginterfaces-cs).

iskiselev:
@lbargaoanu, thanks for example. Will try to add test cases to #1050 then.
Still have no idea how to use GetInterfaceMap for it - as it only provide info which class method implements interface method, but give no idea which type is first interface implementation.

lbargaoanu:
I though that we could return only the interfaces implemented by that type, not the inherited interfaces. The interface map tells you which methods implement the interface. You can get from the method the declaring type and filter the results.

iskiselev:
It will omit marker interface with no methods. Not sure if it is really critical.

lbargaoanu:
In the end every interface must be considered, but only when the current type is the one implementing it, not a base type. So it's sorting after all.

lbargaoanu:
Not sure about marker interfaces, but since they don't have members to map, I would guess they don't matter. Anyway, try to make it fail :)

iskiselev:
I'll try to play with it tomorrow, but I feel that using GetInterfaceMap will end with even more complicated solution than #1050 now.

jbogard:
Which is the right PR to keep? And merge conflicts :(

PR #1039, opened 2016-01-22
Read only configuration

jbogard:
Forcing read-only configuration

### DIFF
--- a/src/AutoMapper/MapperConfiguration.cs
+++ b/src/AutoMapper/MapperConfiguration.cs
@@ -10,19 +10,11 @@ namespace AutoMapper
     using QueryableExtensions;
     using QueryableExtensions.Impl;
 
-    [Obsolete("ConfigurationStore is renamed to MapperConfiguration. Use MapperConfiguration instead.")]
-    public class ConfigurationStore : MapperConfiguration
-    {
-        
-    }
-
     public class MapperConfiguration : IConfigurationProvider, IConfiguration
     {
         private readonly ITypeMapFactory _typeMapFactory;
         private readonly IEnumerable<IObjectMapper> _mappers;
         private readonly IEnumerable<ITypeMapObjectMapper> _typeMapObjectMappers;
-        private readonly object _padlock = new object();
-        private bool _sealed = false;
         internal const string DefaultProfileName = "";
 
         private readonly ConcurrentDictionary<TypePair, TypeMap> _userDefinedTypeMaps =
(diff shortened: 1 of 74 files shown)


jbogard:
Tagging @TylerCarlson1 and @lbargaoanu

src/UnitTests/AutoMapperSpecBase.cs
```diff
@@ -22,7 +22,6 @@ public abstract class NonValidatingSpecBase : SpecBase
         protected abstract MapperConfiguration Configuration { get; }
         protected IConfigurationProvider ConfigProvider => Configuration;
         protected IMapper Mapper => Configuration.CreateMapper();
```
lbargaoanu: Maybe "=" instead of "=>"? The same below.

jbogard: Won't compile like that

src/AutoMapper/MapperConfiguration.cs
```diff
@@ -34,49 +26,42 @@ public class MapperConfiguration : IConfigurationProvider, IConfiguration
...
             configure(this);
-        }
 
-        public MapperConfiguration(Action<IConfiguration> configure, IEnumerable<IObjectMapper> mappers, IEnumerable<ITypeMapObjectMapper> typeMapObjectMappers) 
-            : this(mappers, typeMapObjectMappers)
-        {
-            configure(this);
+            Seal();
+
+            ExpressionBuilder = new ExpressionBuilder(this);
```
lbargaoanu: It should pay for use. Now it is, because the ExpressionBuilder constructor doesn't do much. But we should keep it that way or create it lazily.

jbogard: Pay for use? What do you mean?

lbargaoanu: I mean if they don't use ProjectTo, they shouldn't pay for it. You create the ExpressionBuilder eagerly, and now it's OK, but I think we should keep that in mind.

src/UnitTests/AutoMapperSpecBase.cs
lbargaoanu: Yes. Initializing in the constructor works.

src/AutoMapper/MapperConfiguration.cs
jbogard: Ah I see, thanks! It would be easy to make it `Lazy<T>`

jbogard:
@TylerCarlson1 you're right that did simplify things quite a bit

TylerCarlson1:
@jbogard Awwww you didn't lambda the single line _defaultProfile calls.

PR #1043, opened 2016-01-23
Obsolete Mutable Profile calls

TylerCarlson1:
Right now Profiles are mutable.  You can edit profiles from the class itself, and getting out of configuration mappings.
```
var profile = new MyProfile();
var config = new MapperConfiguration(cfg => cfg.AddProfile(profile));
profile.CreateMap<Foo, Bar>();
```
So like MapperConfiguration is immutable this is a step towards that for profiles.  Can't remove functionality so obsoleted functions that change state, so in 5.0 they can be taken out and profiles can become immutable as part of the configuration.

`IConfiguration.CreateProfile(string profileName);` also has this problem where you can change it outside the configuration level, but that might be a little too anal about this.  You could even say the other CreateProfile has this problem because in the Action you can set a profile variable to the configuration and edit it later, but at this point it would feel more like a hack and you know this wasn't the intended way of doing things.
This PR doesn't address this, but I can add it or make a new one.

### DIFF
--- a/src/AutoMapper/MapperConfiguration.cs
+++ b/src/AutoMapper/MapperConfiguration.cs
@@ -650,11 +650,6 @@ namespace AutoMapper
             public NamedProfile(string profileName) : base(profileName)
             {
             }
-
-            protected override void Configure()
-            {
-                // no-op
-            }
         }
 
     }
--- a/src/AutoMapper/Profile.cs
+++ b/src/AutoMapper/Profile.cs
@@ -8,7 +8,7 @@ namespace AutoMapper
     using System.Collections.Generic;
     using System.Reflection;
     using Internal;
-
+    
     /// <summary>
     /// Provides a named configuration for maps. Naming conventions become scoped per profile.
     /// </summary>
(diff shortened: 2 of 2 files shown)


jbogard:
Yeah I saw that as I was going through. What I'd like is something similar
to the MapperConfiguration class. Force a lambda to configure itself.

TylerCarlson1:
That's more or less what I did.  I didn't do the lambda through the constructor though.  Only from setting Action that's run when configuring the profile.  Can only be set from inheriting Profile right now.  Shouldn't be too hard to fix

TylerCarlson1:
Do you want me to change it to more match MapperConfiguration with the constructor having overload to pass configuration lambda?

Or do you want to do it yourself, so you can have a consistent feel over the new API?  I didn't have a great explanation for Obsoleting function calls.

jbogard:
I'll merge this in and create a separate PR.

jbogard:
So, I went down a little bit of a hole here to get the Profile to work like MapperConfiguration, but it was too hard without totally removing the static API calls. So, unfortunately I'll have to revert this PR in the meantime.

For 5.0, the two interfaces will look very similar to the Registry class in StructureMap - one is the API to record your configuration, and then the explicitly implemented one is to apply the configuration to the MapperConfiguration class and TypeMaps. Once configuration is complete, you wouldn't have the Profile classes stick around - all that configuration dictates how the TypeMaps behave, so I can push all that config down into the TypeMap itself.

TylerCarlson1:
So removing things like
```
new MapperConfiguration(cfg => {
    cfg.AddProfile(profile);
    var profile2 = cfg.CreateProfile(string profileName);
});
```
taking away any possibility of it being edited outside of configuration?
So who cares if profile is mutable or not.  There's no way to edit it after configuration

PR #1050, opened 2016-01-26
Added sorting of interfaces (2)

iskiselev:
Here is more complicated version, that tries to fix problem that @lbargaoanu noted for original (#1037) version.

This version will iterate interfaces in such order, that more derived interfaces will be before less derived (for IInterfaceBase:IInterfaceDerived, order will be IInterfaceDerived, IInterfaceBase) and iterate interfaces implemented in more derived type before interface implemented in less derived types (so, if we have Base:IInterface1 and Derived:Base,IInterface2 order will be IInterface2, IInterface1).

I was not able found any real example where it may really matter, so probably #1037 will be enough and this is not necessary.

### DIFF
--- a/src/AutoMapper/MapperConfiguration.cs
+++ b/src/AutoMapper/MapperConfiguration.cs
@@ -408,7 +408,9 @@ namespace AutoMapper
                 baseType = baseType.BaseType();
             }
 
-            foreach (var interfaceType in type.GetTypeInfo().ImplementedInterfaces)
+            var allInterfaces = type.GetTypeInfo().ImplementedInterfaces.OrderByDescending(t => t, new InterfaceComparer());
+
+            foreach (var interfaceType in allInterfaces)
             {
                 yield return interfaceType;
             }
@@ -530,6 +532,23 @@ namespace AutoMapper
                         DryRunTypeMap(typeMapsChecked, memberContext);
                     }
                 }
+            }
+        }
+        private class InterfaceComparer : IComparer<Type>
+        {
+            public int Compare(Type x, Type y)
+            {
+                var left = x.IsAssignableFrom(y);
+                var right = y.IsAssignableFrom(x);
+                if (left & !right)
+                {
+                    return -1;
+                }
+                if (!left & right)
+                {
+                    return 1;
+                }
+                return 0;
             }
         }
     }
--- a/src/AutoMapper/MapperConfiguration.cs
+++ b/src/AutoMapper/MapperConfiguration.cs
@@ -399,6 +399,21 @@ namespace AutoMapper
 
         private IEnumerable<Type> GetAllTypes(Type type)
         {
+            var typeInheritance = GetTypeInheritance(type);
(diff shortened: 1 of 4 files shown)


lbargaoanu:
The test is missing :)
This is too complicated. It can be solved more easily using [GetRuntimeInterfaceMap](https://msdn.microsoft.com/en-us/library/system.reflection.runtimereflectionextensions.getruntimeinterfacemap%28v=vs.110%29.aspx).

iskiselev:
I thought about tests, but was not sure how to implement them properly.
It definitely need unit test, that will check order of interfaces. I can do it. Do it need integration test like issue in #1036? If need how it should be implemented?

Unfortunately, I don't see how GetRuntimeInterfaceMap may help here.
It is really complicated now. As I don't see any case, where #1037 is not enough, probably only #1037 should be discussed/merged when it will be good.

lbargaoanu:
You're right, I can't make it fail if the interfaces don't inherit each other. So you can close this PR. The first should be enough. But that one needs a test too :) Like [this one] (https://github.com/AutoMapper/AutoMapper/blob/0e0e54d4f825b43225464d0d302b26876f39a1c8/src/UnitTests/Bug/IntToNullableDecimal.cs).

iskiselev:
@lbargaoanu, right now, if interface are not inherit each other, I try to sort them based on first implementation type relation in inheritance hierarchy. With your suggestion of  GetRuntimeInterfaceMap it will be sorted based on last type in inheritance hierarchy that implement/override at least one interface method.
I believe it may be incorrect for something like [this](https://gist.github.com/iskiselev/9a4212fb0e26ea7c7612) - I've added explicit method implementation to class2.

lbargaoanu:
Yes, but that seems fair.

iskiselev:
Ok, tomorrow will add test cases here and add PR with 3-rd based on Interface map so that you may choose what looks better :)

iskiselev:
I personally prefer compare based on first usage, probably with fallback to last usage if equals, but it is arguable.

iskiselev:
Pull request updated with test cases.

jbogard:
Merge conflicts - and is this different than #1062 ?

iskiselev:
Merge conflicts fixed. Not sure why CI have build failed, probably it should just be run again.
This and #1062 both try to sort interfaces that are tried to be mapped, so that most appropriate (usually most derived interface) was used.
This differs from #1062 in interface that will be preferred if two compared interface doesn't implement each other.
This PR will prefer interface that was implemented first in more derived concrete type, and #1062 will prefer interface with at least one method implemented in more derived concrete type (based on @lbargaoanu suggestion).

I personally prefer this PR, as #1062 is more complicated, but may wok better in very specific exotic edge cases.
I believe that both this and #1062 solve more problems than #1058.

jbogard:
I tend to agree - this one I can explain better. These are all pretty big edge cases. Can you fix merge conflicts (again)?

PR #983, opened 2015-11-12
Change MemberInfo to IValueResolver

TylerCarlson1:
Destination property rules can now go beyond just properties and fields

Constructor Property Maps now use what Property Maps use to resolve there expressions now.
That means Null substitution has a throw not implemented exception, but none of the test cases call this with the change, so don't know what to do about it.

#670's scenarios cannot work currently with convention system in place.

With configuration of
```
Mapper.Initialize(cfg =>
            {
                cfg.DefaultMemberConfig.AddMember<CultureInfoLocilization>();
                cfg.CreateMap<Question, QuestionDTO>();
            });
```
with class code of
```
public class CultureInfoLocilization : IChildMemberConfiguration
    {
        public bool MapDestinationPropertyToSource(IProfileConfiguration options, TypeDetails sourceType, Type destType,
            string nameToSearch, LinkedList<IValueResolver> resolvers, IMemberConfiguration parent)
        {
            var memberType = sourceType.PublicReadAccessors.FirstOrDefault(mi => IsEnumerable(mi, nameToSearch));
            if (memberType == null)
                return false;
            var itemType = memberType.GetMemberType().GetGenericElementType();
            var cultureInfoMember = itemType.GetProperties().First(p => p.GetMemberType() == typeof (CultureInfo));
            var searchProperty = itemType.GetProperty(nameToSearch);
            var firstMethodInfo = typeof(Enumerable).GetMethods().First(m => m.Name == "First" && m.GetParameters().Count() > 1).MakeGenericMethod(itemType);
            var sourceParam = Expression.Parameter(sourceType.Type, "s");
            Expression expression = Expression.PropertyOrField(sourceParam, memberType.Name);
            var lambdaParam = Expression.Parameter(itemType, "e");
            var callFirstExpression = Expression.Call(firstMethodInfo, expression,
                Expression.Lambda(
                    Expression.Equal(Expression.PropertyOrField(lambdaParam, cultureInfoMember.Name),
                        Expression.Constant(CultureContext.CurrentCulture)), lambdaParam));
            var searchPropertyOfFirstExpression = Expression.PropertyOrField(callFirstExpression, searchProperty.Name);
            var lambdaExpression = Expression.Lambda(searchPropertyOfFirstExpression, sourceParam);
            var delegateResolverType = typeof (DelegateBasedResolver<,>).MakeGenericType(sourceType.Type,searchProperty.GetMemberType());
[...]

PR #1056, opened 2016-01-28
Add DestinationType back into resolving member info conventions.

TylerCarlson1:
Still keep Property Destination Type.

Problem is in #953 which made the example possible in #983.  But in doing so another kind of member mapping that I had before is impossible now.

This is what I have now
```
public class ReverseSourceToDestinationNameMapperAttributesMember : ISourceToDestinationNameMapper
    {
        private static readonly IDictionary<TypeDetails, Dictionary<MemberInfo, IEnumerable<MapFromAttribute>>> Cache = new ConcurrentDictionary<TypeDetails, Dictionary<MemberInfo, IEnumerable<MapFromAttribute>>>();

        public MemberInfo GetMatchingMemberInfo(IGetTypeInfoMembers getTypeInfoMembers, TypeDetails typeInfo, Type destType, string nameToSearch)
        {
            var destTypeDetails = new TypeDetails(destType);
            if(!Cache.ContainsKey(destTypeDetails))
                Cache.Add(destTypeDetails, getTypeInfoMembers.GetMemberInfos(destTypeDetails).ToDictionary(mi => mi, mi => mi.GetCustomAttributes(typeof(MapFromAttribute), true).OfType<MapFromAttribute>()));

            var matchingMapTo = Cache[destTypeDetails].SelectMany(kp => kp.Value.Where(_ => string.Compare(kp.Key.Name, nameToSearch, StringComparison.OrdinalIgnoreCase) == 0)).FirstOrDefault();
            return matchingMapTo == null
                ? null
                : getTypeInfoMembers.GetMemberInfos(typeInfo)
                    .FirstOrDefault(mi => mi.Name == matchingMapTo.MatchingName);
        }
    }

[AttributeUsage(AttributeTargets.Field | AttributeTargets.Property)]
    public class MapFromAttribute : Attribute
    {
        public string MatchingName { get; private set; }

        public MapFromAttribute(string matchingName)
        {
            MatchingName = matchingName;
        }
    }
```

Before #953 it worked but after it doesn't cause destType is now the member info type instead of the typeMap destination type.

This change passes both over, so both custom member mappers can be made


TylerCarlson1:
Merged with development

TylerCarlson1:
not again

jbogard:
Yes sorry, I removed the Internal namespace, moved things to the Configuration and Execution namespaces.

TylerCarlson1:
Great merged and now the new Generics tests introduced fail, and I have no clue why.

jbogard:
The develop build was broken

TylerCarlson1:
Yea I eventually figured it was broken.  It's nice that after the merge for the fix it rebuilt the PR automaticly.

jbogard:
Thaaaaaat was not automatic, I had to go through all the failing builds and hit "rebuild PR" in AppVeyor.

jbogard:
Do you have a test that might show how to use this?

TylerCarlson1:
Yea I can add an example in the Original Post to the PR with passing tests.  Also while I'm at it I can add #983's example as well, to show how they both use different types.

The project that uses this convention uses LinqToSQL so the DataContext objects are Auto-Generated, so I can't add `MapToAttribute` their properties.

jbogard:
Oh that would be nice, thanks!

TylerCarlson1:
Added unit tests.  Turns out I don't have an example that needs destination member type, but that doesn't mean it shouldn't be in there.  You might want to filter localization filter to only work if mapping strings to strings and nothing else, in which case you need the destination type.

Also I figured out although the localization works with `Map` it doesn't work with `ProjectTo`.  There's no strategy that works with `DelegateBasedResolver<,>`; I get a null exception.
I might be able to make one that would add projection test to this PR, or make a new one.  I think it's kind of the point for localization convention to usable in `ProjectTo`, otherwise there's no gain.

TylerCarlson1:
Found a work around, but the problem and/or solution is bypassing `CustomExpression` for Expression Translation.  I personally would like to take this further and get rid of `CustomExpression` altogether.  Enough if statements in weird places as it is now.  I'll see if I can remove them.

Solution is passing LambdaExpression in DelegateBasedResolver, and making a new interface to hold the lambda expression.  Using this add a new IExpressionResultConverter that handles this type and fixes the ProjectTo problem.

TylerCarlson1:
Submitted the fix for `ProjectTo` to work, with projection in each test.
The concept of a delegate expression resolver  needs to be in the system, because without it all these custom resolvers can't harness linq projection.  So with this DelegateBasedResolver would be the preferred method for making custom member resolver expressions.

Don't know if either of these would ever be considered part of AutoMapper to be default functionality.  Maybe they should be, personally I'm fine with making my own and re-using as needed (thx Fixie).  Still I would be interested in implementing other convention ideas if anyone has any.

I also don't know if it's worth it to even attempt to make a wiki explaining how to do this stuff.  It's so complex and deep into the systems, that a person would have to ask for it, or be good enough to go in and figure it all out for themselves.

P.S: I looked at getting rid of custom expression and it's not worth the hassle and there's no real benefit.

jbogard:
Re: resolvers. I would love to go in and try to make everything a delegate/func based system, which would make this sort of stuff easier, the whole value resolver thing is just a little too complicated.

jbogard:
Can you re-open this PR?

TylerCarlson1:
Can't open this one and new PR is saying no differences between this and development

TylerCarlson1:
I need to rebase or something to re-submit this.  Time to improve my Git-Fu

PR #1059, opened 2016-01-29
Compile the mapping lambda for the constructor argument; closes #1047

### LINKED ISSUE #1047: Problem with mapping constructor arguments from nested classes
Here is an example:
```
public class Destination
{
    public Destination(string data)
    {
        Data = data;
    }
    public string Data { get; private set; }
}

public class NestedSource
{
    public string Data { get; set; }
}

public class SourceWrap
{
    public NestedSource NestedSource { get; set; }
}

public class AutomapperTest
{
    public void Run()
    {
        Mapper.CreateMap<SourceWrap, Destination>()
            .ForCtorParam("data", opt => opt.MapFrom(s => s.NestedSource.Data))
            .IgnoreAllPropertiesWithAnInaccessibleSetter();
        var source = new SourceWrap
        {
            NestedSource = new NestedSource { Data = Guid.NewGuid().ToString() }
        };
        var destination = Mapper.Map<Destination>(source);
    }
}
```

Exception will be thrown on the last line of code.
I investigated source code and found out that problem occurs when class `CtorParamConfigurationExpression` is parsing `sourceMember` expression. It passes expression to `MemberInfoFinderVisitor`, which enumerates expression tree and creates `PropertyGetter` for each `PropertyAccess` node. Problem is that `ExpressionVisitor` enumerates expression in this order (from example before):
`Lambda`
`.`
`Data`
`.`
`NestedSource`
`s`
In the same order properties passed to `ConstructorParameterMap`, which latter applies them to source model. During mapping it tryes to access property `Data` of class `SourceWrap` and fails.
I think if you reverse `Members` list of `MemberInfoFinderVisitor` before passing it to `ConstructorParameterMap` it will partially solve the problem.
But there are still a problem if someone tryes to do something like that:
`.ForCtorParam("data", opt => opt.MapFrom(s => s.NestedSource.Data + s.SomeOtherData))`
So you probably should check if `sourceMember` expression is only accessing properties. Or just pass whole expression to `ConstructorParameterMap` and then compile it to access data from source member.

lbargaoanu:


jbogard:
Merge conflicts :(

lbargaoanu:
Updated.

PR #1058, opened 2016-01-29
Order interfaces by type hierarchy; closes #1016 closes #1036

lbargaoanu:

PR #1062, opened 2016-01-30
Added sorting of interfaces (3)

iskiselev:
Here is more complicated version, that tries to fix problem that @lbargaoanu noted for original (#1037) version and use GetRuntimeInterfaceMap(per suggestion in #1050).

This version will iterate interfaces in such order, that more derived interfaces will be before less derived (for IInterfaceBase:IInterfaceDerived, order will be IInterfaceDerived, IInterfaceBase), and additional prefer interface that have mthod implementation in more derived types.

I'm not sure which option is better (this or #1050), but both of them fix more issues than #1037 or #1058.

### DIFF
--- a/src/AutoMapper/MapperConfiguration.cs
+++ b/src/AutoMapper/MapperConfiguration.cs
@@ -408,7 +408,9 @@ namespace AutoMapper
                 baseType = baseType.BaseType();
             }
 
-            foreach (var interfaceType in type.GetTypeInfo().ImplementedInterfaces)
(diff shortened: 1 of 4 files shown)


jbogard:
Merge conflicts :(

iskiselev:
Merge conflicts were solved. I'm not sure why CI build fails. Probably #1050 should be preferred, as #1050 a little bit easier.

jbogard:
I took #1050

PR #1079, opened 2016-02-03
Fixes #1077

### LINKED ISSUE #1077: 4.2.0 .Include<DerivedSourceType, DerivedTargetType> throws exception with base type names
Creating a map and including a map which doesn't exist  throws an exception that doesn't really help find the missing map.

Consider the following:

`         

        public class SourceDerived
        {
        }

        public interface IMyInterface
        {
        }
       
        public class TargetDerived : IMyInterface
        {
        }

        [TestMethod]
        public void MyTestMethod()
        {
                cfg.CreateMap<object, IMyInterface>()
                    .Include<SourceDerived, TargetDerived>()
                );
        }
`
The test throws an exception with message : 

> System.InvalidOperationException: Missing map from Object to IMyInterface. Create using Mapper.CreateMap\<Object, IMyInterface>.

It is right to throw because I have not defined a map from SourceDerived to TargetDerived. 

In my opinion it would be much more helpful if the message read: 

> System.InvalidOperationException: Missing map from SourceDerived to TargetDerived. Create using Mapper.CreateMap\<SourceDerived , TargetDerived>.

@jbogard If you agree, I would happily provide a PR

lbargaoanu:
#1077

### DIFF
--- a/src/AutoMapper/MapperConfiguration.cs
+++ b/src/AutoMapper/MapperConfiguration.cs
@@ -328,11 +328,12 @@ namespace AutoMapper
 
         private IEnumerable<TypeMap> GetDerivedTypeMaps(TypeMap typeMap)
         {
-            foreach (var derivedMap in typeMap.IncludedDerivedTypes.Select(FindTypeMapFor))
+            foreach (var derivedTypes in typeMap.IncludedDerivedTypes)
             {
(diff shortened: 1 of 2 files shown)


jbogard:
Merge conflict :(

lbargaoanu:
Updated.

PR #1085, opened 2016-02-05
Read only profiles

jbogard:
Moving all configuration to the initialization, sealing made private/internal. All config is done after creating a `MapperConfiguration`.

Recording configuration actions and playing them back during initialization for all config parts.

@TylerCarlson1 @lbargaoanu this is a big one for sure, completely overhauls the internal config

Still to be done is get rid of the whole Internal namespace and move to the correct places.


src/AutoMapper/IMappingExpression.cs
```diff
@@ -304,6 +284,13 @@ public interface IMappingExpression<TSource, TDestination>
         IMappingExpression<TSource, TDestination> BeforeMap(Action<TSource, TDestination> beforeFunction);
 
         /// <summary>
+        /// Execute a custom function to the source and/or destination types before member mapping
+        /// </summary>
+        /// <param name="beforeFunction">Callback for the source/destination types</param>
+        /// <returns>Itself</returns>
+        IMappingExpression<TSource, TDestination> BeforeMap(Action<TSource, TDestination, ResolutionContext> beforeFunction);
```
TylerCarlson1: Yay.  Can now get IMapper instead of having to static instance it out and pass it over.

Also I'm not sure.  Do you even need `TSource` and `TDestination` for the action since you can figure it out from ResolutionContext?  Or can they be different with inheritance playing a factor in the context?  I don't think this is a big deal either way.

PR #1089, opened 2016-02-08
Fix the generics test

lbargaoanu:
The build was broken because of [this](https://github.com/AutoMapper/AutoMapper/blob/5222f5cf13bb6dcc8117f6657647e8bec241e510/src/UnitTests/MappingInheritance/GenericsAndInterfaces.cs) test.
Generics and interfaces bug 944

### DIFF
--- a/src/AutoMapper/TypeDetails.cs
+++ b/src/AutoMapper/TypeDetails.cs
@@ -81,13 +81,14 @@ namespace AutoMapper
                 genericInterfaces.Add(Type);
 
             explicitExtensionMethods.AddRange(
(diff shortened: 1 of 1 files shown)


jbogard:
Ah I was juuuuuust about to revert that other commit :)

lbargaoanu:
:) Profile.IncludeExtensionMethods adds all extension methods from an assembly. That's why all kind of methods get to test that query. Per type seems more appropriate.

jbogard:
That makes sense. C# does it by namespace, but by type is most explicit and obvious.

PR #1090, opened 2016-02-08
Restore some static api

jbogard:
Configuration is still readonly, there's no Mapper.Reset. Just Mapper.Initialize. Also restored the queryable extension methods with overloads that grab the static versions.

Downside is I had to keep the explicit implementations of IMapper on Mapper class. To balance that, I put the factory method for CreateMapper on the IConfigurationProvider interface (you can't do `new Mapper()` and start doing stuff, those are static methods).

@lbargaoanu and @TylerCarlson1 this should be a decent compromise, I think.

### DIFF
--- a/src/AutoMapper/IConfigurationProvider.cs
+++ b/src/AutoMapper/IConfigurationProvider.cs
@@ -121,5 +121,7 @@ namespace AutoMapper
         bool AllowNullCollections { get; }
 
         IExpressionBuilder ExpressionBuilder { get; }
+        IMapper CreateMapper();
+        IMapper CreateMapper(Func<Type, object> serviceCtor);
     }
 }
--- a/src/AutoMapper/IMapper.cs
+++ b/src/AutoMapper/IMapper.cs
@@ -107,7 +107,7 @@ namespace AutoMapper
 
 
         /// <summary>
-        /// Configuration provider for performaing maps
+        /// Configuration provider for performing maps
         /// </summary>
         IConfigurationProvider ConfigurationProvider { get; }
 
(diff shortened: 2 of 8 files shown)


TylerCarlson1:
I wonder if anyone will complain about `var mapper = new Mapper();` and I can't map anything why is that?  Since they are all explicit implementations now

jbogard:
Probably. Can't have both though, so hopefully the factory method is good enough.

Removing the static API forced me to overhaul the configuration internals, so I'm still somewhat happy I did it, even if I am putting some of it back in. The most important thing is the configuration is done once.

TylerCarlson1:
Yea the static was never the problem, except CreateMap.  Also I think you have to do it as instances no matter what, else how are you going to do all those unit tests where you Reset all the time?

MattHoneycutt:
This is great news, and exactly what I was hoping for. :)

jbogard:
@TylerCarlson1 I'm keeping everything going through the instance maps on the testing side. The idea being the static part is so small it won't really matter.

lbargaoanu:
Separate classes would also work, with the one you care about more being named Mapper :) A clean break would be to rename the static class.

jbogard:
What I figure is that people that want to use instances will use DI, so they'll always have an IMapper injected. Could be a broad assumption there.

eddiepaz:
Great news, indeed. I put off upgrading to let such as huge change settle down - too much code to change for me.

I do have a question, however, if we want to use Mapper.Map(), we must use Mapper.Initialize(); and to use IMapper (DI), we must new up MapperConfiguration() and use its CreateMapper(), is this correct?

PR #1091, opened 2016-02-08
Broke out configuration validation from the mapper configuration

jbogard:
MapperConfiguration was geting a little unwieldy with the validation in there


TylerCarlson1:
I'm all for breaking those god like classes up into smaller sections.  So hard to know what goes to what functionality with these big classes that do multiple things.

PR #1096, opened 2016-02-09
Add DestinationType back into resolving member info conventions. Part 2

TylerCarlson1:
Now Have ExpressionBasedResolver to distinguish the two resolver types

### DIFF
--- a/src/AutoMapper/Configuration/Conventions/CaseInsensitiveName.cs
+++ b/src/AutoMapper/Configuration/Conventions/CaseInsensitiveName.cs
@@ -6,7 +6,7 @@ namespace AutoMapper.Configuration.Conventions
 
     public class CaseInsensitiveName : ISourceToDestinationNameMapper
     {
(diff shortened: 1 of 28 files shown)


TylerCarlson1:
I think all IValueResolvers should be IExpressionResolver except for the DelegateBaseResolver which can be IValueResolver and all the IMemberGetters and other property and field based resolvers can be converted into ExpressionResolverBase by passing in a generated lambda expression.  That way you should only have to do a parameter naming ExpressionVisitor and get rid of a lot of expression building logic.

TylerCarlson1:
DelegateBaseResolvers are the only one's excluded because they are primarily used for ResolveUsing, which uses ResolutionContext, which can't be used in Expression translation in ProjectTo.
Everything else that only depends on source and destination can be made into an expression, and that expression can then be compiled for normal mapping, while the expression can be used directly in ProjectTo.

PR #1099, opened 2016-02-10
Fixed issue #1905 mapping nullable type to nullable type

jamesloi:
#1905 Resolving a nullable decimal to a nullable decimal throws Missing type map configuration or unsupported mapping at runtime. version 4.2.0. Aslo adding unit test.

### DIFF
--- a/src/AutoMapper/Mappers/MapperRegistry.cs
+++ b/src/AutoMapper/Mappers/MapperRegistry.cs
@@ -26,6 +26,7 @@ namespace AutoMapper.Mappers
 #if !PORTABLE
             new TypeConverterMapper(),
 #endif
+            new NullableMapper(),
             new NullableSourceMapper(),
             new ImplicitConversionOperatorMapper(),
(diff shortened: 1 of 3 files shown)


jbogard:
Build failed :(

lbargaoanu:
This was handled in a different way in #1093.

jamesloi:
@lbargaoanu was that merged back into the master? As the issue was still happening for me with the latest from master.

@jbogard fixed the build issue.

lbargaoanu:
Try the MyGet build.

jbogard:
@jamesloi it was merged into develop, not master. Master branch is the NuGet release branch, develop is MyGet.

billrob:
FWIW, I don't think this is a good fix. The issue is the way ResolveUsing is handling nulls.  Not that is has to do with a nullable<T>.  I suspect if the non nullable<T> bug is fixed, the nullable<T> will get resolved as well.

jbogard:
Yeah, this is fixed in the 5.0 release now.

Issue #1083, opened 2016-02-05
Stackoverflow using AfterMap with custom collection types


long2know:
I have a custom collection that I use for pagination.  It inherits from `List<T>` like so:

```
public class PaginatedList<T> : List<T>
{
    public int PageNumber { get; set; }
    public int PageSize { get; set; }
    public int TotalCount { get; set; }
    public int TotalPages { get; set; }
    public string SortBy { get; set; }
    public string SortDirection { get; set; }
}
```

With version 3.x of AutoMapper, I would use an extension method to deal with instances of including a map of `PaginatedList<Src>` to `PaginatedList<Dest>` like this:

`Mapper.CreateMap<Type1, Type2>().IncludePagedResultMapping();`

The extension method code used AfterMap after specifying the paging members:

```
public static IMappingExpression<TSrc, TDest> IncludePagedResultMapping<TSrc, TDest>(this IMappingExpression<TSrc, TDest> expression)
{
    Mapper.CreateMap<PaginatedList<TSrc>, PaginatedList<TDest>>()
        .ForMember(dest => dest.PageNumber, opt => opt.MapFrom(src => src.PageNumber))
        .ForMember(dest => dest.PageSize, opt => opt.MapFrom(src => src.PageSize))
        .ForMember(dest => dest.TotalCount, opt => opt.MapFrom(src => src.TotalCount))
        .ForMember(dest => dest.TotalPages, opt => opt.MapFrom(src => src.TotalPages))
        .ForMember(dest => dest.SortBy, opt => opt.MapFrom(src => src.SortBy))
        .ForMember(dest => dest.SortDirection, opt => opt.MapFrom(src => src.SortDirection))
        .AfterMap((s, d) => AutoMapper.Mapper.Map<List<TSrc>, List<TDest>>(s, d));
    return expression;
}
```

With version 4.2 of AutoMapper, I have rewritten this extension method to take in the current configuration object and utilize a static reference to the config object in the AfterMap.  It looks like this now:

```
public static class AutoMapperConfig
{
	private static bool _mapped;
	private static IMapper _mapper;
	private static MapperConfiguration _config;
	public static IMapper Mapper { get { return _mapper; } }

	public static void RegisterMappings()
	{
		if (_mapped == false)
		{
			_mapped = true;
			_config = new MapperConfiguration(cfg => Configure(cfg));
			_mapper = _config.CreateMapper();
		}
	}
	
	private static void Configure(IMapperConfiguration cfg)
[...]

PR #1110, opened 2016-02-15
Return the generic maps whenever there is a match

lbargaoanu:
Helps with #1083.


TylerCarlson1:
Further muddles the lines between FindTypeMap and ResolveTypeMap.

Also if you are gonna do this maybe you should add Conventions logic in there as well, because that's going to have the same problems as generic type maps have when it comes to resolving things.

lbargaoanu:
I think this was fixed in another PR.

Issue #1015, opened 2016-01-07
Multiple sources of the same type to one destination - aka $.extend


andrewm1986:
### TL;DR

jQuery.extend (https://api.jquery.com/jquery.extend/) like functionality would be a really good addition to AutoMapper.

### Detail
I quite often have a need to merge multiple objects of the same type back into either the same type or a different type.

I have an example now where I need to merge multiple objects of type `Insurance` into my `InsuranceViewModel`

```c#
public class Insurance 
{
    public decimal? Premium { get; set; }
    public decimal? OutOfPocket { get; set; }
}

public class InsuranceViewModel
{
    public decimal? Premium { get; set; }
    public decimal? OutOfPocket { get; set; }
}
```

Behind the scenes a user's insurance level is an aggregation of different insurances that may or may not have various properties set.

Currently with AutoMapper I would do something like this

```c#
var insurance1 = new Insurance { Premium = 123 };
var insurance2 = new Insurance { OutOfPocket = 456 };

InsuranceViewModel viewModel = Mapper.Map<InsuranceViewModel>(insurance1);
Mapper.Map<Insurance, InsuranceViewModel>(insurance2);
```

However, in my example that would result in `viewModel` having a `Premium` of `null`, because it overwrites the `Premium` value from `insurance1`.

I suggest a `Mapper.Merge` method that works exactly the same as `Mapper.Map` but with no differences

1. Overloads for accepting an array of source.
2. When the value it is going to overwrite the destination property with is `null` it does not overwrite the value.

This would allow me to write the above as

```c#
var insurance1 = new Insurance { Premium = 123 };
var insurance2 = new Insurance { OutOfPocket = 456 };

InsuranceViewModel viewModel = Mapper.Merge<InsuranceViewModel>(new[] { insurance1, insurance2 });

// or

InsuranceViewModel viewModel = Mapper.Merge<InsuranceViewModel>(insurance1, insurance2);


```

`viewModel` would now have `Premium` set to `123` and `OutOfPocket` set to `456`

If this is something that the project wants then I would be happy to look at doing the work myself.  I would quite enjoy it.

jbogard:
Give it a go! I like using a different named mapping method.

andrewm1986:
Rightio.  I'll open a PR

lbargaoanu:
@andrewm1986 Do you still want to do this?

andrewm1986:
Sorry, I've been out of he country for the past few weeks!

[...]

PR #1122, opened 2016-02-18
Multiple sources of the same type to one destination - aka $.extend

andrewm1986:
Referencing issue #1015 

Basic unit tests and enough scaffold code to compile - does not work and fails tests.

This should give you an idea of what I am imagining before I write the actual code.

How does this seem?

### DIFF
--- a/src/AutoMapper/IMapper.cs
+++ b/src/AutoMapper/IMapper.cs
@@ -32,6 +32,15 @@ namespace AutoMapper
         TDestination Map<TSource, TDestination>(TSource source);
 
         /// <summary>
+        /// Execute a merge mapping from the source object(s) to a new destination object.
+        /// </summary>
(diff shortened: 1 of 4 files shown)


jbogard:
I like it! But would all sources be the same type? What about two disparate
sources?

andrewm1986:
@jbogard that had never occurred to me actually

I guess conceptually there's no reason why you wouldn't want that to be the case.

I'll make sure that's one of my unit tests

andrewm1986:
@jbogard What's your opinion on whether it should be part of IMapper as additional Merge methods or part of it's own interface, IMerge?

jbogard:
IMapper is good, I've only really broken out the LINQ stuff into its own thing.

jbogard:
Closing this PR for now, looks good if you want to implement!

PR #1130, opened 2016-02-21
Small changes to expression-binding behaviour II

jasonholloway:
Hi, this is a simplified version of #852, this time skipping the gratuitous refactoring (and also some of the side issues, which I'll treat separately).

Basically, this PR: 
* Moves the `AssignableExpressionBinder` down the list of binders, so that it allows other binding behaviours a chance to do their stuff when projecting like to like
* Fixes up `EnumerableExpressionBinder` to add null-guards when `config.AllowNullCollections` is set to true - nulls were previously always copied over by `AssignableExpressionBinder`

### DIFF
--- a/src/AutoMapper/QueryableExtensions/ExpressionBuilder.cs
+++ b/src/AutoMapper/QueryableExtensions/ExpressionBuilder.cs
@@ -31,10 +31,10 @@ namespace AutoMapper.QueryableExtensions
         private static readonly IExpressionBinder[] Binders =
         {
             new NullableExpressionBinder(),
-            new AssignableExpressionBinder(),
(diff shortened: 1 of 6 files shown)


lbargaoanu:
Some tests fail.

jasonholloway:
You're not wrong. I'm going to remove the MappedTypeExpressionBinder 'fix', which is causing EF to error on complex types, and is pretty tangential to this PR anyway.

jasonholloway:
MappedTypeExpressionBinder should cast to object though, it's just that doing so exposes problems elsewhere. If null-guards could be turned off for complex types (as they should be) everything would be fine.

PR #1136, opened 2016-02-24
Make TypePair a struct

lbargaoanu:
[Before](https://cloud.githubusercontent.com/assets/4517428/13281693/e85e1e18-daed-11e5-8920-7def940347d8.png) & [after](https://cloud.githubusercontent.com/assets/4517428/13282395/70fb9310-daf1-11e5-941d-65d0ece6f158.png).
We could try the same for ResolutionContext and ResolutionResult. It's true that the ResolutionContext is too big right now.
[The test project](https://github.com/lbargaoanu/AutoMapperBenchmark).

### DIFF
--- a/src/AutoMapper/MapperConfiguration.cs
+++ b/src/AutoMapper/MapperConfiguration.cs
@@ -351,7 +351,7 @@ namespace AutoMapper
 
             var typeMap = _profiles
                 .Cast<IProfileConfiguration>()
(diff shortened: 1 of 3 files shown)


jbogard:
I would love to get rid of ResolutionResult, I don't think it provides much value TBQH. You can get the value back, and the type itself I would want to have figured out from the metadata

lbargaoanu:
I've [started](https://github.com/lbargaoanu/AutoMapper/tree/RemoveResolutionResult) removing the ResolutionResult.

PR #1145, opened 2016-02-29
Remove ResolutionResult

lbargaoanu:


jbogard:
Looks good! I think after this one, I'll follow up with one that has a generic IValueResolver instead of the base class. Then I can do the work of wrapping the generic with the non-generic version.

jbogard:
Oh and this resulted in about a 15-20% performance boost in my benchmarks

lbargaoanu:
I guess the fastest code is the one that doesn't run :)

PR #1148, opened 2016-02-29
Refactor resolution context

jbogard:
This removes most of the complexity of resolution context, and moves the creation of these for the most part to inside of the Mapper class. No one should care about the MapperEngine thingy any more.

Also, I refactored ITypeConverter to be generic by default and got rid of the abstract base class


jbogard:
@TylerCarlson1 @lbargaoanu hopefully this simplifies things. It was always weird to have to create a ResolutionContext just to map something and all those factory methods were....weird.

TylerCarlson1:
Looks like I have some homework to do tonight.

lbargaoanu:
I think Types should go :) It's the same thing stored twice. Maybe a method would help.

PR #1155, opened 2016-03-02
Reuse the context whenever possible

lbargaoanu:


jbogard:
So this is a wild idea, and I feel like it's moving this direction anyway.

Types are used to figure out what mapping action to take. Most of the mapping actions we already know what we should take based on the source/destination types in property maps. Why not pre-wire during config time? If a lot of things just go to the AssignableMapper, why go the route of Mapper -> MappingEngine -> ConfigurationProvider when the answer to how to property A to B is known?

The times when this wouldn't work are when the runtime type doesn't match the compile-time type, AND those runtime types would result in a different mapping path.

I still like this PR, but I think the eventual end result would be to pare down the ResolutionContext to not include any source/destination value/types.

TylerCarlson1:
From what I see it doesn't look like this PR isn't prewiring anything.  It's just saving the step of going back to the top level to get a new resolution context that's going to be the equivalent of what you have at the point of mapping..

lbargaoanu:
I think smth like this needs to happen as long as the ResolutionContext remains a class and we need more than one context (for each property or for each collection item). No matter how the code will look like if we have an explicit type map I presume we need the context and the problem remains. I think :)

lbargaoanu:
@jbogard I think this still applies (we still allocate a context for each item in the collection) but I see you've rolled back the similar stuff I did for the enumerable mapper, so close this then?

jbogard:
I think so too, I just held back on it because we were looking at doing a bit of refactoring around some of the enumerable mappers and it was easier to unwind it.

PR #1158, opened 2016-03-04
[WIP] Remove type map registry

jbogard:


jbogard:
@TylerCarlson1 This is what I meant by the collapsing of all that indirection first, I think I was a little to pattern-happy in the early days of AutoMapper lol

TylerCarlson1:
Yea weren't we all.  Never noticed the list of one there.
I wonder if anyone has added to that list every and will complain about this missing feature?

TylerCarlson1:
I bet everyone who might of has already went and made their own "Superior" mapper class with more speed and less functionality.

TylerCarlson1:
I see what you are doing here.  I'm not a fan of the concept of if statements like that, but speed is speed.

Also for speed's [sake](https://github.com/AutoMapper/AutoMapper/commit/a727ef1f24d114823ae456a5ef1ebf3641f14b72#diff-8cd4ef65cc031d5beedd2778ea6946f6R118)
```
if (!CanResolveValue())
{
    _mapperFunc = (mappedObject, context) =>{};
    return;
}
```

TylerCarlson1:
If you want to push it to dev I can modify my GenericIValueResolver branch to put the IMemberResolvers to have expressions and use them to generate the expression method.  It shouldn't hurt the code in any way as is now.

Or I can join the PR a PR club.

jbogard:
No yeah it's just a starting point. PR a PR is fine, I'd like to get the
expression stuff in.

src/AutoMapper/MappingEngine.cs
```diff
@@ -51,5 +49,72 @@ public object Map(ResolutionContext context)
                 throw new AutoMapperMappingException(context, ex);
             }
         }
+
+        private object Map(object source, ResolutionContext context)
```
lbargaoanu: This code looks like it belongs in TypeMap.

src/AutoMapper/PropertyMap.cs
```diff
@@ -112,6 +115,39 @@ internal void Seal()
                 (inner, res) => ctxt => res.Resolve(inner(ctxt), ctxt));
 
             SourceType = resolvers.OfType<IMemberResolver>().LastOrDefault()?.MemberType;
+
+            if (resolvers.All(r => r is IMemberResolver)
```
lbargaoanu: Maybe IExpressionResolver here? IMemberResolver has the type and I needed the type to remove the ResolutionResult so now the DelegateBasedResolver implements IMemberResolver :)

lbargaoanu:
Unrelated, I tried running the tests in parallel and it is faster.

src/AutoMapper/MappingEngine.cs
jbogard: It certainly could, I think that's the next step towards pre-wiring a mapping plan for a type map. Good call.

src/AutoMapper/PropertyMap.cs
jbogard: Ahhhhh yes that makes sense.

TylerCarlson1:
I'm a good percentage of the way done with a PR.  The problem right now making all the tests pass.  The if statement to know if should action resolve is not complete.  Have to add IEnumerable check cause that can't auto map, and possibly some other stuff

jbogard:
Lol I was just seeing that.

Right now just going down the existing path (not using expressions, just
short circuiting decisions) it's like 5x faster. Expressions should help
that last little bit.

One thing the property getter resolvers do is null propagated checks. Which
the operator ?. is not supported in expressions or compiled. You'll have to
do src==null ? null : src.Value, all the way down.

jbogard:
Oh and doing null to default values for value types.

TylerCarlson1:
Yea short Circuiting decisions is the Biggest performance improvement.  The Expression mappings as I had them now are giving an additional 5% on top of short circuiting.  I'm trying to figure out right now is the edge cases with just pure short circuiting Map recursion call.  AllowNullDestinations = false is giving most the issues right now.

TylerCarlson1:
That last commit is what I'm trying to do right now.  Except I'm translating the expressions. and use that for _valueResolverFunc

TylerCarlson1:
Yea those null checks sucks.  Also you can't do that for structs because can't do comparisons that way.  Last thing is NullReferenceExceptionSwallowingResolver to Visit and wrap the null exceptions.

I might be able to get around 2-3 failed tests, then I'll PR against this.  Mostly problems with Generic Types can't make expressions without a type present or something.

jbogard:
@TylerCarlson1 I like it. I'm pushing a change to get the TypeMap available in the PropertyMap, and the profile config available in TypeMap.

One thing to help get rid of the casts is if we create the expression as a strongly typed one and cast the input at the top:

```
var expr = (ResolutionContext ctxt, object dest, Func<Foo, FooDto> inner) => inner((Foo)ctxt.SourceValue, (FooDto)dest));  
```

Basically what the old generic value resolver abstract class did, but in expression form. Something like that.

TylerCarlson1:
The getter and setter expressions are already like that.  The only real problem is merging the two together into an expression, and account for nullable values.

Also I would like to make the PropertyMaps and TypeMaps Generic as well so that the casting happens once at the type map level, and not for every single property setter.  But that's for another PR perhaps.  Also if you use Generic calls for Map there's no need to cast at all.

jbogard:
Ah, I gotcha now. Yeah I don't know if that would really be possible,
making those generic would be really tough since they're config APIs.
Perhaps those expressions could be bubbled up to the TypeMap level.

src/AutoMapper - Copy/Execution/IDelegateFactory.cs
```diff
@@ -0,0 +1,20 @@
...
+    public delegate void LateBoundPropertySet<TSource,TValue>(TSource target, TValue value);
+
+    public delegate void LateBoundValueTypeFieldSet(ref object target, object value);
+
+    public delegate void LateBoundValueTypePropertySet(ref object target, object value);
+
+    public delegate object LateBoundCtor();
+
+    public delegate object LateBoundParamsCtor(params object[] parameters);
+}
```
lbargaoanu: I don't think there's much value in having different delegates with the same signature.

src/AutoMapper - Copy/Execution/PropertyAccessor.cs
```diff
@@ -0,0 +1,28 @@
...
+
+    public class PropertyAccessor<TSource, TValue> : PropertyGetter<TSource, TValue>, IMemberAccessor
+    {
+        private readonly Lazy<Expression<LateBoundPropertySet<TSource, TValue>>> _lateBoundPropertySetExpression;
+        private readonly Lazy<LateBoundPropertySet<TSource, TValue>> _lateBoundPropertySet;
+
+        public PropertyAccessor(PropertyInfo propertyInfo)
+            : base(propertyInfo)
+        {
+            var HasSetter = propertyInfo.GetSetMethod(true) != null;
```
lbargaoanu: It should be hasSetter.

src/AutoMapper/Execution/IDelegateFactory.cs
```diff
@@ -1,14 +1,14 @@
...
-    public delegate object LateBoundFieldGet(object target);
+    public delegate TValue LateBoundFieldGet<TSource, TValue>(TSource target);
 
-    public delegate void LateBoundFieldSet(object target, object value);
+    public delegate void LateBoundFieldSet<TSource, TValue>(TSource target, TValue value);
 
-    public delegate void LateBoundPropertySet(object target, object value);
+    public delegate void LateBoundPropertySet<TSource,TValue>(TSource target, TValue value);
 
     public delegate void LateBoundValueTypeFieldSet(ref object target, object value);
```
lbargaoanu: I don't think there's much value in having different delegates with the same signature.

src/AutoMapper/Execution/PropertyAccessor.cs
```diff
@@ -1,24 +1,28 @@
...
 
         public PropertyAccessor(PropertyInfo propertyInfo)
             : base(propertyInfo)
         {
             var HasSetter = propertyInfo.GetSetMethod(true) != null;
-            _lateBoundPropertySet = HasSetter
-                ? new Lazy<LateBoundPropertySet>(() => DelegateFactory.CreateSet(propertyInfo))
-                : new Lazy<LateBoundPropertySet>(() => (_, __) => { });
+            _lateBoundPropertySetExpression = HasSetter
+                ? new Lazy<Expression<LateBoundPropertySet<TSource, TValue>>>(() => DelegateFactory.CreateSet<TSource, TValue>(propertyInfo))
```
lbargaoanu: These types with multiple angle brackets could be made more readable by wrapping them inside a struct at zero cost.

src/AutoMapper/Execution/ReflectionHelper.cs
```diff
@@ -136,47 +136,78 @@ public static MemberInfo FindProperty(LambdaExpression lambdaExpression)
...
+                        Activator.CreateInstance(
+                            typeof (FieldGetter<,>).MakeGenericType(accessorCandidate.DeclaringType,
+                                accessorCandidate.GetMemberType()), accessorCandidate) as IMemberGetter;
+
+                if (accessorCandidate is MethodInfo)
+                    return
+                        Activator.CreateInstance(
+                            typeof (MethodGetter<,>).MakeGenericType(accessorCandidate.DeclaringType,
+                                accessorCandidate.GetMemberType()), accessorCandidate) as IMemberGetter;
+
```
lbargaoanu: Duplicate code.

src/AutoMapper/Execution/ReflectionHelper.cs
```diff
@@ -136,47 +136,78 @@ public static MemberInfo FindProperty(LambdaExpression lambdaExpression)
...
+                return null;
+            }
+            catch (Exception)
```
lbargaoanu: This can't be good :)

src/AutoMapper/Execution/ReflectionHelper.cs
```diff
@@ -136,47 +136,78 @@ public static MemberInfo FindProperty(LambdaExpression lambdaExpression)
...
+
+                var propertyInfo = accessorCandidate as PropertyInfo;
+                if (propertyInfo != null)
+                    return accessorCandidate.DeclaringType.IsValueType()
+                        ? Activator.CreateInstance(
+                            typeof (ValueTypePropertyAccessor<,>).MakeGenericType(accessorCandidate.DeclaringType,
+                                accessorCandidate.GetMemberType()), accessorCandidate) as IMemberAccessor
+                        : Activator.CreateInstance(
+                            typeof (PropertyAccessor<,>).MakeGenericType(accessorCandidate.DeclaringType,
+                                accessorCandidate.GetMemberType()), accessorCandidate) as IMemberAccessor;
```
lbargaoanu: Duplicate code.

src/AutoMapper/Execution/ReflectionHelper.cs
```diff
@@ -136,47 +136,78 @@ public static MemberInfo FindProperty(LambdaExpression lambdaExpression)
...
+
+                return null;
+            }
+            catch (Exception)
```
lbargaoanu: This can't be good :)

src/AutoMapper/PropertyMap.cs
```diff
@@ -107,11 +117,62 @@ internal void Seal()
...
+            
+            if (resolvers.All(r => r is IMemberGetter)
+                && CanResolveValue()
+                && !_hasPreCondition
+                && !_hasCondition
+                && SourceType != null
+                && (!SourceType.IsEnumerableType() || SourceType == typeof(string))
+                && typeMapRegistry.GetTypeMap(new TypePair(SourceType, DestinationPropertyType)) == null
+                && ((EnumMapper.EnumToEnumMapping(new TypePair(SourceType, DestinationPropertyType)) && !EnumMapper.EnumToNullableTypeMapping(new TypePair(SourceType, DestinationPropertyType))) || !EnumMapper.EnumToEnumMapping(new TypePair(SourceType, DestinationPropertyType)))
+                && DestinationPropertyType.IsAssignableFrom(SourceType))
```
lbargaoanu: This condition deserves it's own method :)

src/AutoMapper/PropertyMap.cs
```diff
@@ -107,11 +117,62 @@ internal void Seal()
...
+            {
+                var expression2 = resolvers.OfType<IMemberResolver>().Aggregate<IMemberResolver, LambdaExpression>((Expression<Func<ResolutionContext, object>>)
+                    (ctxt => ctxt.SourceValue),
+                    (expression, resolver) => new ExpressionConcatVisitor(resolver.GetExpression).Visit(expression) as LambdaExpression);
+                var exp = Expression.Lambda(Expression.Convert(expression2.Body, typeof(object)), expression2.Parameters) as Expression<Func<ResolutionContext, object>>;
```
lbargaoanu: As should be casts I believe. The variable names don't say much.

src/AutoMapper/PropertyMap.cs
```diff
@@ -18,18 +24,22 @@ public class PropertyMap
         private IValueResolver _customMemberResolver;
         private bool _sealed;
         private Func<object, object, ResolutionContext, bool> _condition = (_, __, ___) => true;
+        private bool _hasCondition;
```
lbargaoanu: Now with these bools I don't know if the NullObject pattern makes much sense.

src/AutoMapper/PropertyMap.cs
```diff
@@ -259,5 +322,60 @@ protected override Expression VisitMember(MemberExpression node)
...
+                    return node;
+                if (node.NodeType != ExpressionType.Lambda && node.NodeType != ExpressionType.Parameter)
+                {
+                    var expression = node;
+                    if (node.Type == typeof(object))
+                        expression = Expression.Convert(node, _overrideExpression.Parameters[0].Type);
+
+
+                    var a = new ReplaceExpressionVisitor(_overrideExpression.Parameters[0], expression);
+                    var b = a.Visit(_overrideExpression.Body);
```
lbargaoanu: a & b ? :)

src/AutoMapper/Execution/NullReferenceExceptionSwallowingResolver.cs
```diff
@@ -5,6 +10,7 @@ namespace AutoMapper
     public class NullReferenceExceptionSwallowingResolver : IMemberResolver
```
lbargaoanu: I guess this needs a new name.

src/AutoMapper/Execution/NullReferenceExceptionSwallowingResolver.cs
```diff
@@ -23,6 +29,20 @@ public object Resolve(object source, ResolutionContext context)
             }
         }
 
+        public LambdaExpression GetExpression => Visitor.Visit(_inner.GetExpression) as LambdaExpression;
```
lbargaoanu: As should be a cast I believe.

src/AutoMapper/Execution/DelegateBasedResolver.cs
```diff
@@ -4,30 +4,29 @@ namespace AutoMapper.Execution
 {
     using System;
 
-    public class DelegateBasedResolver<TSource, TMember> : IMemberResolver
+    public class DelegateBasedResolver<TSource, TMember> : IValueResolver
```
lbargaoanu: This exposed the type so PropertyMap.SourceType would use it. I presume now works differently.

src/AutoMapper/Execution/NullReferenceExceptionSwallowingResolver.cs
```diff
@@ -23,6 +29,20 @@ public object Resolve(object source, ResolutionContext context)
...
+        private class IfNotNullVisitor : ExpressionVisitor
+        {
+            private readonly IList<MemberExpression> AllreadyUpdated = new List<MemberExpression>(); 
+            protected override Expression VisitMember(MemberExpression node)
+            {
+                if (AllreadyUpdated.Contains(node))
+                    return base.VisitMember(node);
+                AllreadyUpdated.Add(node);
+                var a = DelegateFactory.IfNotNullExpression(node);
+                return Visit(a);
```
lbargaoanu: a?

src/AutoMapper/MappingEngine.cs
```diff
@@ -51,5 +49,72 @@ public object Map(ResolutionContext context)
...
+            {
+                throw new InvalidOperationException("Cannot create destination object. " + context);
+            }
+
+            if (context.SourceValue != null && !context.Options.DisableCache)
+                context.InstanceCache[context] = mappedObject;
+
+            context.TypeMap.BeforeMap(context.SourceValue, mappedObject, context);
+            context.BeforeMap(mappedObject);
+
```
lbargaoanu: Here you don't reuse the context anymore.

src/AutoMapper/Execution/ReflectionHelper.cs
TylerCarlson1: They differ on typeof(), but everything else is the same yes.  You could make a function and just pass in the type and save a few lines.  I left it this way because I didn't want to differentiate too much from the original.

I'm not opposed to changing it in the future though.

src/AutoMapper/Execution/ReflectionHelper.cs
TylerCarlson1: This try catch is there for when you try to make type maps for GenericDefinitionType.  Like `IFoo<T>`, it can't make an expression using this type and will throw and exception.  And for that I need it to not be null to stop NullReference exceptions down the execution path, I think for comparison overrides in propertymap.  These are for generic type maps and thus will never be used in actual mappings.  I also ignore generic mapping types in AssertConfiguration is valid, since it's impossible for them to be used, and they were causing exceptions.

I admit this is kind of a hack to get the thing to build and pass, and can probably be implemented better.

src/AutoMapper/PropertyMap.cs
TylerCarlson1: My default names for things when I brute force the solution.

src/AutoMapper/Execution/DelegateBasedResolver.cs
TylerCarlson1: DelegateBasedResolver uses ResolutionContext, and for simple mapping using expression translation, I'm only using functions of TSource and TMember to get and set values.  Maybe in the future it can but for now I'm ignoring translating anything with ResolutionContext.

TylerCarlson1:
@lbargaoanu My PR had some less than readable or understandable bits,  It's still a work in progress so I expect everything to get refactored and renamed so it's more understandable.  I'm just brute forcing the solution to work right now, not knowing what all I have to change to get it to work.

src/AutoMapper - Copy/Execution/PropertyAccessor.cs
jbogard: I think this file is named weird here too.

src/AutoMapper/Execution/PropertyAccessor.cs
jbogard: What, the `Lazy<Expression<LateBoundPropertySet<TSource, TValue>>>` type?

src/AutoMapper/PropertyMap.cs
jbogard: It doesn't and will perform better not pre-allocating I think.

src/AutoMapper/Execution/PropertyAccessor.cs
TylerCarlson1: I think so.  I'm not sure if you need the lazy part anymore.  I tried removing it once before and it caused tests to fail, so this refactoring I left it in.

I'm under the impression if someone has gone this deep into the code base, they can figure it out what this means.

src/AutoMapper - Copy/Execution/PropertyAccessor.cs
TylerCarlson1: PropertySetter?  PropertyGetterSetter?

jbogard:
Also some of those conditions can go away, because I can surround the basic assignment func with all the other stuff too.

jbogard:
Actually, I'm a bit tired of refactoring this one without banking some of these perf improvements.

What's left? Well, I'd like to see the conditional of building up the assignment down to just a type decision. And at that point, it's a question of calling Mapper.Map or just doing the assignment.

The rest of the conditionals are going to be checking PropertyMap config, and altering the mapping function with additional behavior by wrapping Func's (Expressions would just be for the basic assignment, Func's for the rest).

TylerCarlson1:
I'm fine with submitting it now, I would just suggest refactoring some of the names of functions and values to something more appropriate.  Like a and b variables, and possibly having minor naming conventions.  Just make it more organized and understandable for now. @jbogard you could probably refactor it and get the jist of what everything does.

For you what's left I'm not sure what exactly you mean.  Might be the same thing I'm thinking though.
I don't see how you can call just Mapper.Map unless you want to find the matching type pair match and cache it for mapping in the future?

I see it as getting rid of as many if statements in the _mapFunc as possible for each type, but I'm concerned that the functionality used to determine this is spread around the IObjectMappers and thus duplicating code possibly.

src/AutoMapper/Execution/PropertyAccessor.cs
lbargaoanu: Yes, the Lazy one :)

lbargaoanu:
@jbogard How useful is the instance cache on the context? :) Because getting rid of it would get us some nice memory savings, around 20%.

jbogard:
That one I think we can make it opt-in instead of opt-out. You have to
disable the cache explicitly instead of enable it, and it's only there for
circular references. I'd rather see a PreserveReferences or similar on a
mapping configuration.

TylerCarlson1:
Oh on making a lambda expression and compile a func to set the property directly, I don't think I'll have something like that any time soon.  Maybe this weekend if I decide to look at it.

After you decide to take out calling Map for every property that didn't need it, everything else feels like diminishing returns in comparison.  It's going to take a while before we get orders of magnitude performance gain again, I think.

TylerCarlson1:
I see what you are doing in breaking things down.  I'm not in the mindset to refactor those things right now. I'm just worried if it's broken up into so many parts with duplicate pieces it gets edited in one place and not the other and causes stupid mistakes.

Other than that looks good, get rid of as many un-necessary if statements and map calls as humanly possible.

TylerCarlson1:
Also maybe the benchmarks should include these other type of mappings, just so you can see the performance on each kind.  Kind of like Mapster does, with flattening and complex types.  Or just use what they have and just change the wiki. :)

PR #1159, opened 2016-03-04
No More HasSetter

TylerCarlson1:
Get Rid of HasSetter and it's wasteful wasteful type and property checking.

Not much performance gain, but some


TylerCarlson1:
Wow that was fast

PR #1160, opened 2016-03-04
Show performance differences

TylerCarlson1:
The more casting I did the more it got worse.

This is essentially what happens at https://github.com/AutoMapper/AutoMapper/blob/develop/src/AutoMapper/PropertyMap.cs#L113

I have generics with IValueResolver, but I still need to make a whole bunch of tests past first.


jbogard:
I added a manual mapper that emulates how we actually map.

PR #1162, opened 2016-03-05
[WIP] Remove type map registry

TylerCarlson1:
Still have 3 failing tests, 2 of which have to do with open generics types.

Still can optimize more with a complete set action, but for now it can get and set.

### DIFF
--- a/src/AutoMapper.Net4/AutoMapper.Net4.csproj
+++ b/src/AutoMapper.Net4/AutoMapper.Net4.csproj
@@ -49,8 +49,440 @@
     <Reference Include="System.Core, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089, processorArchitecture=MSIL" />
   </ItemGroup>
   <ItemGroup>
-    <Compile Include="..\AutoMapper\**\*.cs">
-      <Link>%(RecursiveDir)%(Filename)%(Extension)</Link>
+    <Compile Include="..\AutoMapper\AssemblyInfo.cs">
+      <Link>AssemblyInfo.cs</Link>
+    </Compile>
+    <Compile Include="..\AutoMapper\AutoMapperConfigurationException.cs">
+      <Link>AutoMapperConfigurationException.cs</Link>
+    </Compile>
+    <Compile Include="..\AutoMapper\AutoMapperMappingException.cs">
+      <Link>AutoMapperMappingException.cs</Link>
+    </Compile>
+    <Compile Include="..\AutoMapper\ConfigurationValidator.cs">
+      <Link>ConfigurationValidator.cs</Link>
+    </Compile>
+    <Compile Include="..\AutoMapper\Configuration\Conventions\AllMemberInfo.cs">
+      <Link>Configuration\Conventions\AllMemberInfo.cs</Link>
+    </Compile>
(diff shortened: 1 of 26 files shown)


TylerCarlson1:
Seeing about 10-12% performance increase.  I could be able to convert the resolved function from using resolution context to use the mapping object directly.  Also remove the casts in the set and the only cast will be from source to the expected action type.

TylerCarlson1:
Ignored Validating GenericTypeDefinition Type maps for configuration validation.  Can't map `Type<>` anyways.

Oh I might of missed one map that's failing but ignored another one.  I think the only thing the expressions can't do is deal with creating a new object if the final result is null

src/AutoMapper.Net4/AutoMapper.Net4.csproj
```diff
@@ -49,8 +49,446 @@
     <Reference Include="System.Core, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089, processorArchitecture=MSIL" />
   </ItemGroup>
   <ItemGroup>
-    <Compile Include="..\AutoMapper\**\*.cs">
-      <Link>%(RecursiveDir)%(Filename)%(Extension)</Link>
```
jbogard: Uhhhhhhh whoops :)

TylerCarlson1: Yea it always does that.  You know this.  I thought since it's a PR of a PR it will get fixed after merge to development.  Didn't even bother with this.

TylerCarlson1:
Ok got all the tests to pass now.  Looks like spaghetti code now in PropertyMap.cs
Definitely needs serious tidying up.

At least it works.  Would be nice if AppVoyer trigger built all PR's on update instead of just PRs for development. :)

jbogard:
So I accidentally merged this one :(

I think the condition was a little eager in filtering out assignable types. Here's the before/after from benchmarking:

image

If you run in dottrace you'll see that all the string mapping goes through the worst case func.

jbogard:
Is there any reason to go through the current propertygetter etc classes for expression creation at all? Really all you need are the MemberInfo objects to build out expression trees. I think we can skip all the LateBound etc etc delegates completely and just build expressions based on metadata directly.

TylerCarlson1:
2 Things.
First the .Net trace if you looked more showed string -> string wasn't using resolved expression.  That was because string is IEnumerable and if you replace the line with `(!SourceType.IsEnumerableType() || SourceType == typeof(string))` it's fine.  It has nothing to do with the expression being slower than the value resolvers.

Second, the reason I went through them was because I was trying to get it to work.  I originally had them using delegate base, but had issues and you made this PR so I decided to start over fresh.  Also with Delegate based resolvers, they use expressions instead of func, so I can then use them as IMemberResolvers, so they too will also get the benefit of expression resolving.  Even Resolve using will do this, if you use the version that doesn't have Resolution Context as a parameter.

TylerCarlson1:
Oh but changing that code for strings to work makes some of the tests fail.  I guess I gotta go back and fix that.  If you want to try Reverting your Revert and add this line you can see the performance difference.

TylerCarlson1:
Ok I got a fix for the strings where you get the increased performance gain from using Aggregate resolver.  Did only 1 test but got a 6% performance gain.

So Rebase and new PR?

PR #1164, opened 2016-03-05
Remove type map registry

TylerCarlson1:
This should have a proper build and passing tests.


TylerCarlson1:
I'm trying to see how to set values in structures, but it looks like it's not so easy to make an expression to set a value in structures, considering it makes a new object when you compile the expression, meaning the property doesn't get set.

I'm still looking into this, but another alternative is to make it work, but either you leave it as is for now, or you have to ignore struct mapping to get the speed boosts.

TylerCarlson1:
Ok looks like I figured out how to make ref work with struct and set the values through an expression.

Now comes the "Fun fun FUUUN" part of merging the two LambdaExpressions into a single LambdaExpression using ExpressionVisitor.  Struct and Class use 2 different Lambda Types so I probably need to pass in the get lambda in and make the expression at the Accessor level to account for this.

ExpressionVisitor is very very powerful, but it just sucks working with the abstraction of dynamically generating code.  And I got to do it twice more :(.

lbargaoanu:
With reflection it's done through [TypedReference](http://stackoverflow.com/questions/1711393/practical-uses-of-typedreference).

TylerCarlson1:
I might hold off on making the reflection for setting for now, since there's extra use cases when doing the mapping, for null substitution.

I want to wait for the way property mapping is finalized before trying to work on optimizing it more.  If are able to pass in profile or configuration to handle null values and can remove the if statement, then that means I have to re-work my solution.  I'll wait for @jbogard for now, unless something else comes up.

jbogard:
Should be able to pass in whatever you need, heck I'm not sure why TypeMap
doesn't have a ref to its profile.

TylerCarlson1:
You want to merge and pass the type map so that you don't have to call the if statement?  I can make another PR to refine it more after that.

Or I can add this myself.  I would prefer you verify my current PR is ok before proceeding to more refactoring.  Don't know how far you want to take this, but the performance is more than significant and everything passes right now

Sent from my iPhone

On Mar 6, 2016, at 12:37 PM, Jimmy Bogard <notifications@github.com<mailto:notifications@github.com>> wrote:

Should be able to pass in whatever you need, heck I'm not sure why TypeMap
doesn't have a ref to its profile.

On Sunday, March 6, 2016, Tyler Carlson <notifications@github.com<mailto:notifications@github.com>> wrote:

> I might hold off on making the reflection for setting for now, since
> there's extra use cases when doing the mapping, for null substitution.
>
> I want to wait for the way property mapping is finalized before trying to
> work on optimizing it more. If are able to pass in profile or configuration
> to handle null values and can remove the if statement, then that means I
> have to re-work my solution. I'll wait for @jbogard
> <https://github.com/jbogard> for now, unless something else comes up.
>
> -
> Reply to this email directly or view it on GitHub
> <https://github.com/AutoMapper/AutoMapper/pull/1164#issuecomment-192818335>
> .
>

-

PR #1172, opened 2016-03-08
Disable the cache by default

lbargaoanu:
Breaking change...

### DIFF
--- a/src/AutoMapper/IMappingOperationOptions.cs
+++ b/src/AutoMapper/IMappingOperationOptions.cs
@@ -22,7 +22,7 @@
         /// <summary>
         /// Disable the cache used to re-use destination instances based on equality
         /// </summary>
-        bool PreserveReferences { get; set; }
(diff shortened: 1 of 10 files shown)


src/AutoMapper/IMappingOperationOptions.cs
```diff
@@ -22,7 +22,7 @@ public interface IMappingOperationOptions
         /// <summary>
         /// Disable the cache used to re-use destination instances based on equality
         /// </summary>
-        bool DisableCache { get; set; }
+        bool PreserveReferences { get; set; }
```
jbogard: Let's move this up to the TypeMap config level. Disabling is per operation but enabling is per map.

lbargaoanu:
MaxDepth enables the cache.

jbogard:
I think I broke your merge :(

lbargaoanu:
Rebased.

PR #1174, opened 2016-03-09
[WIP] Moving map method to TypeMap

jbogard:


jbogard:
@lbargaoanu I removed the DontPreserveReferences in this one. Since it's opt-in I can't see why you would opt-out for an individual Map call.

jbogard:
@TylerCarlson1 It's slightly ugly with all the Funcs, but if we're ever able to get this to a single compiled expression, having these smaller Funcs makes it easier to convert to an Expression. In theory.

TylerCarlson1:
@jbogard Yea on expression translation I'll have to see how that looks after things have settled down.  I'm thinking now that making the map methods call as little stuff as possible is going to provide by far the biggest boost in performance gain out of anything hands down.  I don't really have any intentions right now of trying to reduce function calls in a map, especially when you and @lbargaoanu are changing things left and right.

Maybe expression visiting will only really benefit by removing boxing, and un-boxing performance cost.  Maybe I should wait until the performance gain can become an order of magnitude :)

jbogard:
I'm also waiting for the work the StructureMap folks are doing to use
Roslyn, that might be even easier than expression tree junk. Expression
tree compilation is pretty expensive.

TylerCarlson1:
IDK, It might have a slightly more versatile ExpressionVisitor.  I wouldn't care too much about expression tree compilation compared to mapping time, cause you do it once at startup.

jbogard:
Weeeeell I've had reports of 30 sec startup time when everything was
compiled up front instead of with `Lazy<T>`

TylerCarlson1:
So we can get rid of Lazy now?  Since it's being compiled anyways on seal?

jbogard:
What I might have to do is wrap the whole thing in a Lazy is what I meant,
but I have to profile on a real world app.

TylerCarlson1:
With normal maps you only have to compile once, vs every flattening would have to compile for each flattening.

For it being ugly I might be able to make it look nice and be understandable by making a dictionary of conditions and actions and run through the list.

src/AutoMapper/TypeMap.cs
```diff
@@ -459,5 +457,163 @@ internal LambdaExpression DestinationConstructorExpression(Expression instancePa
...
+            var destinationFunc = CreateDestinationFunc();
+
+            var assignmentFunc = CreateAssignmentFunc();
+
+            var mapperFunc = CreateMapperFunc(assignmentFunc);
+
+            return (source, context) => mapperFunc(source, context, destinationFunc);
+        }
+
+        private Func<object, ResolutionContext, Func<ResolutionContext, object>, object> CreateMapperFunc(Func<object, ResolutionContext, object, object> assignmentFunc)
```
lbargaoanu: using MapFunc = Func<ResolutionContext, object> would help.

PR #1178, opened 2016-03-10
Refactor type map based constructions

jbogard:


jbogard:
@TylerCarlson1 @lbargaoanu did the same treatment on the constructing of objects. There's probably now dead/redundant code because of these optimizations in places where objects were created.

TylerCarlson1:
Yea probably.  Well that can be resolved at a later PR, along with all the other redundancies of course :)

PR #1179, opened 2016-03-10
Unnecessary allocations in ConstructorMap

lbargaoanu:
[Before](https://cloud.githubusercontent.com/assets/4517428/13670203/31a2a01e-e6d1-11e5-99a5-591222ae1ee7.png) & [after](https://cloud.githubusercontent.com/assets/4517428/13670202/3192685c-e6d1-11e5-92cb-008d230079b6.png).

### DIFF
--- a/src/AutoMapper/ConstructorMap.cs
+++ b/src/AutoMapper/ConstructorMap.cs
@@ -68,30 +68,33 @@ namespace AutoMapper
 
         public object ResolveValue(ResolutionContext context)
         {
-            var ctorArgs = new List<object>();
-
-            foreach (var map in CtorParams)
+            var ctorArgs = new object[CtorParams.Length];
+            for(int index = 0; index < CtorParams.Length; index++)
             {
-                var result = map.ResolveValue(context);
-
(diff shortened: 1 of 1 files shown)


lbargaoanu:
@jbogard Unrelated, I profiled tests with collections. Two possible improvements.
When mapping a collection of objects that contain a collection and the outer collection is big, the trace looks bad because the reflection to get the enumerable element type happens inside a loop.
When mapping any collection, the source is copied. That can be bad is the collection is big enough. It doesn't have to be copied if it's an ICollection, a list or an array for example.
I was not sure whether these cases are relevant enough.

lbargaoanu:
[Reflection](https://github.com/AutoMapper/AutoMapper/blob/099ad42defac98eae52116d0401d698bb83c308a/src/AutoMapper/Mappers/EnumerableMapperBase.cs#L24) and [copying](https://github.com/AutoMapper/AutoMapper/blob/099ad42defac98eae52116d0401d698bb83c308a/src/AutoMapper/Mappers/EnumerableMapperBase.cs#L21).

PR #1180, opened 2016-03-11
Unnecessary allocations in EnumerableMapper

lbargaoanu:
[Before](https://cloud.githubusercontent.com/assets/4517428/13701838/1e96f930-e793-11e5-81e0-c2d8c0df02df.png) & [after](https://cloud.githubusercontent.com/assets/4517428/13701837/1e8536d2-e793-11e5-9904-0ae5b639ed95.png). One idea is to have only one cache, but only ICollections need the EnumerableMapper and arrays don't need the element type cache.

cache the element type of the enumerable
cache the enumerable mapper object for the destination type
don't copy the source if it's ICollection

### DIFF
--- a/src/AutoMapper/Mappers/CollectionMapper.cs
+++ b/src/AutoMapper/Mappers/CollectionMapper.cs
@@ -1,24 +1,15 @@
+using System;
+using System.Collections.Generic;
+using AutoMapper.Configuration;
+
 namespace AutoMapper.Mappers
 {
-    using System;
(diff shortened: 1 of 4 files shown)


TylerCarlson1:
@lbargaoanu I think with all the refactoring to expressions.  This might be obsolete.  Or at least needs to be heavily refactored

lbargaoanu:
Definitely stale :)

Issue #1181, opened 2016-03-12
ProjectTo throws an exception when an interface is mapped with a base-interface property mapping


abrobston:
If an interface is used as the source type in a mapping, and if a property from an inherited interface is specified using `MapFrom`, the `ProjectTo` method causes an exception.  [This question](http://stackoverflow.com/q/2132791/239394) explains the problem, namely that `System.Type.GetMember()` does not return members from inherited interfaces regardless of any `BindingFlags` specified (or their equivalent in portable libraries).

Pull request with a unit test and fix is forthcoming.

PR #1182, opened 2016-03-12
Query inherited interfaces to find member

abrobston:
Addresses #1181.  When the source type of a mapping is an interface, and a property from a base interface is referenced in `MapFrom`, the `ProjectTo` method fails with an exception.  The `MemberResolverExpressionResultConverter.VisitMember` method presumed that `Type.GetMember(string)` always returns at least one `MemberInfo` instance.  However, that method does not return members from inherited interfaces, so those interfaces need to be queried separately to find the desired `MemberInfo` instance.


lbargaoanu:
I think it's more readable as a [LINQ query](https://gist.github.com/lbargaoanu/a2deff11b8040d65b81b).

jbogard:
I think I'll merge #1186 since it closes 2 issues, that OK?

jbogard:
Fixed by #1186

PR #1189, opened 2016-03-17
Change ReplaceName's PossibleNames() to merge enumerables with Union …

rossover:
Change ReplaceName's PossibleNames() to merge enumerables with Union instead of Concat to avoid duplicates.


rossover:
Is there anything else I need to do on this?

Thanks!

Ross

jbogard:
Conflicts :(

PR #1190, opened 2016-03-17
CreateMap benchmark

rossover:
Added benchmark for CreateMap and replaced IObjectToObjectMapper with a less-qualified interface (IBenchmarker).


jbogard:
Conflicts :(

PR #1201, opened 2016-03-24
[Wip] Constructor performance

TylerCarlson1:
Generate Constructor expression instead of calling a function that calls another function that calls a lambda that creates object

This performance should be similar to compiling expressions for mapping vs making calls that make other calls inside other calls.

### DIFF
--- a/src/AutoMapper/Execution/DelegateFactory.cs
+++ b/src/AutoMapper/Execution/DelegateFactory.cs
@@ -10,10 +10,10 @@ namespace AutoMapper.Execution
 
     public class DelegateFactory
     {
-        private readonly ConcurrentDictionary<Type, LateBoundCtor> _ctorCache =
-            new ConcurrentDictionary<Type, LateBoundCtor>();
+        private readonly ConcurrentDictionary<Type, Expression> _ctorCache =
+            new ConcurrentDictionary<Type, Expression>();
 
-        private readonly Func<Type, LateBoundCtor> _generateConstructor;
+        private readonly Func<Type, Expression> _generateConstructor;
 
         public DelegateFactory()
         {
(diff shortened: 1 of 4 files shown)


TylerCarlson1:
hmm maybe I'm missing something for the performance gain part.  I was also trying to add generics to type and property map to reduce casting, hence all the commits.  Maybe that had something to do with it.

src/AutoMapper/Execution/DelegateFactory.cs
```diff
@@ -145,10 +144,8 @@ private static LateBoundCtor GenerateConstructor(Type type)
                     .Select(p => Expression.Constant(p.DefaultValue, p.ParameterType)).ToArray();
 
                 //create the ctor expression
-                var ctorExpression =
-                    Expression.Lambda<LateBoundCtor>(Expression.Convert(Expression.New(ctorWithOptionalArgs, args),
```
TylerCarlson1: No reason to cast to object because allready is one.  Very slight performance gain

src/AutoMapper/Execution/DelegateFactory.cs
```diff
@@ -112,21 +112,20 @@ public static Expression IfNotNullExpression(MemberExpression member)
...
         }
 
-        private static LateBoundCtor GenerateConstructor(Type type)
+        private static Expression GenerateConstructor(Type type)
         {
             //handle valuetypes
             if(!type.IsClass())
             {
-                var ctorExpression =
-                    Expression.Lambda<LateBoundCtor>(Expression.Convert(Expression.New(type), typeof(object)));
```
TylerCarlson1: Needs the convert to object because it's a struct.  If used generics then convert wouldn't have to be called.

TylerCarlson1:
Looks like might not have much gain here.  But Constructor map can get a lot of improvement if you generate a func to cast the source type and use the properties to map like generating a user defined construct using.  Would give significant speedup on constructors

jbogard:
Should I close this one unmerged?

TylerCarlson1:
I have to take another look at this.  You already submitted something that should do something similar and I think this PR is incompatible with that.

TylerCarlson1:
Yea this isn't compatible with my other PR.  It will be best if I redo it.  Now I'm able to remove un-necessary casts to object and then back to type when caling the constructor with the other PR, so I'll do that.

jbogard:
I think we can get rid of Seal on PropertyMap and ConstructorMap and have
those just build an expression and return it. Doesn't need to be a Lambda
expression either, just an Expression at this point.

jbogard:
I was starting on that this evening, having those given the passed in parameter expressions.

TylerCarlson1:
Working on that now as well.  Trying to get proxy type generation to work, then I can move on to eliminating object casting.

PR #1204, opened 2016-03-25
Use the runtime type of the source object when mapping collections

### LINKED ISSUE #1203: Dynamic mapping in 4.2.1 not behaving as expected
I'm not sure if I'm doing this wrong but here is a contrived example of a mapping from an ExpandoObject to an Author type (requires JSON.NET library to run):

```C#
    class Program
    {
        {
            var config = new MapperConfiguration(cfg => cfg.CreateMissingTypeMaps = true);

            Author a = new Author()
            {
                Name = "Charles Dickens",
                Books = new[]
                {
                    new Book()
                    {
                        Name = "Great Expectations",
                    },
                    new Book()
                    {
                        Name = "Oliver Twist"
                    }
                }
            };

            var jsonString = JsonConvert.SerializeObject(a);
            dynamic authorDynamic = JsonConvert.DeserializeObject<ExpandoObject>(jsonString, new ExpandoObjectConverter());
            Author mappedAuthor = mapper.Map<Author>(authorDynamic);
        }
    }

    public class Author
    {
        public Book[] Books { get; set; }
    }

    public class Book
    {
    }
```

The result of this is that `mappedAuthor` has the `Name` property set correctly. The `Books` array has 2 elements but for each book in there, the `Name` property is null. The mapping will work if I change my config to this:

```C#
            {
                cfg.CreateMap<ExpandoObject, Book>()
                    .ForMember(
                        dest => dest.Name,
                        opts => opts.MapFrom(src => ((IDictionary<string, object>) src)["Name"])
                    );
            });
```

This feature is new so I'm not sure if this is a bug or if my configuration is wrong.

lbargaoanu:
fixes #1203

### DIFF
--- a/src/AutoMapper/Mappers/EnumerableMapperBase.cs
+++ b/src/AutoMapper/Mappers/EnumerableMapperBase.cs
@@ -41,8 +41,10 @@ namespace AutoMapper.Mappers
             var itemContext = new ResolutionContext(context);
             foreach (object item in enumerableValue)
             {
-                var typeMap = context.ConfigurationProvider.ResolveTypeMap(item, null, sourceElementType, destElementType);
-                itemContext.Fill(item, null, sourceElementType, destElementType, typeMap);
+                var sourceItemType = item?.GetType() ?? sourceElementType;
+                var typeMap = context.ConfigurationProvider.ResolveTypeMap(sourceItemType, destElementType);
+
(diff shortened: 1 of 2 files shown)


TylerCarlson1:
There will be a slow down from this having to get type maps from each item in a collection.

lbargaoanu:
Just the type changed, nothing to see here :)

PR #1211, opened 2016-03-30
[WIP] Property map expressions

jbogard:
@TylerCarlson1 this is what I was mentioning, my next step is to get all the lambdas converted to expressions in all those functions being built up


TylerCarlson1:
Still looking through all the changes, but it looks like what I wanted to do.  A lot of cleanup too :)

So what you want to do is change _mapperFunc in PropertyMap to be a compiled expression? instead of a bunch of lambdas wrapping around each other.  If so that's what my end goal was as well.

Also with compiling expressions for mapping, will you be re-using or re-purposing the expression generation code for ProjectTo in any way?  I mean the non resolve and ResolutionContext stuff should be identical between the ProjectTo and the Map calls execution wise.

jbogard:
One step further, actually. Have PropertyMap build out an expression, then the TypeMap fold all PropertyMap expressions into one compiled expression, including removing all that casting that has to happen.

On the ProjectTo stuff, the problem is a lot of that is based on what is supported for LINQ. With pure expression trees things can get way more complex and compiled down.

TylerCarlson1:
Great that's what I wanted in the end as well.  To have a compiled single execution of the map.  And possibly have that expression in the debug menu or something, showing code generated by the actual map does.

So what happens to resolution context at this point?  You going to set that up in the generated expression?  Or try to reverse engineer the point of origin to get the typemap/propertymap involved?  Or just ignore it and have a different form of exception message?

TylerCarlson1:
Are you going to remove the casting to objects at the type map levels or are you going to make the property maps not have the casting?

As far as remove object castings for setting properties I would suggest just using pure expressions and not even worry about making the types match until converting to lambda at the very end.
Also with expression generation you might as well have a static object and resolution context for now that both TypeMap and PropertyMap can use.  That way they are all using the same inputs and you don't have to compile the lambda statement at the propertymap level, cause you can just add it to a block expression.
I could probably make a PR for this even.

Also for casting is that at the resolved object level or just setting the properties level?  Cause if it's for resolving the object I would suggest Generics so you can pass an `Action<T, ResolutionContext>`

src/AutoMapper/ConstructorMap.cs
```diff
@@ -1,81 +1,83 @@
...
+            _runtimeCtor = new Lazy<LateBoundParamsCtor>(() => DelegateFactory.CreateCtor(ctor, CtorParams));
+        }
+
+        private static readonly IExpressionResultConverter[] ExpressionResultConverters =
+        {
+            new MemberGetterExpressionResultConverter(),
+            new MemberResolverExpressionResultConverter(),
+        };
+
+        public bool CanResolve => CtorParams.All(param => param.CanResolve);
```
lbargaoanu: This will allocate stuff.

src/AutoMapper/Mappers/ObjectCreator.cs
```diff
@@ -39,6 +40,17 @@ public static object CreateDictionary(Type dictionaryType, Type keyType, Type va
             return CreateObject(type);
         }
 
+        public static object CreateDictionary(Type dictionaryType)
+        {
+            Type keyType = dictionaryType.GetTypeInfo().GenericTypeArguments[0];
+            Type valueType = dictionaryType.GetTypeInfo().GenericTypeArguments[1];
```
lbargaoanu: This could be cleaner.

src/AutoMapper/Mappers/ObjectCreator.cs
```diff
@@ -59,7 +71,9 @@ public static object CreateObject(Type type)
                 ? CreateArray(type.GetElementType(), 0)
                 : type == typeof (string)
                     ? null
-                    : DelegateFactory.CreateCtor(type)();
+                    : type.IsDictionaryType()
```
lbargaoanu: Maybe IsInterface here?

src/AutoMapper/PropertyMap.cs
```diff
@@ -68,39 +83,53 @@ public MemberInfo SourceMember
...
+            if (NullSubstitute == null)
+            {
+                SetNullSubstitute(inheritedMappedProperty.NullSubstitute);
+            }
+            if (_mappingOrder == 0)
+            {
+                SetMappingOrder(inheritedMappedProperty._mappingOrder);
+            }
+            SourceType = SourceType ?? inheritedMappedProperty.SourceType;
+            CustomValue = CustomValue ?? inheritedMappedProperty.CustomValue;
```
lbargaoanu: This is weird :)

lbargaoanu:
Pretty hard core. If speed is the purpose, it'd better be fast :)

src/AutoMapper/ConstructorMap.cs
jbogard: Which will? The closure?

src/AutoMapper/Mappers/ObjectCreator.cs
jbogard: Yeah, I think in the long run I'll use something that creates an Expression, not a delegate.

src/AutoMapper/PropertyMap.cs
jbogard: The idea is I don't want to overwrite something already set. This would be easier if the inherited property map got applied BEFORE the property map stuff. Which I could probably do eventually.

src/AutoMapper/ConstructorMap.cs
lbargaoanu: I believe the foreach based on IEnumerable. Reverting should be ok.

jbogard:
@lbargaoanu So the goal for me, eventually, is to have TypeMap build an entire Expression at Seal. Then a `Lazy<Func>` that would compile at first map time. Right now just trying to incrementally get to expressions. I'm also thinking that all the code building expressions would get moved out (like IObjectMapper). Or folded in to IObjectMapper, not quite sure yet. There are some repeated conditionals based on types between the expression building up and IObjectMappers (assignable, strings etc).

lbargaoanu:
I guess it's a process :) Hopefully it will look much better when it's done.

src/AutoMapper/ConstructorMap.cs
lbargaoanu: Yes, it's [enumerators](https://cloud.githubusercontent.com/assets/4517428/13640021/ec66519a-e61b-11e5-9053-c230f928715d.png).

src/AutoMapper/PropertyMap.cs
```diff
@@ -195,117 +199,137 @@ internal void Seal(TypeMapRegistry typeMapRegistry)
...
+                Expression sourceFunc;
                 if (ValueResolverConfig.SourceMember != null)
                 {
-                    var newParam = Expression.Parameter(typeof (object), "m");
-                    var expr = new ConvertingVisitor(ValueResolverConfig.SourceMember.Parameters[0], newParam).Visit(ValueResolverConfig.SourceMember.Body);
-                    var lambda = Expression.Lambda<Func<object, object>>(Expression.Convert(expr, typeof (object)), newParam);
-                    Func<object, object> mapFunc = lambda.Compile();
-
-                    sourceFunc = ctxt => mapFunc(ctxt.SourceValue);
+                    sourceFunc = new ConvertingVisitor(ValueResolverConfig.SourceMember.Parameters[0], srcParam).Visit(ValueResolverConfig.SourceMember.Body);
```
TylerCarlson1: I would suggest making an extension method for this.  It's probably going to be used quite a bit with all the expression translations
```
public static Expression ReplaceParameters(this LamdbaExpression lambda, param Expression[] replacements)
{
    Expression convertedExpression = lambda.Body;
    for(var i = 0;i < replacments.Length;i++)
        convertedExpression = new ConvertingVisitor(lambda.Parameters[i], replacements[i]).Visit(convertedExpression);
    return convertedExpression;
}
```

src/AutoMapper/PropertyMap.cs
```diff
@@ -225,33 +199,145 @@ internal void Seal(TypeMapRegistry typeMapRegistry)
...
+                }
 
-        public void ChainTypeMemberForResolver(IValueResolver valueResolver)
-        {
-            ResetSourceMemberChain();
-            _customMemberResolver = valueResolver;
+                Expression sourceFunc;
+                if (ValueResolverConfig.SourceMember != null)
+                {
+                    sourceFunc = new ConvertingVisitor(ValueResolverConfig.SourceMember.Parameters[0], srcParam).Visit(ValueResolverConfig.SourceMember.Body);
```
TylerCarlson1: https://github.com/AutoMapper/AutoMapper/pull/1211#discussion_r58068857 use extension method here

src/AutoMapper/PropertyMap.cs
```diff
@@ -225,33 +199,145 @@ internal void Seal(TypeMapRegistry typeMapRegistry)
...
+            }
+            else if (SourceMemberName != null)
+            {
+                valueResolverFunc = Expression.MakeMemberAccess(
+                        Expression.Convert(srcParam, _typeMap.SourceType),
+                        _typeMap.SourceType.GetFieldOrProperty(SourceMemberName));
+            }
+            else if (CustomExpression != null)
+            {
+                var expr = new ConvertingVisitor(CustomExpression.Parameters[0], srcParam).Visit(CustomExpression.Body);
```
TylerCarlson1: https://github.com/AutoMapper/AutoMapper/pull/1211#discussion_r58068857 use extension method here

src/AutoMapper/PropertyMap.cs
```diff
@@ -371,60 +471,98 @@ protected override Expression VisitMember(MemberExpression node)
...
+                return node;
+            if (node.NodeType != ExpressionType.Lambda && node.NodeType != ExpressionType.Parameter)
             {
-                return _oldExpression == node ? _newExpression : base.Visit(node);
+                var expression = node;
+                if (node.Type == typeof(object))
+                    expression = Expression.Convert(node, _overrideExpression.Parameters[0].Type);
+
+
+                var a = new ReplaceExpressionVisitor(_overrideExpression.Parameters[0], expression);
```
TylerCarlson1: https://github.com/AutoMapper/AutoMapper/pull/1211#discussion_r58068857 use extension method here

TylerCarlson1:
Worked on it a little myself and got basically the same thing as Expressions test passing commit + setting expression.  Didn't have the lamdbas but it would be good to pass the lambdas back to replace parameters with type mapped resolved objects at the type map level.

TylerCarlson1:
Possibly help with Expression woes

Tested, except IfNull which looks like it should work as intended
```
internal static class ExpressionVisitorExtensions
    {
        public static Expression ReplaceParameters(this LambdaExpression exp, params Expression[] replace)
        {
            Expression replaceExp = exp;
            for (var i = 0; i < replace.Count(); i++)
                replaceExp = new ReplaceExpressionVisitor(exp.Parameters[i], replace[i]).Visit(exp.Body);
            return replaceExp;
        }

        public static Expression Replace(this Expression exp, Expression old, Expression replace)
        {
            return new ReplaceExpressionVisitor(old, replace).Visit(exp);
        }

        public static LambdaExpression Concat(this LambdaExpression expr, LambdaExpression concat)
        {
            return (LambdaExpression) new ExpressionConcatVisitor(expr).Visit(concat);
        }

        public static Expression IfNull(this Expression expression, params Expression[] ifElse)
        {
            return ifElse.Any()
                ? IfThenElse(NotEqual(expression, Constant(null)), expression, ifElse.First().IfNull(ifElse.Skip(1).ToArray()))
                : expression;
        }
    }
```

Examples for IfNull (Please name better)
Others are pretty self explanatory
```
// For using Map
var sourceType = SourceExpression.IfNull(Call(SourceExpression, typeof(object).GetMethod("GetType")), Constant(SourceType), Property(ResolutionContextExpression, "SourceType"));
```
```
else if (!_typeMap.Profile.AllowNullDestinationValues)
  valueResolverFunc = valueResolverFunc.IfNull(Call(CreateNonNullValue, Constant(SourceType).IfNull(Constant(DestinationPropertyType))))
```

src/AutoMapper/PropertyMap.cs
```diff
@@ -219,7 +219,24 @@ internal void Seal(TypeMapRegistry typeMapRegistry)
...
+            Expression mapperExpr;
+            if (DestinationProperty.MemberInfo is FieldInfo)
+            {
+                mapperExpr = Expression.Assign(destMember, Expression.Convert(valueResolverExpr, DestinationPropertyType));
+            }
+            else
+            {
+                var prop = (PropertyInfo) DestinationProperty.MemberInfo;
+                var setter = prop.GetSetMethod(true);
+                if (setter == null)
```
TylerCarlson1: Stupid question, but don't you filter out properties by if they have a setter somewhere else in code?

jbogard: Yeah, the IMemberAccessors have that logic....kinda. They also pre-calculate a lot of stuff, which I don't want. Eventually going to push that stuff back down but right now it breaks stuff.

jbogard:
Some numbers - with this change, the 5.0 line is now 10x faster in my benchmarks than 4.x and 3.x

jbogard:
The next thing I'm tackling is the TypeMap one. Converting all of that to expressions.

And probably the last-last thing, moving all the expression building out to something else. It takes a TypeMap, and builds an expression on how to map it, moving all the stuff out of PropertyMap too. Then TypeMap and PropertyMap are just about the configuration of how you've told AutoMapper to map things together, but the execution plan is moved out. Whew.

TylerCarlson1:
Guess I can move out Expression Generation Code into it's own class and call funcs on property map.  Along with making it more readable.

PR #1218, opened 2016-04-02
Ctors based on expressions

jbogard:
Ugly but it works (tm)

### DIFF
--- a/src/AutoMapper/ConstructorMap.cs
+++ b/src/AutoMapper/ConstructorMap.cs
@@ -8,24 +8,21 @@ namespace AutoMapper
     using System.Linq;
     using System.Linq.Expressions;
     using System.Reflection;
-    using Configuration;
-    using Execution;
-    using Mappers;
 
     public class ConstructorMap
     {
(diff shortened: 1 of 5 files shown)


jbogard:
Some inefficiencies here, it still tries all the properties even though they're mapped with a constructor parameter map.

I'm not sure how to resolve that since I can't know for sure if a constructor parameter is equivalent to a property.

@TylerCarlson1 @lbargaoanu yes it's ugly, I know.

TylerCarlson1:
Yes it is.  That will take a good while before it's looking good.  Real question is any performance gain on Ctors mappings?

jbogard:
Yeah they're twice as fast now (compare the AppVeyor numbers with the
previous build).

jbogard:
More than that, .92 sec to .28. Not too shabby.

TylerCarlson1:
I always compare the ratio between the AutoMapper vs hard coded ratios.  Because the server could be slower for whatever reason, but the comparison between optimal the speed should be relatively the same during that time.

PR #1219, opened 2016-04-02
[WIP] Remove PropertyMap Casting

TylerCarlson1:
Moves the Expression creation to another class, and then replaces parameters at the PropertyMap level.
Made some extension methods as well to help things be more readable.

Property map will convert parameters to Source and Destination type, but the expression generated will assume the types are already set correctly.  What's good about this that you can convert the same expression over for the TypeMap.  You can just replace the parameters with variables you set when generating the TypeMap expression, and not have to re-write the PropertyMap Expression generating code.

Only real problems having right now are with struct's, so what probably needs to be done is convert the expressions to use ref if ValueType, so can pass object over to be set correctly, or something like that.

### DIFF
--- a/src/AutoMapper/PropertyMap.cs
+++ b/src/AutoMapper/PropertyMap.cs
@@ -1,6 +1,7 @@
 using AutoMapper.Configuration;
 using AutoMapper.Mappers;
 using AutoMapper.QueryableExtensions.Impl;
+using static System.Linq.Expressions.Expression;
 
 namespace AutoMapper
 {
@@ -146,14 +147,14 @@ namespace AutoMapper
                 return;
             }
 
-            var srcParam = Expression.Parameter(typeof(object), "src");
-            var destParam = Expression.Parameter(typeof(object), "dest");
-            var ctxtParam = Expression.Parameter(typeof(ResolutionContext), "ctxt");
+            var srcParam = Parameter(typeof(object), "src");
+            var destParam = Parameter(typeof(object), "dest");
+            var ctxtParam = Parameter(typeof(ResolutionContext), "ctxt");
 
             var valueResolverExpr = BuildValueResolverFunc(typeMapRegistry, srcParam, ctxtParam);
             var innerResolverExpr = valueResolverExpr;
-            var destMember = Expression.MakeMemberAccess(
-                Expression.Convert(destParam, _typeMap.DestinationType),
+            var destMember = MakeMemberAccess(
+                Convert(destParam, _typeMap.DestinationType),
                 DestinationProperty.MemberInfo);
 
             Expression getter;
@@ -161,7 +162,7 @@ namespace AutoMapper
             if (DestinationProperty.MemberInfo is PropertyInfo &&
                 ((PropertyInfo) DestinationProperty.MemberInfo).GetGetMethod(true) == null)
             {
-                getter = Expression.Default(_typeMap.DestinationType);
+                getter = Default(_typeMap.DestinationType);
             }
             else
             {
@@ -170,7 +171,7 @@ namespace AutoMapper
 
             var destValueExpr = UseDestinationValue
                 ? getter
-                : Expression.Constant(null, DestinationPropertyType);
+                : Constant(null, DestinationPropertyType);
 
             if (SourceType == null 
                 || (SourceType.IsEnumerableType() && SourceType != typeof (string)) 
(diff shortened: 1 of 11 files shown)


src/AutoMapper/PropertyMap.cs
```diff
@@ -146,151 +147,15 @@ internal void Seal(TypeMapRegistry typeMapRegistry)
...
-                    Expression.Condition(Expression.Equal(sourceExpr, Expression.Constant(null)), 
-                        ifTrue,
-                        ifFalse),
-                    Expression.Constant(DestinationPropertyType),
-                    ctxtParam
-                    );
-                valueResolverExpr = Expression.Block(typeof(object), new[] {sourceExpr}, first, second);
-            }
+            var lambda = this.CreateExpression(typeMapRegistry);
+            var exp = lambda.ReplaceParameters(source, destination);
```
TylerCarlson1: Replace type parameters with conversions of objects.  Later will replace at the TypeMap level with the source and destination values allready converted once, instead of converting them for each type map.

TylerCarlson1:
PS: @lbargaoanu  I know it's ugly in places as well

src/AutoMapper/PropertyMap.cs
```diff
@@ -655,19 +387,292 @@ public override Expression Visit(Expression node)
...
+            return replaceExp;
+        }
+
+        public static Expression Replace(this Expression exp, Expression old, Expression replace) => new ReplaceExpressionVisitor(old, replace).Visit(exp);
+
+        public static LambdaExpression Concat(this LambdaExpression expr, LambdaExpression concat) => (LambdaExpression)new ExpressionConcatVisitor(expr).Visit(concat);
+
+        public static Expression IfNotNull(this Expression expression) => IfNullVisitor.Visit(expression);
+
+        public static LambdaExpression CreateExpression(this PropertyMap propertyMap, TypeMapRegistry typeMapRegistry)
```
TylerCarlson1: This might want to go into another class, or stay here not sure right now.

jbogard:
So eventually I'd like ConstructorMap and PropertyMap to just have returned
Expressions. If TypeMap passes in the ParameterExpressions during Seal it
should make things easier too.

TylerCarlson1:
FYI I'm starting to work on converting TypeMap to use expression as well.  So you can replace parameters over to the constructor and property maps

TylerCarlson1:
Got TypeMaps generating expressions and ignoring all casting for the property maps.  There's still a bunch of failing tests I have to deal with and not all scenarios are supported.

From quick benchmark testing it looks like it might be 2.5 to 3 times faster.

And obviously at this point the code looks uglier than ever.

TylerCarlson1:
Got every test passing except one.  Conditions aren't stopping property mappings from being called once before.  Possibly to get type.

Other thing not there yet is the try catch wrapper around every property map call, to set the property map for the exception handling.

Other things that could be done is for property maps to cache results of resolving in variables, so don't call the resolver multiple times.  The Func code did it this way, but the expressions are not.

jbogard:
Is your latest pushed? I might be able to get those last couple passing.

TylerCarlson1:
I think I got them forgot `.GetTypeInfo()` and had to use Block expression

TylerCarlson1:
Oh for that condition one I had I just put the resolve using under a try catch and return default(DestType).  I don't know if it's the right way, but it made the test pass.  Maybe you might recall what it was suppose to do.

Last Commit I got the build to pass.  The performance looks different from my machine.

The ratio is like 29x slower than manual where  I get like 13x slower which is a difference of 2 times.

I think I'm done unless you can find something.  All I can think of now is tiding up the code and refactor the expression generation out into a single class.  Could be done in another PR though.

Also try catch should work the same for when property maps fail, but there's no unit tests for it. :(

jbogard:
Yeah, looks good. I figure that after this there'd be another pass through to do the kind of refactoring done in the expression mapper - have matchers based on the source/destination type so that we can build out expression builders for other converters. LINQ doesn't support that many but if you look at the IObjectMapper list the non-collection ones would be a great target.

If it's not type mapped and not assignable, go through the expression mappers like we do for regular mapping.

I don't imagine removing IObjectMappers any time soon or ever really, But the collection ones should be able to be optimized based on configuration.

TylerCarlson1:
So something like making IObjectMappers generating expressions that can be passed back and run instead of calling Map?  Or do you want to remove IObjectMappers with expressions to replace them?

I would think IObjectMapper should make the expression, and some can be LINQ supported while others can not,  maybe two interfaces with 2 different implementations.  That way collection support can be put into one place for ProjectTo and Map, so it's easier to compare to each other.

jbogard:
Yep exactly. Maybe not that exact interface but similar.

Issue #339, opened 2013-05-22
SetMappingOrder / Condition statement not evaluated on Included child objects


robvangeloven:
The SetMappingOrder statement and the Condition statement do not propagate to child objects as the code demonstrates below.
This makes it difficult to use the inheritance structure provided by AutoMapper as the mapping order has to be specified for each field in each child object (where needed of course)

//SetMappingOrder demo
public class Source
    {
        public string ClientID
        {
            get;
            set;
        }

```
    public string Foo
    {
        get;
        set;
    }
}

class Dest
{
    public Dest()
    {
    }

    private string id;
    public string ID
    {
        get
        {
            return id;
        }
        set
        {
            id = string.Concat(ID, value);
        }
    }

    private string foo;
    public string Foo
    {
        get
        {
            return foo;
        }
        set
        {
            foo = value;
            ID = value;
        }
    }
}

class ChildDest : Dest
{
    public string Bar
    {
        get;
        set;
    }
}

class Program
{
    static void Main(string[] args)
    {
        Mapper.CreateMap<Source, Dest>()
            .ForMember(dest => dest.ID, opt =>
                {
                    opt.Condition(src => !string.IsNullOrWhiteSpace(src.ClientID));
                    opt.SetMappingOrder(-1000);
                    opt.MapFrom(src => src.ClientID);
                })
            .Include<Source, ChildDest>();

        Mapper.CreateMap<Source, ChildDest>()
            .ForMember(dest => dest.Bar, opt => opt.Ignore());

        var source = new Source()
        {
            ClientID = "Test",
            Foo = "Bar"
        };

        var parentDest = Mapper.Map<Dest>(source);
        var childDest = Mapper.Map<ChildDest>(source);

        Debug.Assert(parentDest.ID == "TestBar");
        Debug.Assert(childDest.ID == "TestBar");
    }
}
```

//Condition demo
class Source
{
public string ClientID
{
get;
set;
}

```
public string Foo
{
    get;
    set;
}
```

}

class Dest
{
    public Dest()
    {
        ID = "12345";
    }

```
public string ID
{
    get;
    set;
}

public string Foo
{
    get;
    set;
}
```

}

class ChildDest : Dest
{
    public string Bar
    {
        get;
        set;
    }
}

class Program
{
    static void Main(string[] args)
    {
        Mapper.CreateMap<Source, Dest>()
            .ForMember(dest => dest.ID, opt =>
                {
[...]

PR #1230, opened 2016-04-13
Inherit the mapping order

lbargaoanu:
https://github.com/AutoMapper/AutoMapper/issues/339#issuecomment-209358649

### DIFF
--- a/src/AutoMapper/TypeMap.cs
+++ b/src/AutoMapper/TypeMap.cs
@@ -105,7 +105,7 @@ namespace AutoMapper
 
         public PropertyMap[] GetPropertyMaps()
         {
-            return _sealed ? _orderedPropertyMaps : _propertyMaps.Concat(_inheritedMaps).ToArray();
+            return _orderedPropertyMaps ?? _propertyMaps.Concat(_inheritedMaps).ToArray();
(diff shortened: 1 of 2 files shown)


src/AutoMapper/TypeMap.cs
```diff
@@ -105,7 +105,7 @@ public int MaxDepth
 
         public PropertyMap[] GetPropertyMaps()
         {
-            return _sealed ? _orderedPropertyMaps : _propertyMaps.Concat(_inheritedMaps).ToArray();
+            return _orderedPropertyMaps ?? _propertyMaps.Concat(_inheritedMaps).ToArray();
```
jbogard: I'm not sure this is the best place, there's a ctor in PropertyMap that sets all the inherited properties (if they're not already set).

lbargaoanu: I'm not sure what you mean, this is about the order of properties in a TypeMap, so how can a PropertyMap constructor help with that? The order field is set, but is not used because sealed is not set to true yet.

jbogard: Ahhhh I see now. Thanks!

PR #1243, opened 2016-04-24
Generic mappers

jbogard:
Type converters and value resolvers are all generic now.

Mapping directly goes against values without casting to object, ignoring what is in the ResolutionContext.


src/AutoMapper/Mapper.cs
```diff
@@ -186,11 +186,11 @@ TDestination IMapper.Map<TDestination>(object source)
         {
             var types = new TypePair(source.GetType(), typeof(TDestination));
 
-            dynamic func = _configurationProvider.GetMapperFunc(types);
+            var func = _configurationProvider.GetMapperFunc(types);
 
             var context = new ResolutionContext(source, null, source.GetType(), typeof(TDestination), null, _defaultMappingOptions, this);
 
-            return func((dynamic)source, null, context);
+            return (TDestination) func.DynamicInvoke(source, null, context);
```
TylerCarlson1: Doesn't DynamicInvoke have a performance hit cause it has to use reflection to figure out parameter usage?

TylerCarlson1:
So will resolution context's types be phased out in later changes?

src/AutoMapper/Mapper.cs
jbogard: Yep, I'm hoping to fix this in a later PR. Ideally it's only external Mapper calls that use this, not internal ones....eventually.

jbogard:
@TylerCarlson1 Could be? A lot of the IObjectMappers could be converted to something that just builds an expression instead of going off of ResolutionContext. But I don't think they'd ever go away, they're pretty useful.

TylerCarlson1:
@jbogard I guess maybe an IExpressionMap interface inheriting from IObjectMapper that does CanExpress and runs the Expressions through some kind of converter or something.  Maybe refactor the simple assign logic into AssignMapper's implementation, get it to work and go from there?  After assign works and the logic is moved out to IExpressionMap, it should be pretty simple to incrementally do the other mappers who can boil it down to expressions, while leaving the other ones alone.

PR #1248, opened 2016-04-26
Move expression building

jbogard:
This moves the expression building out of TypeMap/PropertyMap and into a separate class.


TylerCarlson1:
I'm so happy to see the stupid fields with Function calls replaced with simple Property getters and setters.

Also wish expression generation looked a bit nicer, but eh that's hard enough as is.
Maybe MakeMemberAccess changed to Property for instances where you just use property name.

TylerCarlson1:
I think I can do the IExpressionMap idea, but it's definitely going to conflict with this PR cause of the movement of expression translation.

No rush right now, will take a while to translate all the mappings.

jbogard:
@TylerCarlson1 I think with the expression mapping stuff that can allow these pieces to get refactored more nicely. But it's just all in one spot at the moment.

PR #1250, opened 2016-04-27
[WIP] Generic object mapper

jbogard:
An example of what I was looking at - a generic IObjectMapper to replace some of the existing ones. I think this is the path of least change for the existing IObjectMappers, and some might also implement expression mappers for inline mapping.

### DIFF
--- a/src/AutoMapper/MapperConfiguration.cs
+++ b/src/AutoMapper/MapperConfiguration.cs
@@ -34,7 +34,7 @@ namespace AutoMapper
             MapperRegistry.Mappers, 
             new[]
             {
-                typeof(ExpressionMapper),
+                typeof(ExpressionMapper<,>),
                 typeof(HashSetMapper<,>),
                 typeof(StringMapper<>),
 #if !PORTABLE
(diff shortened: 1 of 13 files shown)


src/AutoMapper/MapperConfiguration.cs
```diff
@@ -244,6 +254,25 @@ public bool AllowNullCollections
...
+
+                if (genericMapperType != null)
+                {
+                    var srcParam = Parameter(typeof(TSource), "src");
+                    var destParam = Parameter(typeof(TDestination), "dest");
+                    var ctxtParam = Parameter(typeof(ResolutionContext), "ctxt");
+
+                    var mapExpression = Lambda(
+                        ToType(Call(
+                            New(genericMapperType),
```
TylerCarlson1: I still hate you have to new it up every time.  I would much rather it be static call.  Yes you lose interfaces, but at least save on newing up an object on every map.

jbogard: One other motivation is I'd like to see about encouraging these over type
converters. Those are a bit strange because none of the other type map
configuration applies.

Heck, if we can get the config API right we might be able get rid of type
converters all together. It's already weird to have it all mixed together.

PR #1251, opened 2016-04-28
Expand object mappers to provide expression-based mapping

TylerCarlson1:
No need to mess with anything outside the object mappers themselves


TylerCarlson1:
Another thing I want to do with the expressions is where you do [MapTypeCheck](https://github.com/AutoMapper/AutoMapper/blob/develop/src/AutoMapper/Execution/TypeMapPlanBuilder.cs#L306) is to replace it with IObjectMappers.IsMatch and then return the expression for source.

Because all those if statements are really just a compilation of conditions from all the IObjectMappers, just make the mappers make the expressions and not have to go through the IsMatch at map time.

jbogard:
Hmmmm OK, so IObjectMappers are stateless, that's one big difference from type converters. Would you just have either IObjectMapper or IObjectMapperExpression? Would IObjectMapper still be a thing? And would it be expected that anyone would actually implement a custom one of these?

I still want to elevate type converters outside of type maps. You can actually break AutoMapper right now with type converters when you try to apply conventions or global member configurations. And now that type converters are generic there's not much point to doing CreateMap at all - you can infer everything from the generic parameters of `ITypeConverter`, whether it's closed or not.

jbogard:
And yeah - I'm with you on that the IsMatch piece. That was my end goal as well. So there are some good refactoring pieces from my original PR, moving some stuff around the type map plan builder to have it more closely match "resolve value from source" then "map resolved value to destination". I think I'll cherry pick those out into a new PR.

jbogard:
One last thing - I was trying to provide a connection to the IsMatch method and source/destination types. In your cases, you have to duplicate the type information in IsMatch and the generic parameters/constraints in your static method. My other goal was to have that realized once, in the generic type parameters. That would be nice, because now all object mappers will have the same static method. Is there a way we don't have to force every object mapper to have to have this static method? I can see the simple ones, string etc wouldn't have it, but the others?

TylerCarlson1:
As far as IObjectMapper goes I would want to put the Expression function call into IObjectMapper interface.  It still needs a normal Map from resolution context, for things like collection and enum mapping that won't ever use expressions, like class properties.

For now I'm just doing them one by one with IObjectMapperExpression, so I can actually incrementally test and make sure it works.

So are we going with the way I'm doing it to make the mappers that rely on generics more readable, and debug-able?  Possibly expressions as well.  And then you handle passing the IObjectMappers over to the `TypeMapPlanBuilder` and handle the IsMatch and expression generation?

P.S. I'm not super familiar with `ITypeConverter` changes, and their effects on the system as a whole, at the moment. I'll have to look into it at a later time if it doesn't conflict with what I want to do here.

jbogard:
Ah, I remember why I wanted sorta two levels of these. So for some mappers, you want to create a resolution context and junk, some you don't. Those ones that do just primitive conversion, and the enums even, those could be inlined expressions. There's no need to call a Map method anywhere to do ToString or Enum.Parse or those bazillion Convert.ToInt or conversion operators.

I don't think either approach is more or less debuggable, mine was still a class that just closed a generic type, yours is a generic method, but the contents of the method doing the work are more or less the same.

Rather than get all of the object mappers converted, what does it look like with it integrated with TypeMapPlanBuilder? I'm also investigating my approach (that blends both).

TylerCarlson1:
The static method can go anywhere and be named anything.  I just put it in the class since it makes the most sense for the mapping function to be in the same class.  You want to debug has sets, you go to the HashSetMapper class and breakpoint the generic Map method.

If these generic maps are using the Expressions, then it Auto-Associates IsMatch with IObjectMapper by the code it generates.
Instead of 
```
src.Collection = ctxt.Mapper.Map(src.Collection , dest.Collection , sourceType, destType, ctxt);
```
you get
```
src.Collection = CollectionMapper.Map/*<Correct Generics here>*/(src.Collection, dest.Collection, ctxt);
```
which means there's no more IsMatch having to be called.
Assignable mapper and the other non generics will be super simple normal lines of code, and multi line generic stuff will reference a function call which if you look at it you can find and debug.

TylerCarlson1:
I need to replace entire if statement in [MapTypeCheck](https://github.com/AutoMapper/AutoMapper/blob/develop/src/AutoMapper/Execution/TypeMapPlanBuilder.cs#L306) with something like this
```
var typePair = new TypePair(propertyMap.SourceType, propertyMap.DestinationPropertyType);
var match = configurationProvider.GetAllMappers().First(m => m.IsMatch(typePair));
valueResolverExpr = match.MapExpression(valueResolverExpr ,destValueExpr  , ctxtParam);
```
This assumes IObjectMapperExpression interface is merged into IObjectMapper.  All you would need to do is be able to pass the config provider into this function can and update the code and that should be it.  Don't have to change any other existing code base like you had in #1250

Might want to do a FirstOrDefault and null check and throw exception, or ignore map or something.  Not entirely sure what should do if it can fail to match a MappingExpression.

But all those conditions in the if statement are basically the rules that apply to other IObjectMappers that are not AssignableMapper.  And for Assignable mapper it's already returns source which will be valueResolverExpr, so that functionality stays exactly the same.

Also reworked Convert and Dictionary mapper

TylerCarlson1:
Oh yea for ConvertMapper you had casting of bool and byte mixed up, but I fixed that with the change

TylerCarlson1:
Updated code to show how it would work in generating the expression.
If type map exists and has TypeConverter, then seal, if need be, and use that map. (Seal shouldn't have any problems because it shouldn't be relying on sealing other type maps.)
Else if IObjectMapExpression exists that matches and use the converted expression.
Else do normal map.

Hopefully from here on out it's just a matter of making all the IObjectMappers implement IObjectMapExpression and pass all tests.  Then move IObjectMapExpression function into IObjectMapper and be done with it.

Eventually normal map should go away. and you have your TypeConverter then IObjectMapper logic in the expression generation.  Normal Map functionality remains untouched.

TylerCarlson1:
Oh another thing I would like to do with regards to TypeConverters is instead of putting them in TypeMap as a property, put them into a dictionary with typepair and make IObjectMapper that can read those and return the map call from there.  This object mapper would be on top of the Mappers list so it would continue to get priority over all other IObjectMappers, and you get rid of another special if statement.

Not 100% sure if this would cause any issues, but this can be done in a separate commit

TylerCarlson1:
For the IObjectMappers that need to create sub contexts. I have to replace `context.Mapper.CreateObject(context)` with
```
context.ConfigurationProvider.AllowNullDestinationValues
                        ? ObjectCreator.CreateNonNullValue(typeof (TDestination))
                        : ObjectCreator.CreateObject(typeof (TDestination)))
```
in order to get these sub context mappers to work in generating new objects.

Also as far as who adds IObjectMappers to the collection, the only person I know who does it is myself.

P.S. Enums might not be THAT easy to make one line.  It doesn't help that enum mapper has 3 conditions that do 3 different mapping functions

TylerCarlson1:
Any updates on this?  Haven't done anything since last talked about it

jbogard:
Yeah I've tried a couple more. Initially it seems easier, but the problem is the generic parameters aren't as powerful as IsMatch. So it becomes ahhhhh bit of a mess.

Is yours good to go? I thought about perhaps separating out expression mappers from object mappers, but I think you have the right idea there.

jbogard:
And the enums, those really just need to be broken out into multiple mappers. An IObjectMapper shouldn't have more than one condition.

TylerCarlson1:
Yea right now it's good.  The only thing is some IObjectMapper classes inherit from IObjectMapExpression but not all.  The ones that do will make an expression and the ones that don't just use the existing IMappingEngine.Map call.  The rest can be converted in later PRs at any time.

As for separating them out I think it's better they be the same class as they would share the same functionality, especially for the multi lined complex solutions, and IsMatch logic.

Also I did the flag enums already, but that should be easy to break up into 3 separate mappers

TylerCarlson1:
I like it this way now that it's optional to make it generate an expression or not.  That way when I change my Automapper.Collection I just have to update my custom collection mappers to use IObjectMapExpression interface to get the performance boost.

Also if you have any concerns or things wanted changed I have time to work on them tomorrow.
NullableSourceMapper doesn't generate expressions now.  I didn't get the expression generated correctly and commented it out so it would call IMapper.Map and have a working copy that passed all tests.

jbogard:
I think it's OK that some don't do both. It's an opt-in feature and some just don't make sense to do both, like the expression mapper.

PR #1265, opened 2016-05-10
Always generate typed mapping funcs

### LINKED ISSUE #1257: Dynamic mapping with nested dynamic objects
With AutoMapper 4.2.1, when mapping from a dynamic to an instance of an object, AutoMapper preserves top-level destination object properties with null/default values which do not have a corresponding property in the source object. For any nested properties which are dynamic objects, the same does not hold true. This is probably easiest to explain with an example:

```C#
    class SubSetting
    {
        public int SubTimeout { get; set; }
        public string SubColour { get; set; }
    }

    class Settings
    {
        public int Timeout { get; set; }
        public string Colour { get; set; }
        public SubSetting SubSettings { get; set; }
    }

    class Program
    {
        {
            // The SubSettings property is another ExpandoObject.
            dynamic baseSettings = new ExpandoObject();
            baseSettings.Timeout = 1;
            baseSettings.Colour = "Red";
            baseSettings.SubSettings = new ExpandoObject();
            baseSettings.SubSettings.SubTimeout = 11;
            baseSettings.SubSettings.SubColour = "Green";

            // Create another object we will map onto the one above. 
            // Notice that we do not set a Colour or SubColour property.
            dynamic overrideSettings = new ExpandoObject();
            overrideSettings.Timeout = 2;
            overrideSettings.SubSettings = new ExpandoObject();
            overrideSettings.SubSettings.SubTimeout = 22;

            var config = new MapperConfiguration(cfg => { });

            var settings = mapper.Map<Settings>(baseSettings);

            // After this next mapping, settings.Colour == "Red" but settings.SubSettings.SubColour == null.
            mapper.Map(overrideSettings, settings);
        }
    }
```
So as mentioned in the code comments, after the second mapping `settings.SubSettings.SubColour` is null. Is this the intended behaviour for dynamic mapping? I also tried upgrading to the latest MyGet build `5.0.0-ci-1057` to see if it made a difference but that throws the following exception when making the second call to `Map()`:

```
System.InvalidCastException

Unable to cast object of type 'System.Func`4[System.Object,System.Object,AutoMapper.ResolutionContext,System.Object]' to type 'System.Func`4[System.Dynamic.ExpandoObject,TestConsoleApp.Settings,AutoMapper.ResolutionContext,TestConsoleApp.Settings]'.
```

lbargaoanu:
consider existing destinations when creating objects
fixes #1257

### DIFF
--- a/src/AutoMapper/Mapper.cs
+++ b/src/AutoMapper/Mapper.cs
@@ -337,6 +337,10 @@ namespace AutoMapper
 
         object IRuntimeMapper.CreateObject(ResolutionContext context)
         {
+            if(context.DestinationValue != null)
+            {
+                return context.DestinationValue;
+            }
             return !_configurationProvider.AllowNullDestinationValues
                 ? ObjectCreator.CreateNonNullValue(context.DestinationType)
                 : ObjectCreator.CreateObject(context.DestinationType);
--- a/src/AutoMapper/MapperConfiguration.cs
+++ b/src/AutoMapper/MapperConfiguration.cs
@@ -217,13 +217,12 @@ namespace AutoMapper
             return _mapPlanCache.GetOrAdd(mapRequest, _createMapperFunc);
         }
 
-        private Delegate CreateMapperFunc(MapRequest mapReq)
+        private Delegate CreateMapperFunc(MapRequest mapRequest)
         {
-            var tp = mapReq.RuntimeTypes;
-            var typeMap = ResolveTypeMap(tp);
+            var typeMap = ResolveTypeMap(mapRequest.RuntimeTypes);
             if(typeMap != null)
             {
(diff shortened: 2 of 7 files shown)


lbargaoanu:
On second thought I guess this is just replicated code. I should simply make the generic version call the non-generic one.

jbogard:
Yeah, I think so. It's something that I never got back around to.

jbogard:
@lbargaoanu so what was the exact fix for #1257? This test is broken with a later PR and I'm trying to nail down what broke it.

lbargaoanu:
The problem before was that two funcs with different signatures were cached for the same types. Now it's only one signature, the type safe one. Does that make sense? :)

lbargaoanu:
The problem seems to be unrelated. It's not mapping to the passed destination object, it creates a new one. Before it called CreateObject(context).

PR #1275, opened 2016-05-13
Finish IObjectMapExpression conversions

TylerCarlson1:
Want to also point out INullableSourceMapper changed a bit cause there's no reason to null check and do the context assert nulls, because the `Nullable<T>` is a stuct, thus it can't be null and the check isn't needed.  Also it's much simplified to read.

Also make sure when to check nullable to null they are the same types so that double? -> int won't be tried and it will say Un-mapped types instead of can't cast double? to int.

Still going to go through a few more mappers to convert them to IObjectMapExpression.

### DIFF
--- a/src/UnitTests/IMappingExpression/NonGenericProjectEnumTest.cs
+++ b/src/UnitTests/IMappingExpression/NonGenericProjectEnumTest.cs
@@ -73,7 +73,7 @@
         {
             var customers = new[] { new Customer() { FirstName = "Bill", LastName = "White", CustomerType = CustomerType.Vip } }.AsQueryable();
 
-            var projected = _mapper.Map<CustomerDto[]>(customers);
+            var projected = customers.ProjectTo<CustomerDto>(_mapper.ConfigurationProvider);
             projected.ShouldNotBeNull();
             Assert.Equal(customers.Single().CustomerType.ToString().ToUpper(), projected.Single().CustomerType);
         }
--- a/src/AutoMapper/ConstructorParameterMap.cs
+++ b/src/AutoMapper/ConstructorParameterMap.cs
@@ -66,9 +66,6 @@ namespace AutoMapper
 
             if ((SourceType.IsEnumerableType() && SourceType != typeof (string))
                 || typeMapRegistry.GetTypeMap(new TypePair(SourceType, DestinationType)) != null
-                || ((!EnumMapper.EnumToEnumMapping(new TypePair(SourceType, DestinationType)) ||
-                     EnumMapper.EnumToNullableTypeMapping(new TypePair(SourceType, DestinationType))) &&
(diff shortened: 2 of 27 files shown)


TylerCarlson1:
Got all of the IObjectMappers over to IObjectMapExpression except the Enum mapper.  That one's turning out to be really tricky.

Even if I get them all I would still prefer for it to be optional, so that existing IObjectMappers can still work without the knowledge of expression translation.

jbogard:
Oh yeah.

The EnumMapper mapper I could only see that if it were broken out like the
ConvertMapper was, much smaller individual mappers or expressions.

TylerCarlson1:
Ok I converted all of them over.  The EnumMapper I took out of some if statements for CtorParameters, but it didn't cause any tests to fail.  In fact I bet the same logic to PropertyMap expressions can be used for Ctor Parameters to get that performance boost for CtorBenchmarking.  I'll do in another PR if I do though.

So what this means is by default all property maps should never call recursive IMapper.Map calls ever again.  They will still be called for the top most level map, but all the recursive mappings will all come down to function calls or one line statements.

We don't have a benchmark to do so, but I think the Automapper clones have tests for that, so can see things like maps with lists and enums and other things.

src/UnitTests/IMappingExpression/NonGenericProjectEnumTest.cs
```diff
@@ -73,7 +73,7 @@ public void ProjectingEnumToString()
         {
             var customers = new[] { new Customer() { FirstName = "Bill", LastName = "White", CustomerType = CustomerType.Vip } }.AsQueryable();
 
-            var projected = _mapper.Map<CustomerDto[]>(customers);
+            var projected = customers.ProjectTo<CustomerDto>(_mapper.ConfigurationProvider);
```
TylerCarlson1: @jbogard I assume this is what it should of been.  I don't know of mapping IQueryable should enforce projection instead of calling one of the enumerable mappers.

What was funny though was when I refactored Enum Mapper it failed.  But only when I had IObjectMapExpression activated.  It passed fine with just IObjectMapper though.

jbogard: I have no idea honestly.

TylerCarlson1: OK then it's official.  AM doesn't support projection when mapping IQueryable lol

jbogard: Is that some kind of inception thing?

TylerCarlson1: IDK I was just saying if the great @jbogard didn't know then I guess my assumption is correct by default.

PR #1279, opened 2016-05-16
Upgrading project files to the preview 1 tooling

jbogard:


lbargaoanu:
RC2 hurts :)
I can't make it work without changing things. I deleted the global.json (apparently is not required) and recreated the xproj. I'm not if the problems are only on my end or you need to make some changes. The xproj doesn't load because of the error below. Dnvm cannot find "1.0.0-preview1-002702". It does find "1.0.0-rc2-20221".

C:\Projects\AutoMapper\src\AutoMapper\AutoMapper.xproj : error  : The imported project "C:\Program Files (x86)\MSBuild\Microsoft\VisualStudio\v14.0\DotNet\Microsoft.DotNet.targets" was not found. Confirm that the path in the <Import> declaration is correct, and that the file exists on disk.  C:\Projects\AutoMapper\src\AutoMapper\AutoMapper.xproj

jbogard:
You need to install the preview 1 tools as well as the .NET core sdk

lbargaoanu:
That did it. Thanks.

PR #1284, opened 2016-05-17
Make seal last and move newing up create map functions before do config and seal.

TylerCarlson1:
Issue with AutoMapper.Collection it's going to need to call IConfiguration.ResolveTypeMap which is done on seal now.  The problem is the thing gets sealed without the functions to actively run ResolveTypeMaps,  therefore I get exceptions.

### DIFF
--- a/src/AutoMapper/MapperConfiguration.cs
+++ b/src/AutoMapper/MapperConfiguration.cs
@@ -39,15 +39,15 @@ namespace AutoMapper
 
             _profiles.Add(_defaultProfile);
 
+            ExpressionBuilder = new ExpressionBuilder(this);
+            _getTypeMap = GetTypeMap;
+            _createMapperFunc = CreateMapperFunc;
+
             _validator = new ConfigurationValidator(this);
 
             configure(this);
 
             Seal();
-
-            ExpressionBuilder = new ExpressionBuilder(this);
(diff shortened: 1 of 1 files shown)


TylerCarlson1:
Looks like the Auto-Build is stuck downloading. :(

jbogard:
Conflicts :(

TylerCarlson1:
nooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooo

Ok I'll fix

TylerCarlson1:
Done Just wait for build again :(

TylerCarlson1:
Fixed my problem.  AM.Collections is compatible with 5.0 now.  Question is can you make a nuget package depend on another package in a myget feed?  And if so is it going to be super confusing?  Can't depend on beta1 right now cause it won't work. :(

jbogard:
Ah. Nuts.

TylerCarlson1:
Yea sorry about that.  I wait until AutoMapper is in some form of release before trying to integrate into my applications, and I always run into problems when I do.

TylerCarlson1:
Soooooo  Any idea when 5.0 Beta 2 or 5.0 release will be available on NuGet?

jbogard:
5.0 can't be released until .Net Core RTM's

PR #1290, opened 2016-05-19
Reuse the context for collections

lbargaoanu:


lbargaoanu:
@TylerCarlson1 Some cleanup in those mappers would be nice :)

jbogard:
@lbargaoanu What needs to be cleaned up?

src/AutoMapper/Mappers/DynamicMappers.cs
```diff
@@ -21,6 +21,7 @@ public class FromDynamicMapper : IObjectMapExpression
                 destination = (TDestination) (!context.ConfigurationProvider.AllowNullDestinationValues
```
TylerCarlson1: @jbogard @lbargaoanu This perhaps?  It's pretty ugly and probably can be optimized more.

lbargaoanu: Yes, I don't have VS handy right now, but I saw the same code in many of the mappers I've changed. Maybe that code duplication thing in VS would help.

lbargaoanu:
Also in Mapper, all those Map methods look alike.

TylerCarlson1:
The reason they are all named Map for the static call is when you look at the expression generated it will be CollectionMapper.Map or EnumerableMapper.Map and it's consistent between all mappers.  The static class name tells you the mapper that uses it and would be redundant to name the function the same thing.  I don't know if we'll ever show the expression like in debug mode, so that people can see what's going on under the hood, but if so it would be better to rename it to something else, so they don't get confused against the 2 calls both named Map.

I won't be offended if someone else cleans it up a bit and makes it prettier/more readable.  I just brute forced every mapper until all the tests passed, so right now I'm just happy the my performance gain solution worked as intended.

src/AutoMapper/Mappers/DynamicMappers.cs
TylerCarlson1: I did refactor some of these in some of the other mappers where this code was.  Things like Nullable you know the type can only be nullable, so you can just new it up as such.  Maybe these you can do that to some extent, but it wasn't as obvious as the others were.

lbargaoanu:
Yeah, maybe someone will clean it up. Santa? :)

lbargaoanu:
A nice expression [debugger visualizer](https://visualstudiogallery.msdn.microsoft.com/2d5de770-50e9-4dcf-87e9-ea1ed1b43b68).

TylerCarlson1:
That's cool.  Was wondering if there was a way to translate it into something readable.
If you were going to use something like that to show people the expression I would be more inclined to make the mappings look nicer.

Would be great if they can give us the ?. operator though for all the null checking stuff.

jbogard:
Ohhhh that might be interesting for when something blows up, in the
exception message. "Here's your mapping plan" too.

TylerCarlson1:
Gonna have a [nuget](https://www.nuget.org/packages/AgileObjects.ReadableExpressions) dependency at that point though.  Although it looks like it's a PCL so it might be compatible?

PR #1291, opened 2016-05-20
We know you aren't null already quit asking

TylerCarlson1:
Have check for nulls on each property but the source was already checked for null before you got to this point, so there's no need to check source every time.  It's the only place it's used right now so it works with the base being parameter expression.

Not big time saver from what I've seen, but the expression is less cluttered.


TylerCarlson1:
Oops I missed a scenario.  I have to think about this more.

PR #1304, opened 2016-05-24
Configuration expression separating

jbogard:
This allows you to have the configuration expression passed around and modified, before you then pass it to the MapperConfiguration for consumption.


lbargaoanu:
I think someone will try to call CreateMap or something after you've read the config and called Seal. So maybe add a bool sealed and throw on any subsequent call?

jbogard:
InvalidOperationException?

lbargaoanu:
I'm afraid so :) Maybe call into some kind of NullProfile. Although reading could still be useful.

jbogard:
Eh, I can also see that it might be useful to have a base config and then
modify it for other configs.

lbargaoanu:
That could be done by cloning or smth, a different instance anyway.

PR #1314, opened 2016-05-26
Allow custom resolved values to validate; I didn't try to get the typ…

lbargaoanu:
…e returned by the resolver
I think that's the way it worked at some point. Getting the type seemed difficult.


lbargaoanu:
I see the build failed. Some infrastructure thing I suppose :)

PR #1315, opened 2016-05-27
Eliminate some allocations and some refactoring

lbargaoanu:
That mark thing seems useful to me. You can exclude the configuration.


lbargaoanu:
Updated.

PR #1320, opened 2016-06-01
Array mapper improvements

jbogard:
@TylerCarlson1 thoughts? The idea is to expand the expression mappers to be able to examine config and pre-wire some of the decisions they make. Source value/collection null etc.

Also makes the actual mapping part a little more obvious, it's just a LINQ select/toarray, surrounded by expressions that have the extra juice.

### DIFF
--- a/src/AutoMapper/Execution/TypeMapPlanBuilder.cs
+++ b/src/AutoMapper/Execution/TypeMapPlanBuilder.cs
@@ -349,7 +349,7 @@
                     var match = configurationProvider.GetMappers().FirstOrDefault(m => m.IsMatch(typePair));
                     var expressionMapper = match as IObjectMapExpression;
                     if (expressionMapper != null)
-                        valueResolverExpr = expressionMapper.MapExpression(valueResolverExpr, destValueExpr,
+                        valueResolverExpr = expressionMapper.MapExpression(typeMapRegistry, configurationProvider, valueResolverExpr, destValueExpr,
                             ctxtParam);
                     else
                         valueResolverExpr = SetMap(propertyMap, valueResolverExpr, destValueExpr, ref propertyContext);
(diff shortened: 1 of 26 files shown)


TylerCarlson1:
I like the first commit.  That's pretty obvious why changing that is a good thing.

The expression changing logic not so much.  I get the idea and see the performance gain potential.
I don't like it cause the logic is sprinkled in a lot of places and it's harder to debug and view.
Maybe if the way it was implemented was better I wouldn't mind as much IDK.  Personal Preference I guess.

As far as readable and debug-able I guess it's not that bad all things considered.
I wouldn't want to go back and try to debug it if it looks like this again lol.

jbogard:
I added some explaining comments and a small refactoring. Yeah, it's slightly uglier, but I do like the idea of keeping the core mapping function debuggable (that new Map method), but the pre-wired decisions are wrapping that core piece.

TylerCarlson1:
Yea well can't complain too much about syntax.  There definitely is more optimizations that can be done.

I guess after working with Expression generation so much I don't really want to deal with it again anytime soon.
Also whenever you get done refactoring all the IObjectMapExpression mappers can you make a new beta so I can have a working version of Automapper.Collection 2.0 beta go to nuget?

jbogard:
Oh as far as any others - I'm only looking at ones that look at TypeMap configuration to make a decision, to pre-wire that. Only a couple others I think do that.

TylerCarlson1:
Maybe AM will be 4.99 times slower now other than 5.07 now.

Although from what I seen in roadmap, it doesn't map to existing objects, so it's kind of hard to beat that performance wise skipping null checks.

jbogard:
Oh, there's a ton of stuff Mapster can't do that AutoMapper can. The current perf issues are really around ResolutionContext and supporting any potential thing you might want to do with them. For example, we can't inline an expression in a Select because we fill the resolution context on the off chance something below might need it.

TylerCarlson1:
Yea I know ResolutionContext, TypePair, MapRequest, and other object generation is the biggest thing.  Hence the less than 0.1 improvement guesstimate.

Yea I was just looking at it and was surprised that they couldn't even do that.

PR #1327, opened 2016-06-09
It doesn't have to be DynamicObject

lbargaoanu:


TylerCarlson1:
Yes that's technically true, but it doesn't match with the IsMatch.  Also if you end up calling it somewhere else in code in the future you would probably want that dynamic constraint on it.  That's also a big what if that I don't see happening, but you never know.

PR #1329, opened 2016-06-12
[WIP] Expression mapping for collections

jbogard:
This builds out explicit expressions for some of the expression mappers


src/AutoMapper/ExpressionExtensions.cs
```diff
@@ -11,6 +12,38 @@
...
     internal static class ExpressionExtensions
     {
+        public static Expression ForEach(Expression collection, ParameterExpression loopVar, Expression loopContent)
+        {
+            var elementType = loopVar.Type;
+            var enumerableType = typeof(IEnumerable<>).MakeGenericType(elementType);
+            var enumeratorType = typeof(IEnumerator<>).MakeGenericType(elementType);
+
+            var enumeratorVar = Variable(enumeratorType, "enumerator");
+            var getEnumeratorCall = Call(collection, enumerableType.GetMethod("GetEnumerator"));
```
lbargaoanu: Here it would be interesting to not go through IEnumerable (use the pattern), in order to avoid allocating the Enumerator, just use the struct, for lists for example. I'm just not sure it would work without special casing collection types.

TylerCarlson1:
I wonder with the collection mapping, you will be using their collection types instead of the actual types.  So like for things that inherit from the base element types won't be found and used possibly.  I think I've read other mappers having that problem with collections, and it looks like it will introduce said problem by pre-defining the types.

I could be wrong, but I thought normal mapping would use actual type then type pair types to map and object, and pre-compiling it will make sure that won't be the case for collections.  Could cause a problem if you have parent object that has children classes that inherit from one another and uses the base mapping vs figuring out the types before mapping.  Might be a workaround to this though, but not entirely sure atm.

jbogard:
@TylerCarlson1 yeah, I don't know to be honest. All the tests passed today without any modifications, mainly because once I found a matching type map I deferred to the existing Map pipeline. What might be interesting is if I switched the IObjectMapper version to use the expression logic, and see what happened. But I do want to preserve that flexible behavior that we have today.

src/AutoMapper/ExpressionExtensions.cs
jbogard: Yeah, I think that this was intending to emulate the for loop as close as possible. But it also doesn't dispose the enumerator, so who knows.

src/AutoMapper/Mappers/CollectionMapper.cs
```diff
@@ -48,11 +52,84 @@ public bool IsMatch(TypePair context)
...
+            {
+                var destParam = Parameter(newExpr.Type, "dest");
+                blockParams.Add(destParam);
+                blockExprs.Add(Assign(destParam, newExpr));
+                destExpression = destParam;
+            }
+
+            Expression itemExpr;
+
+            var match = configurationProvider.GetMappers().FirstOrDefault(m => m.IsMatch(typePair));
```
TylerCarlson1: This looks like duplication of code.  Maybe make a function in typeMap pairing that has only this code so you don't have two version of this.  Like `GetMapExpression(IEnumerable<IObjectMapper> mappers, TypePair typePair)` or something like that.

TylerCarlson1:
@jbogard Yea making map use expression should show any possible errors.  I would hope the inherited type tests get run through the CollectionMapper and not any of the other enumerable mappers.

jbogard:
This is part of my motivation, performance for enumerable-based mappers SUCKS:

```
AutoMapper: - Flattening - Mapping time:        0.549014682722037s
Manual: - Flattening - Mapping time:    0.0148397660401265s
Manual: - Flattening - Mapping time:    0.0361116440097974s
CtorMapper: - Ctors - Mapping time:     0.281245287564524s
ManualCtorMapper: - Ctors - Mapping time:       0.00638705100807081s
Complex Types: - Mapster - Mapping time:        14.1827046368134s
Deep Types: - Mapster - Mapping time:   2.90722893178847s
```

TylerCarlson1:
Yea I know.  Oh another possibility to increase performance would be to make another public call that's generic but also passes over resolution context.

I think the main problem with performance drop might be dynamic invoke by calling map without generic.  You can call map without context and regenerate it, but it's probably faster than calling dynamic invoke.
Maybe create a generic version of Map in IMapper where you can pass the context over to get the most performance.
It's still will be slower than caching it because it has to call source.GetType() and make MapOptions but it should be faster in general.  Something online looks like dynamic invoke is 200x slower than invoke, so that should really speed things up.

I don't have a PR for it, because there might be some other problems

TylerCarlson1:
Ok I was looking at old code when I said that.  I got that wrong.
You could make a Generic version of `IRuntimeMapper.Map(ResolutionContext context)` and will eliminate some un-needed casting.

Also in unrelated news it looks like ResolutionContext.Map doesn't use sourceType parameter.

TylerCarlson1:
Oh also in your running example you have complex and deep, but you don't have Manual Mappings for them to compare against.  How do you know how much slower they are compared to simple maps?

jbogard:
Ah yes. I'll add that.

TylerCarlson1:
Ok yea from what I can see the compiling of the expression mapping thing looks like there's no conflicts from what I can tell.

Also as an alternative to 100% expression generation what do you think of something like this?
```
public static TDestination Map<TSource, TSourceItem, TDestination, TDestinationItem>(TSource source, TDestination destination, ResolutionContext context, Func<TDestination> createDestFunc, Func<TSourceItem,ResolutionContext, TDestinationItem> newItemFunc)
            where TSource : IEnumerable<TSourceItem>
            where TDestination : class, ICollection<TDestinationItem>
        {
            if (source == null && context.Mapper.ShouldMapSourceCollectionAsNull(context))
                return null;

            TDestination list = destination ?? createDestFunc();

            list.Clear();
            var itemContext = new ResolutionContext(context);
            foreach (var item in source)
                list.Add(newItemFunc(item, itemContext));

            return list;
        }
```

You compile the createDestFunc and newItemFunc and pass them in.  Maybe even `context.Mapper.ShouldMapSourceCollectionAsNull(context)` can be turned into a bool so it's fast.
You lose `list.Clear()` performance gain on new objects, but you get the speed of pre-compiled constructors and mapping operations without having your code be a big blob of expression mess.  Also it is still debugable, to a degree and less complicated Expression generation code.

This might even be possible to put the collection mapper into a single static function across all IEnumerable Mappers, cause the main differences are mostly in the constructor.

TylerCarlson1:
@jbogard Any thoughts about continuing to use a static function and pass over a pre-compiled newCollection and addItem expression?

Note: I tried to get it to work but having issues with parameter passing.  I assume that's why you used MapMethodInfo2 in ArrayMapper to bypass that problem.

jbogard:
So I just added some more commits (sorry for the delay, had some failing tests). I added a new mapper that focuses on generic enumerables - the existing collection mappers deal with IEnumerable and IList, which made them kinda annoying to deal with.

I also made a slight change to remove any call to IObjectMapper.Map, and everything works off of the expression mapper's map method. That has the added benefit that we don't have to support two pipelines. It still supports calling an internal Map method, but everything goes through building an expression.

It doesn't force everything to use an optimized expression tree, but it does encourage you to build an expression.

src/AutoMapper/MapperConfiguration.cs
```diff
@@ -359,6 +359,13 @@ private static LambdaExpression GenerateObjectMapperExpression(MapRequest mapReq
                     var message = Constant("Missing type map configuration or unsupported mapping.");
                     fullExpression = Lambda(Block(Throw(New(ctor, context, message)), Default(destinationType)), source, destination, context);
                 }
+                else if (mapperToUse is IObjectMapExpression)
```
TylerCarlson1: Wow how did we miss this one.  That can almost be it's own commit by itself.

jbogard:
Can I close this @TylerCarlson1 ?

TylerCarlson1:
Yea I think so.  I got most if not all the functionality out of it in the other PR.

PR #1334, opened 2016-06-15
Remove Unneeded casting from ResolutionContext.Map

TylerCarlson1:
Got About 15-20% performance boost from Complex Types from Mapster test in #1329


TylerCarlson1:
Missed some things.  Also now Mapper Expressions generated and All the mappers, except for a few now call Generic Versions of `ResolutionContext.Map`.  That eliminates 4-5 castings back and forth from object to the type per item in Collection and PropertyMap that uses a type map.

So probably not a 20% increase in performance for regular mappings with property sub maps, but there is a performance none the less.

PR #1335, opened 2016-06-15
[WIP] CollectionMapper pass compiled expressions

TylerCarlson1:
@jbogard This is generating the new instance and the add new item instance while sharing the base generate item code in #1329.

The new expression will be pre-compiled based on the type you pass over.  And the new items to be added will be either one of the expression mappings that the IObjectMapExpression generates, or calls ResolutionContext.Map.  You get the performance boost of sub expression generation, without the headache of tons of Expression calls.

I think with refactoring I can get it to work with the other Collection Mappers that don't follow this exact pattern to share the same code base.  ReadOnlyCollection, Dictionary, ect.

### DIFF
--- a/src/AutoMapper/Execution/TypeMapPlanBuilder.cs
+++ b/src/AutoMapper/Execution/TypeMapPlanBuilder.cs
@@ -349,7 +349,7 @@
                     var match = configurationProvider.GetMappers().FirstOrDefault(m => m.IsMatch(typePair));
                     var expressionMapper = match as IObjectMapExpression;
                     if (expressionMapper != null)
-                        valueResolverExpr = expressionMapper.MapExpression(typeMapRegistry, configurationProvider, valueResolverExpr, destValueExpr,
+                        valueResolverExpr = expressionMapper.MapExpression(typeMapRegistry, configurationProvider, propertyMap, valueResolverExpr, destValueExpr,
                             ctxtParam);
                     else
                         valueResolverExpr = SetMap(propertyMap, valueResolverExpr, destValueExpr, ref propertyContext);
--- a/src/AutoMapper/IObjectMapper.cs
+++ b/src/AutoMapper/IObjectMapper.cs
@@ -32,10 +32,11 @@ namespace AutoMapper
         /// </summary>
         /// <param name="typeMapRegistry"></param>
         /// <param name="configurationProvider"></param>
+        /// <param name="propertyMap"></param>
         /// <param name="sourceExpression">Source parameter</param>
         /// <param name="destExpression">Destination parameter</param>
         /// <param name="contextExpression">ResulotionContext parameter</param>
         /// <returns>Map expression</returns>
-        Expression MapExpression(TypeMapRegistry typeMapRegistry, IConfigurationProvider configurationProvider, Expression sourceExpression, Expression destExpression, Expression contextExpression);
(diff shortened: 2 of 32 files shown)


jbogard:
Is this intended to merge with my PR?

TylerCarlson1:
No it's meant to show an alternative implementation that should accomplish the same thing as your PR, if not a bit slower.  Sorry for the confusion.

TylerCarlson1:
It's meant to make `CollectionMapper` like [GenericEnumerableToEnumerableMapper](https://github.com/AutoMapper/AutoMapper/pull/1329/commits/0e6de56d81b932585efb9f8fe7f6e57008aa5063#diff-cb9de3883712820918d3d40fb43530c6R15), as a base for enumerable mappings.  Only instead of defaulting to `List<>` you pass that over, so `HashSet<>` works, and then you can modify it more so that other collection types work.

And then after you get it to work with all Enumerable types, you could then convert the functions code from a map call to #1329's purely expression based code, for maximum performance increase.

TylerCarlson1:
I've refactored to extension class to make it more readable
I've been able to get this to work with Collection, HashSet, and ReadOnlyCollection mapper where I get speed up of precompiling the mappings and collection generation.

I should be able to, with some more tinkering, get Enumerable, Dictionary, EnumerableToDictionary, to share the same update code.

Also can use `CollectionMapperExtensions.NewExpr` to pre-define constructors for other mappers and get rid of code like this
```
(TDestination)(context.ConfigurationProvider.AllowNullDestinationValues
                        ? ObjectCreator.CreateNonNullValue(typeof(TDestination))
                        : ObjectCreator.CreateObject(typeof(TDestination))
```

Once everything is converted and optimized then you can refactor `CollectionMapperExtensions.MapCollectionExpression` and `CollectionMapperExtensions.MappCollection` to reflect #1329's pure expression generation approach for even more performance gain.  I guess I'm just waiting to do Expression Generation as the VERY LAST STEP since I don't want to have to debug it every time I need to change code to make another `IObjectMapExpression` class work with it.  Debugging Expression Generation isn't fun at all.

@jbogard What do you think?

jbogard:
Yeah, I like it. So I've got expression generation in my PR, plus that other thing that pulls in the expression? Just trying to see what I can pull out of mine if I merge this.

TylerCarlson1:
I would say [MapExpression](https://github.com/AutoMapper/AutoMapper/pull/1329/files#diff-cb9de3883712820918d3d40fb43530c6R172) needs to be translated, and [IObjectMapExpression Cache](https://github.com/AutoMapper/AutoMapper/pull/1329/files#diff-f31ea6c507cd85757f8dfe83cc7282d9R362) could be it's own PR independent of this and yours.

The rest would be the benchmark code and possibly the unit tests.
Benchmarks could be their own PR as well.

TylerCarlson1:
I think I've done most of the optimizing I can do for now.  I added last 2 commits of #1329, the benchmark and IObjectMapExpression optimization.

Oh I did change Array mapper to use IObjectMapExpression, so it now uses assignable mapper and one test fails, but I skip it now.  I don't know if it's old or does that matter or not.  I would think it doesn't because AutoMapper is not an Object Cloner.

All you would have to do is update `CollectionMapperExtensions.MapCollectionExpression` and `CollectionMapperExtensions.MappCollection` to use a pure expression based system where you generate the block code to replace `CollectionMapperExtensions.Map`.

Also I'm sure there's a good chunk of refactoring and renaming that will probably occur as well.

@jbogard If you have any questions, comments, concerns feel free to ask.

jbogard:
OK that ArrayMapper stuff, I dunno, people have weird expectations about that. I'm OK with it just assigning, arrays are immutable (size, not contents). Should be OK.

Issue #1330, opened 2016-06-14
Error trying to map System.Collections.Immutable.ImmutableDictionary


carlos-sarmiento:
I want to map a POCO into an `ImmutableDictionary<string, object>` and Automapper is raising an exception because the `Add` operation is not supported in `ImmutableDictionary`. I'm using this mapping:

```
var t = new MapperConfiguration(cfg =>
                    cfg.CreateMap(@event.GetType(), typeof(StoredEvent))
                         .ForMember("DataBag", opt => opt.MapFrom("Data")));
```

and getting this error:
```
Mapping types:
Dictionary`2 -> ImmutableDictionary`2
System.Collections.Generic.Dictionary`2[[System.String, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089],[System.Object, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089]] -> System.Collections.Immutable.ImmutableDictionary`2[[System.String, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089],[System.Object, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089]]

Destination path:
StoredEvent.DataBag.DataBag.DataBag
```

To solve this I tried to use a custom resolver:
```
public class ImmutableDictionaryResolver : IValueResolver
{
    public ResolutionResult Resolve(ResolutionResult source)
    {
        var dictionary = Mapper.Map<Dictionary<string, object>>(source.Value);
        return source.New(dictionary.ToImmutableDictionary());
    }
}
```

With this mapping:
```
var t = new MapperConfiguration(cfg =>
                    cfg.CreateMap(@event.GetType(), typeof(StoredEvent))
                        .ForMember(nameof(StoredEvent.DataBag),
                            opt => opt.ResolveUsing<ImmutableDictionaryResolver>().FromMember("Data")));
```

But I'm still getting the same error. In the second instance, it complaints with this:

```
Mapping types:
ImmutableDictionary`2 -> ImmutableDictionary`2
[...]

PR #1338, opened 2016-06-15
Allow overriding the built in object mapper with a custom type converter

lbargaoanu:
As in #1330. I could swear this used to work :)
Whitespace in UnitTests/TypeConverters.cs. The test is at the top.


jbogard:
Oh weird. The familiar "resolving vs. explicitly loading" a type map problem.

PR #1341, opened 2016-06-17
Collapse object mapper

jbogard:


src/AutoMapper/Mappers/CollectionMapper.cs
```diff
@@ -44,35 +44,6 @@ public static class CollectionMapperExtensions
             return resolutionContext.Map(item, default(TDestinationItem));
         }
 
-
```
TylerCarlson1: Don't forget to remove MapItemMethodInfo, MapKeyValuePairMethodInfo, MapItemFunc, and MapKeyValuePairFunc since they tie in with MapCollection.

PR #1344, opened 2016-06-18
Better Array Mappings

TylerCarlson1:
ToArray has to double enumerate.
You could go a bit further and make array mapper pure expression based, and from that you can use source.Length in cases of array instead of relying on source.Count() which will save you time in array to array mappings.  That can be saved for a later PR I guess

### DIFF
--- a/src/AutoMapper/Mappers/ArrayMapper.cs
+++ b/src/AutoMapper/Mappers/ArrayMapper.cs
@@ -18,8 +18,13 @@ namespace AutoMapper.Mappers
         {
             var itemContext = new ResolutionContext(context);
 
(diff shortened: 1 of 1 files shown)


lbargaoanu:
I think that's the way it was before. In the general case Count() will enumerate too.

TylerCarlson1:
I think ToArray actualy enumerates the lamda twice, so you get 2 Maps per item, and that's why this is better.  ToList doesn't have this problem because you don't have to allocate the size of it before hand.

lbargaoanu:
No, it's exactly the other way around :) For non ICollection-s ToArray enumerates only once, your way twice (Count() and foreach). But maybe nobody will notice.

PR #1346, opened 2016-06-19
[WIP] Compile method in class instead of general expression.

TylerCarlson1:
Got 50% performance gain from benchmarks the problem is that you have to make things public to get it to work.
I also took everything out of the map like generating type maps and resolution contexts, and before/aftermap and a lot of other things and got native code level speeds.  Without this I got 5-6x slower, so this looks to be "The Gold Standard" for compiling expressions.

Only problem is everything has to be public that you call, since you are making a new class object to run the code.  Also constants can't be used they have to be generated.  So a lot of unit tests fail because of this.

Links to explain the madness: http://stackoverflow.com/a/5573273 and http://stackoverflow.com/questions/5568294/compiled-c-sharp-lambda-expressions-performance

@jbogard What are your thoughts on this?  I'm not sure where and how I'm going to change the code at this point.  I have to re-do parts of the expression tree, cause I can't pass constants like the PropertyMap in this way, or call the before and after map functions I think.  If you want to look into this one maybe I'll let you lol.


TylerCarlson1:
Aww it looks like this functionality will only work in .NET and as far as I can tell right now, it's not going to be supported in .NET Core :(

TylerCarlson1:
Welp @jbogard do you want to change something that makes AutoMapper expressions run as fast as hand code, but only works up to .NET 4.5 and make it work with .Net Core with "possible" slight degraded performance?

TylerCarlson1:
Oh and you could probably get rid of resolution context with this as well.  Since they are functions in a class you can have a complete stack trace of the functions called and at what point they failed.  You won't have the property map possibly, but you should have the parent map calls and with proper naming or types you could make the stack trace show the type map that had the exception. Maybe even have line numbers in the class and you can go and look at the compiled expression and figure out the mapping there.

Won't be exactly like the other one in error message format, but you might get rid of Resolution Context completely.  Just pass in the object values and their source and destination types.  The map function and types generated will tell you what was being mapped to what.

TylerCarlson1:
Oh boy this one's a monster to do.  What I got to do to make this work is make everything that invokes an anonymous function, be compiled into a method or call a function on one of the parameters.

That means BeforeAfterMaps, TypeConverters, ConvertUsing, set ResolutionContext.PropertyMap for exception handling, and possibly a bunch more stuff.

Going to take a long time to make all the tests pass.  Possibly miss .Net Core RTM, which I don't know if that means 5.0 will be out of beta when that releases.

lbargaoanu:
I think it would be nice to know why it's faster before doing a lot of work. It's not obvious to me why it would be faster done this way. The links don't help. Maybe it's some implementation detail that could be fixed in the BCL.

TylerCarlson1:
I've looked at alternatives to make it faster with compile, but I can't get a performance increase.  Maybe it has to load the function from memory, or re-generate the expression tree every time, or maybe it has to convert the function into IL code before execution.  I don't know why it's faster I just know it is.  The only thing I have seen a difference is if the lambda is one line it runs just as fast.  If it's multiple lines or a blocking call or something it's slower.  But I haven't confirmed that as a fact.  We obviously need blocking code, so a single method call is out of the question.  Maybe blocking code can be better made to compile faster IDK.

TylerCarlson1:
@jbogard Looks like I need to reference all the constants that are generated into some property or method call in AutoMapper to get this to work.  Made `TypeMap<TSource,TDestination>` static calls so can pass in the typeMap without using a constant.

What's left is to get property maps and actions and all these other constants from said TypeMap.  Maybe the final step is to make TypeMap and PropertyMap generic so I don't have to cast out the types.

Performance on simple maps drops cause of current implementation but still faster and could be optimized more.

jbogard:
One thought on the resolution context thing - I hate the damn thing, my preference would have been to make that thing only have actual contextual stuff. The mapping operation options, the cache, context stuff. The source/dest objects, people do weird junk of trying to walk a hierarchy to a parent object.

The exception handler can capture all the context it needs, so having a resolution context JUST for debugging purposes is silly, we can do it in the catch block.

Eventually I'd want to see a single resolution context per Map call, and not stuff the source/destination objects or types into it. None of the rest of AutoMapper goes off of ResolutionContext anymore anyway.

jbogard:
As a side thing I can create a WIP PR to show what that would look like. If people want to walk parent hierarchies they can just build that into their model and not force it into the stuff I have.

TylerCarlson1:
I think that resolution context stuff could be an independent PR for the most part.  I'm not sure though, as I'm trying to move references around.  Either way a PR would help if you can make it like that.  Maybe only set the properties when an exception has actually happened, or new class with resolution context and typeMap and propertyMap.

I think for the moment, is I'm going to get this thing to work the current way with resolution contexts as is.  Because I can't make this happen in .Net Core it's easier if the exception messages are consistant for now.  But I think if you get your resolution context not to hold the type pair and type maps, and modify the expression and exception handling to get the same result, I should be able to merge that into mine.

jbogard:
Is it OK if the exceptions get the contextual type pair/type map/property map? That's the route I was going.

TylerCarlson1:
Yea that's what I was thinking you would be doing.  I'll just have to modify the sets of those in the exception to work on non-constants, so it works with AppDomain.

TylerCarlson1:
There's some issues with `IgnoreAllPropertiesWithAnInaccessibleSetter` and moving code to Dynamic Assemly.  The new dyanmic dll doesn't have permissions to access private setters for tests that have them.  A lot of tests fail saying they don't have access to the properties.  Also things like protected and internal will try to map by default but won't because of inaccessable from Dll as well.

I think it's because it's generated in memory, it can bypass property rules, but with AppDomain it won't allow them.

@jbogard Thoughts on the current version of AutoMapper having ability to access private, internal, protected properties, if you don't ignore them?

jbogard:
I think that's the bigger deal, there's an expectation that if the member exists you can map to it, at least for setters.

TylerCarlson1:
Funny I always thought if it was private it would be auto-ignored.  I guess that assumes the mapping code you are replacing will be in another class and not in the class that it maps from.  Well I have never had to deal with that before so what do I know.

jbogard:
It ignored private properties but not private setters.

TylerCarlson1:
Welp there's nothing much more I can do with this then I think.  Cause if I'm going to access it privately I need go back to propertyinfo.setvalue possibly and I don't want to do that as it's slower and I then need to load the property info into the class as well.

jbogard:
So what was the original reason for slowness? Not all internal lambdas were compiled immediately? I didn't quite follow the SO post

TylerCarlson1:
IDK either.  I just know there's a 5x-6x slowdown in Expression Tree function execution.

I haven't found a definitive answer why Compile is slower than CompileToMethod.

TylerCarlson1:
Tried removing functionality to find a speed up and tried almost everything, except remove the need of a block statement.  So my best guess why it's slower is the block statement isn't compiled optimally, or the return part of it isn't.  Maybe you can profile the execution of the func to see what it's doing behind the scenes.

TylerCarlson1:
Maybe this is it. http://stackoverflow.com/a/4211766

tl;dr;  `lambda.Compile()` is basically a DynamicMethod

TylerCarlson1:
The only other option for speedup I could see is making the Func be an Action and take a single object that had src, dest, resolutionContext (with your PR of course) and then not return anything, but have the dest value be returned instead.  That way there is 1 parameter and no return value, possibly speeding up by looking at 1 value instead of 4.  But then that opens up a whole new can of worms that you are trying to contain right now.

jbogard:
OK so the other thing I think is the problem is that we're mixing around Lambda and Invoke in many places. Really the Lambda should only be built at the very top MapperFunc level, and everything else should be working with just Expressions. We do a lot of Lambda wrapping Lambdas, which might result in the delayed compilation I think.

Rather than invoking a lambda in a lambda, I think we should get rid of all Invoke's and replace parameters as need be.

jbogard:
At least that's what the accepted answer to the first link said, DynamicMethod forces the whole thing to get compiled. The last PR had another good boost in perf, I think that one's ready to go. And the depth checks I think are simplified - instead of keeping track of a hierarchy of ResolutionContext's, it just increments/decrements a simple counter.

TylerCarlson1:
My understanding of the answers were the compiled lamda vs hard coded lambda execute at the same speed.  It's the fact when you call the methods themselves is where all the difference is.  Even though they are the exact same type and perform the exact same function, the one compiled from lambda will take longer because invoking it requires more steps than invoking a hard coded one.

This is because of something like the parameters are casted to object[] and passed in and result is object casted to result.  I have gotten a Very Simple lamdba statement in flattening with only 3 lines and no invoke calls and the perf hit is still there.

jbogard:
I wonder, is this something we can detect? I know too that when I had Silverlight it would only allow mapping to public members, so I had to detect that at config time. It's such a small difference in code between CompileToMethod and not, is why I ask.

TylerCarlson1:
I think so. The question is this detection .net 4.5 specific or not?  It's harder in .net 4.5 only because I have to find the try catch and remove after full lambda has been generated and about to be compiled. There might be a way to get internal by class attribute but not sure how to even add those.

It should be doable by checking the dest property's setter permissions although I have to google it first to see.

TylerCarlson1:
So... Ignore Private only for 4.5, or ignore private for all versions?

TylerCarlson1:
oh and possible expected AM functionality when trying to map to Private.

jbogard:
More like, if the map only goes to public members, use the faster way, else
revert to Compile, but ONLY for 4.5, the rest stay the same.

TylerCarlson1:
hmm.  I guess I can do a TryCatch on the CompileToMethod and if it fails just do Compile() if that's the case.  I can get all the tests to pass easy like that lol.

jbogard:
Ha! That's not a completely terrible idea...in fact it's what is
recommended to check to see if a type can be used to close a generic one
with constraints, try catch

TylerCarlson1:
Even though the tests pass, this still isn't near done.  I have to double check everything and make sure I didn't screw up anywhere.

Also it's horrible with casting right now, so I need to look into that.

jbogard:
http://stackoverflow.com/questions/5053032/performance-of-compiled-to-delegate-expression/5160513 from the other post @lbargaoanu mentioned. If this works, no need to jump through hoops it seems. If.

TylerCarlson1:
Yea I looked into that one and I couldn't get it to work.  Maybe I don't know where they are suppose to go, or what.

TylerCarlson1:
@jbogard I thought you were talking about solution with
```
[assembly: AllowPartiallyTrustedCallers]
[assembly: SecurityTransparent]
[assembly: SecurityRules(SecurityRuleSet.Level2,SkipVerificationInFullTrust=true)]
```
Which I tried and can't get to work.

The ONLY thing I saw from that SO question that works and give speedup is `CompileToMethod` option, which is what this PR is all about.

If there's some other possible solution/workaround I missed please tell me.

lbargaoanu:
I've tested and it works (in LINQPad :), since I don't have VS handy). But anyway, I'm not even sure we should be messing with those attributes. At some level this must be a bug. There should be a straightforward way to avoid the cost using LambdaExpression.Compile.

TylerCarlson1:
@lbargaoanu Does this go in AutoMapper or Benchmark to get the performance benefit?

It goes into AutoMapper with no change and Benchmark won't work with stopwatch when I have all 3.
And I don't see these attributes in SO questioner's source control example.

lbargaoanu:
I thought in AutoMapper would be enough. You can try both too :) As I said I don't have VS handy. I didn't really follow the logic of the solution. I just profiled and went from there. But I see what he did there, worked around it. It's just silly to have to do something like this. I'll have to think about this a little more to see if it makes sense to tell MS that they need to fix it. Because clearly this is by design. I saw the code where they put the attributes :) But still, looks like a bug to me.

lbargaoanu:
Funny enough, the dynamic method way seems a little faster, but maybe it's just noise.

TylerCarlson1:
@lbargaoanu I did try both and it doesn't help either way.  If you get a chance to look at it in VS again please try to prove it works :).

Funny enough is my original post had a link to an Answer I got from the SQ question you and @jbogard posed.  Which was the `CompileToMethod` solution

TylerCarlson1:
Yay it's finally passing.  **DO NOT SUBMIT**

TylerCarlson1:
There's still a few problems left to fix.

1. Doesn't work with multiple IMappers (uses a static call to get TypeMap,  need to make it not static and possibly an extension method from context to get mapper and map types)  Other options is to pass in TypeMap as a parameter to solve this.

2. Tons of castings could possibly eliminate (Have to invoke everything because it might be pointing to something that the compiled assembly might not have access too)
- Maybe even make dynamic classes for each type map with everything cached with correct types.

3.  Some functionality might still not work (I think IsMatch try catch calls don't pass the typemap when throws exception)

4. Code is a MESS (It has a lot of duplicate code that will probably change with issue 1, so need to make that re-usable and look nicer)

TylerCarlson1:
Fixed most of the issues.  There's probably more things that I possibly missed but for the most part I got everything I could think of that would cause a problem.  Maybe I'll think of more later.

For Item 2:  Still there with property map casting function calls, but I would have to make Dynamic TypeMap and PropertyMap classes in order to cast the delegates into their proper types at expression compile time.  This is do-able I think, but I think the perf gain vs how those funcs are exceptions not the rule, it's not worth the time.

Other thing might be the private property stuff and weird map situations I might of thought of.
Oh and GUID doesn't work with `CompileToMethod()` cause it can't compile `default(GUID)`

Overall I think I'm done with the PR. **BUT** please review it very thoroughly.
I'll be happy to answer questions on anything I changed.

P.S. Invokes are back because the lambda could be pointing to an anon, or private function and then I can't call `CompileToMethod()` and get the perf gain.

jbogard:
CONFLICTSSSSSSS

TylerCarlson1:
Bad StopWatch code from @lbargaoanu's PR

src/Benchmark/BenchEngine.cs
```diff
@@ -16,21 +16,19 @@ public BenchEngine(IObjectToObjectMapper mapper, string mode)
 
 		public void Start()
 		{
-			var timer = Stopwatch.StartNew();
-
 			_mapper.Initialize();
 			_mapper.Map();
 
-			timer.Start();
+            var timer = Stopwatch.StartNew();
```
TylerCarlson1: @lbargaoanu Yea so ` Stopwatch.StartNew()` starts counting immediately.
That's why I was seeing a big slowdown in performance looking at the numbers.

src/Benchmark/BenchEngine.cs
```diff
@@ -16,21 +16,19 @@ public BenchEngine(IObjectToObjectMapper mapper, string mode)
...
-			for (int i = 0; i < 1000000; i++)
+            for (int i = 0; i < 1000000; i++)
 			{
 				_mapper.Map();
 			}
 
 			timer.Stop();
 
-			Console.WriteLine("{0}: - {1} - Mapping time: \t{2}s", _mapper.Name, _mode, timer.ElapsedMilliseconds/1000.0);
+			Console.WriteLine("{0}: - {1} - Mapping time: \t{2}s", _mapper.Name, _mode, timer.ElapsedTicks/(double)TimeSpan.TicksPerSecond);
```
TylerCarlson1: More numbers is better :)

TylerCarlson1:
Should be fine now.

src/AutoMapper/Execution/TypeMapPlanBuilder.cs
```diff
@@ -37,39 +41,27 @@ public static LambdaExpression BuildMapperFunc(TypeMap typeMap, IConfigurationPr
...
-                            Call(
-                                MakeMemberAccess(ctxtParam, typeof (ResolutionContext).GetProperty("Options")),
-                                typeof (MappingOperationOptions).GetMethod("CreateInstance")
-                                    .MakeGenericMethod(type)
-                                ),
-                            converterInterfaceType),
-                        converterInterfaceType.GetMethod("Convert"),
-                        srcParam, ctxtParam
-                        ),
+                    ctxtParam.Property("Options").Call("CreateInstance", type)
```
jbogard: Nice.

src/AutoMapper/TypeMap.cs
```diff
@@ -328,4 +338,39 @@ private void ApplyInheritedTypeMap(TypeMap inheritedTypeMap)
             }
         }
     }
+
+    public class TypeMap<TSource, TDestination>
+    {
+        private static readonly ConcurrentDictionary<IConfigurationProvider, TypeMap<TSource,TDestination>> Cache = new ConcurrentDictionary<IConfigurationProvider, TypeMap<TSource, TDestination>>();
```
jbogard: What's all this then?

jbogard: Because this type is generic, you get multiple caches per concrete generic type.

src/AutoMapper/TypeMap.cs
```diff
@@ -328,4 +338,39 @@ private void ApplyInheritedTypeMap(TypeMap inheritedTypeMap)
             }
         }
     }
+
+    public class TypeMap<TSource, TDestination>
```
jbogard: So why do this instead of just going through the type map registry?

src/AutoMapper/TypeMap.cs
TylerCarlson1: You can have multiple configurations making different mappers.  If both configurations map the same type map but do it 2 different ways, I'll need to know which config it came from so I can get the correct type map.

You know the questions about cloning a config and then changing a TypeMap.  Also requests to have 2 different implementations of mapping the same types to each other, based on some condition or use case.

src/AutoMapper/TypeMap.cs
TylerCarlson1: This was taken from looking at Mapster's code originally.  The idea is since this is a concrete type, it's faster to access than calling ResolveTypeMap in configurationProvider.  If you go through configurationProvider, you have to search a dictionary, where as here you have direct access to the object.

This isn't just for exceptions.  This is needed for all the custom expressions for type map and its property maps.  You don't want to have to call ResolveTypeMap and create a new TypePair every time a condition needs to be called when mapping.

For Mapster it makes it faster to find the mapping expression after created.  No need to re-look for type maps if it's already stored somewhere you can access.  Obviously we can't exactly the same effect, since we have to call `GetType()` and then can't easily call a generic method.

TylerCarlson1: Also this can be done in different classes with different means to get a similar result. I don't have a problem refactoring this into something else.
It works with `CompileToMethod()`'s restrictions and doesn't look ugly as sin, so I left it like this.

src/Benchmark/BenchEngine.cs
lbargaoanu: My bad :) I didn't notice because I was testing with or without the attributes.

src/AutoMapper/TypeMap.cs
jbogard: Why do we have to go through the TypeMap class again? Instead of inlining the expressions? Is it something CompileToMethod requires?

jbogard:
So locally, comparing master and this branch, I'm seeing virtually identical numbers.

TylerCarlson1:
Maybe it has something to do with the Timers changes that you merged in.  I'll take a look at it.

src/AutoMapper/TypeMap.cs
TylerCarlson1: Yes cause inlining the expression will show everything it calls.  But in a lot of the code you translate expressions to invoke functions the user passes to configure it. [Before Map](https://github.com/AutoMapper/AutoMapper/blob/2cb87ad30977f34ddccac9e7bac407893f1626b7/src/AutoMapper/Configuration/MappingExpression.cs#L348) shows this issue cause your expression generated is calling an action that turns into a constant value when you translate it.  And that constant won't be able to be translated into `CompileToMethod().

If you remove CanCompile visitor check when making the delegate and run the tests you can see all the exceptions for yourself.

jbogard:
This PR (I've locally merged master in):
```
AutoMapper: - Flattening - Mapping time:        0.2246414s
Manual: - Flattening - Mapping time:    0.0150164s
CtorMapper: - Ctors - Mapping time:     0.1803494s
ManualCtorMapper: - Ctors - Mapping time:       0.0060393s
Complex Types: - Complex - Mapping time:        2.442033s
Manual Complex Types: - Complex - Mapping time:         0.9180451s
Deep Types: - Deep - Mapping time:      1.3522139s
Manual Deep Types: - Deep - Mapping time:       0.1674475s
```
Master:

```
AutoMapper: - Flattening - Mapping time:        0.2348197s
Manual: - Flattening - Mapping time:    0.0150056s
CtorMapper: - Ctors - Mapping time:     0.1901163s
ManualCtorMapper: - Ctors - Mapping time:       0.0059081s
Complex Types: - Complex - Mapping time:        2.4433725s
Manual Complex Types: - Complex - Mapping time:         0.9320205s
Deep Types: - Deep - Mapping time:      1.3242591s
Manual Deep Types: - Deep - Mapping time:       0.1722072s
```

lbargaoanu:
I see the same. I think performance wise the only benefit you get from compile to assembly is skipping that security check. Profiling would show that.

TylerCarlson1:
I can't even run master's benchmarks on my machine

TylerCarlson1:
Get when try to Run Initialize in FlatteningMapper

An unhandled exception of type 'System.FieldAccessException' occurred in Benchmark.exe

Additional information: Attempt by method 'Benchmark.Flattening.FlatteningMapper.Initialize()' to access field 'Benchmark.Flattening.FlatteningMapper+<>c.<>9__3_0' failed.

lbargaoanu:
I saw that too at a certain point. I thought it was first chance. Anyway, I ran the benchmark :)

TylerCarlson1:
@jbogard can you try removing `[assembly: AllowPartiallyTrustedCallers]` from Benchmark.Program.cs and possibly Automapper.Net4 and see if there's a performance difference?  I seem to get a boost when I run it, but can't confirm master's values on my machine.

TylerCarlson1:
oh hey NOW master works, and I see removing those assembly attributes seems to help execution.  Updating PR now.

TylerCarlson1:
Eh I guess no perf gain anymore, now that everything is being timed correctly, I can't see the difference.

Make sure to add [assembly: AllowPartiallyTrustedCallers] to program in nuget package, or make sure people are aware of it.

jbogard:
Well nuts. Sorry you had to go through all this for us to see our benchmark app was wrong

TylerCarlson1:
Well not ALL of it was because of the benchmark.  Probably last 20% was done after @lbargaoanu put in the benchmark, so I would of saved that.

There's some stuff you could steal here as well.

Making Fuent Expressions, so you can read the TypeMapBuilderPlan. See your Nice comment.
Other thing is the `TypeMap<TSource,TDestination>` which could possibly speed up finding map expression by not going through the dictionary.  (The main bottleneck now is Finding the Expression vs executing it) Maybe 6x times longer finding the expression, than calling it in FlattenMapper, I would guestimate.

PR #1351, opened 2016-06-20
Simplify Resolution Context

jbogard:
I hate the thing. This will be removing contextual types/values and having it only contain the more static stuff like mapping options.


src/AutoMapper/Execution/ReflectionHelper.cs
```diff
@@ -22,7 +22,7 @@ public static object GetDefaultValue(this ParameterInfo parameter)
         public static object MapMember(this ResolutionContext context, MemberInfo member, object value)
         {
             var memberType = member.GetMemberType();
-            return context.Map(value, null, value?.GetType() ?? memberType, memberType);
+            return context.Mapper.Map(value, null, value?.GetType() ?? memberType, memberType, context);
```
lbargaoanu: Here the idea was to reuse the context. This doesn't. If you can make it a struct it won't matter anymore.

lbargaoanu:
So it cannot be a struct because it's mutable. Maybe something like the flyweight pattern to reuse some generic instance and allocate it only when it has something different to store. Or smth :) I see it like static things and rarely used ones, so usually a generic reusable instance.

jbogard:
What would be the issue leaving it as a class? It's now only instantiated once per top-level Map call

lbargaoanu:
Well, nothing really, except that it gets allocated :) Just an idea for further optimization. A small object still counts (header and all).

PR #1353, opened 2016-06-21
[Example] Compiled lambda Performance

TylerCarlson1:
Shows performance difference between compiled expression and actual code.  I can't compile a lamdba to match code since it's multiple lines but it shows the performance hit.  Code run matches expression.

All other things have been eliminated.  Finding function and resolution context generation.

FlattenMapper's compiled map func (No Sub Invokes just static method calls)
```
(src, dest, ctxt) =>
{
    var dest = dest ?? new ModelDto();
    dest.SubSubSubIAmACoolProperty = (((src == null) ? null : src.Sub == null) ? null : src.Sub.SubSub == null)
        ? null
        : src.Sub.SubSub.IAmACoolProperty ?? ((string)ObjectCreator.CreateNonNullValue(typeof(string)));
    dest.SubWithExtraNameProperName = ((src == null) ? null : src.SubWithExtraName == null) ? null : src.SubWithExtraName.ProperName ?? ((string)ObjectCreator.CreateNonNullValue(typeof(string)));
    dest.Sub2ProperName = ((src == null) ? null : src.Sub2 == null) ? null : src.Sub2.ProperName ?? ((string)ObjectCreator.CreateNonNullValue(typeof(string)));
    dest.SubProperName = ((src == null) ? null : src.Sub == null) ? null : src.Sub.ProperName ?? ((string)ObjectCreator.CreateNonNullValue(typeof(string)));
    dest.BaseDate = (src == null) ? default(DateTime) : src.BaseDate;
    return dest;
}
```

Look at FlatteningMappers as benchmarks

I get 10 time slower execution in Release and 2.4 times slower in Debug just from running the execution of the func.

Using CompileToMethod under the same tests the difference is 2%, basically nothing for Debug and 40% more in Release.  (Probably cause dynamic assembly isn't set to optimize code)

### DIFF
--- a/src/AutoMapper/Execution/TypeMapPlanBuilder.cs
+++ b/src/AutoMapper/Execution/TypeMapPlanBuilder.cs
@@ -256,6 +256,7 @@
             ParameterExpression ctxtParam)
         {
             var pmExpression = CreatePropertyMapFunc(pm, configurationProvider, registry, srcParam, destParam, ctxtParam);
+            return pmExpression;
 
             if (pmExpression == null)
                 return null;
--- a/src/AutoMapper/Mapper.cs
+++ b/src/AutoMapper/Mapper.cs
@@ -221,6 +221,18 @@ namespace AutoMapper
 
             return func(source, destination, context);
         }
+        public Func<TSource, TDestination, ResolutionContext, TDestination> Func<TSource, TDestination>(TSource source, out ResolutionContext context)
+        {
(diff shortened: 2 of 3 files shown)


lbargaoanu:
The reason for the performance difference is an extra security check in clr!Security::SecurityCalloutQuickCheck. That happens because the dynamic method is anonymously hosted in a system-provided assembly. And that assembly has a few security attributes set. Apparently there is a workaround. See [this](http://stackoverflow.com/questions/5053032/performance-of-compiled-to-delegate-expression/5160513).

PR #1354, opened 2016-06-21
Inline the lambdas as necessary

jbogard:
This one was big, the collection/complex ones dropped the times nearly in half locally.

### DIFF
--- a/src/AutoMapper/ConstructorParameterMap.cs
+++ b/src/AutoMapper/ConstructorParameterMap.cs
@@ -40,6 +40,7 @@ namespace AutoMapper
 
             if (CustomValueResolver != null)
             {
+                // Invoking a delegate
                 return Invoke(Constant(CustomValueResolver), srcParam, ctxtParam);
             }
 
--- a/src/AutoMapper/Execution/TypeMapPlanBuilder.cs
+++ b/src/AutoMapper/Execution/TypeMapPlanBuilder.cs
@@ -166,7 +166,7 @@
             if (typeMap.Condition != null)
             {
                 mapperFunc =
-                    Condition(Invoke(typeMap.Condition, ctxtParam),
+                    Condition(typeMap.Condition.Body,
                         mapperFunc, Default(typeMap.DestinationType));
                 //mapperFunc = (source, context, destFunc) => _condition(context) ? inner(source, context, destFunc) : default(TDestination);
             }
@@ -234,6 +234,7 @@
             if (typeMap.DestinationType.IsInterface())
             {
                 var ctor = Call(Constant(ObjectCreator.DelegateFactory), typeof(DelegateFactory).GetMethod("CreateCtor", new[] { typeof(Type) }), Call(New(typeof(ProxyGenerator)), typeof(ProxyGenerator).GetMethod("GetProxyType"), Constant(typeMap.DestinationType)));
+                // We're invoking a delegate here
                 return Invoke(ctor);
             }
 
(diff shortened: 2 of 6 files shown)


jbogard:
And ha, with this one it's actually faster than Mapster for the simple Foo->Foo case.

jbogard:
@TylerCarlson1 this will likely make your eyes bleed with the array mapper stuff

TylerCarlson1:
@jbogard Didn't go far enough.
You can do if countMethod to be `source.Count` if source is Collection or `source.Length` if source is Array, then if neither then do `source.Count()`

That way with collections and arrays you'll never have to enumerate it to get the count the first time.

TylerCarlson1:
I think Foo->Foo is faster, and faster than hand code is the fact it's not mapping Foo->Foo in the lists, it's merly setting them to the same object.  Unless your code change ignores assignable mapper somehow

TylerCarlson1:
OK clarification. ToArray from Enumerable doesn't double enumerate.  At least in code base I see.  It actually does it once, the only difference is it has to keep making arrays as it adds items from source enumeration.  The only difference here is you don't have to keep making new array objects and copying them over.

Adding items to a list has that same effect I believe, since there's a base array below the list.

Very minor performance increases, unless you possibly have a list of 10k items.

jbogard:
Eh that wouldn't be that hard to add.

TylerCarlson1:
Up to you.  Either way I think it won't make THAT much difference.  I think we are in Nano Optimizations at this point.

jbogard:
Ah, you're right, it does do the optimization underneath the covers. Maybe the expression building is complicated enough lol.

lbargaoanu:
About Invoke. Check the IL generated from the expressions, that's what matters. It can look quite different from what you have in the expression tree.
About ToArray. It does what is has to do in order not to enumerate twice. In the general case, that might not even be possible, that's the assumption.

TylerCarlson1:
@lbargaoanu It does optimize ToArray under the covers to get best performance.  You can get slightly better with knowing the type to get the count of the enumerable without enumerating it.  That way you can make the array size the correct length and enumerate once.

In ToArray it has to make an array of 4 then keeps doubling it as needed to store new items being enumerated.  This is if it's not an array type, which in our case it will always be since we would be doing Select statement.

Only difference is making array once vs multiple times depending on the size.  But that's assuming you can get the count without enumerating.

lbargaoanu:
I guess I'm saying pretty much the same thing except that here the source is just IEnumerable and you cannot get the count without enumerating. So you end up doing it twice. Or I'm just missing something :)

TylerCarlson1:
Sorry sourceExpression.Type is an IEnumerable, but it's not necessarily of type IEnumerable.  It could be a list or an array or a hash set and the source property type can be set as such.

sourceExpression.Type could be HashSet<int> so therefor we can use ICollection<int>.Count property to get the count without enumerating, or sourceExpression.Type is Array[object] so we can use Array.Length property to get count.

lbargaoanu:
Yes, but if it's just IEnumerable (not ICollection) you're in the bad case. Maybe it's just theory and nobody uses that, but they don't assume that in the BCL. As I said before, maybe it's ok if nobody complains :)

TylerCarlson1:
Ok IEnumerable Use ToArray, everything else use proper count property.  Or better yet just use ToArray for all of it and make it a single line lol.

I concede that ToArray is probably the best and most performance/coding headache value out there.

jbogard:
Right, if it's IEnumerable and not some ICollection, it will be enumerated twice. Arrays implement ICollection, as do most of the built-in generic enumerable types. Dictionaries, lists, arrays, sets, pretty much everything. I could detect ICollection or not but honestly at this point...

jbogard:
@TylerCarlson1 fixed that assignable thing, that looked like a bug

PR #1356, opened 2016-06-23
Try to skip the security check

lbargaoanu:
I got about 40%. [Before](https://cloud.githubusercontent.com/assets/4517428/16297527/da2b7d18-3939-11e6-9197-4dc640baf460.png) and [after](https://cloud.githubusercontent.com/assets/4517428/16297528/da31cbdc-3939-11e6-9e80-b673f8a7ab1b.png).

### DIFF
--- a/src/AutoMapper.Net4/AutoMapper.Net4.csproj
+++ b/src/AutoMapper.Net4/AutoMapper.Net4.csproj
@@ -30,7 +30,7 @@
     <DebugType>pdbonly</DebugType>
     <Optimize>true</Optimize>
     <OutputPath>bin\Release\</OutputPath>
-    <DefineConstants>TRACE;NETSTANDARD1_3</DefineConstants>
+    <DefineConstants>TRACE;NETSTANDARD1_3;NET45</DefineConstants>
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
     <DocumentationFile>bin\Release\AutoMapper.XML</DocumentationFile>
--- a/src/AutoMapper/AssemblyInfo.cs
+++ b/src/AutoMapper/AssemblyInfo.cs
@@ -1,3 +1,9 @@
 ﻿using System;
+using System.Security;
 
 [assembly:CLSCompliant(true)]
+[assembly: AllowPartiallyTrustedCallers]
+[assembly: SecurityTransparent]
+#if NETSTANDARD1_3
+[assembly: SecurityRules(SecurityRuleSet.Level2, SkipVerificationInFullTrust = true)]
+#endif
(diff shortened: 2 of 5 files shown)


lbargaoanu:
It still shows a lot of clr.dll, so maybe I missed something. When I tested at home in a simpler case, the difference was clear, now I'm not so sure :)

lbargaoanu:
The build fails. Apparently I can't get the #if right. I cannot test locally because the .NET Core project doesn't build:

1>C:\Program Files (x86)\MSBuild\Microsoft\VisualStudio\v14.0\DotNet\Microsoft.DotNet.Common.Targets(241,5): error : Object reference not set to an instance of an object.

TylerCarlson1:
Looking at the benchmarks in AppVoyer it looks about the same if not worse :(.
I measure performance by comparing Flattening mapper vs hard coded time taken to execute.
You build has a ratio of 47x where mine is about 13.5x.
Latest accepted PR got about 37x in AppVoyer.
Then I check the other ones, but usually their performance should either all go up or down when you are just changing compile method.

I don't know if it has something to do with the timer change or what.

lbargaoanu:
It's not easy to tell what happens in more detail because symbols for clr.dll are nowhere to be found. The test as it is in the Benchmark doesn't have enough samples for the Flattening mapper. I tested with 10 times as many. The improvement is there to see in the profiler results and the timings for the benchmark. What exactly you did in your PR, I don't know :) This tries to handle just the security check I was talking about. And it does, I'm just not sure it removes it completely because the symbols are missing from the MS servers.

lbargaoanu:
Well, I ran on a different machine where I was able to get the symbols for clr.dll and the check is [nowhere](https://cloud.githubusercontent.com/assets/4517428/16308146/7fca7b8e-396c-11e6-8115-1e651ef27460.png) to be found. Apparently it's the JIT. Reasonably sure this works as expected.

lbargaoanu:
@jbogard @TylerCarlson1 About .NET Core not building. Is it me? :)

TylerCarlson1:
@lbargaoanu I get that sometimes and sometimes it gives me a file and I have to delete the targets or something.  I don't remember the specifics but you have to edit a file that's not part of git.

Also one concern with the PR is because the HighPerformanceTimer stops working, will that have ill effects on other people's code who use it?  Also does that mean they have to manualy set assembly types, or will the nuget do that for them?  Kind of like how Automapper 3.x had to add Automapper.Net4 to the solution's source files.

lbargaoanu:
Thanks, I guess I'll see :) Of course they would be in the same situation. That's why I was saying that this must be a bug at some level. I'm working around something that should work as is. The system assemblies have two of the attributes, but they are marked SecurityCritical which makes sense for them but not for AutoMapper I would say. Maybe there is a better mix :)

PR #1361, opened 2016-06-24
Map just what's required; fixes #1342

### LINKED ISSUE #1342: Map Automatic Property  to ExpandoObject causes many k__BackingField member
Hi.
I am using AutoMapper 4.2.1 in asp.net core 1.0.0 rc2 project.
I try to map object with Automatic Property to dynamic object, and serialize it to Json. This cause k__BackingField  in the Json.

My code is like this:
```csharp
class User
{
}

User user = new User { UserName = "hello" };
dynamic d = Mapper.Map<ExpandoObject>(user);
string json = Json((ExpandoObject)d);
```
The result json like this
```json
{"UserName":"hello","<UserName>k__BackingField":"hello"}
```

If I just serialize the user object, it is OK.
```csharp
User user = new User { UserName = "hello" };
string json = Json(user);
```
result
```json
{"UserName":"hello"}
```
I try to use mapperconfiguration, but it doesn't help,even when I set false to those filter.
```csharp
User user = new User { UserName = "hello" };
Mapper.Initialize(cfg =>
{
    cfg.ShouldMapProperty = p => false;
    cfg.ShouldMapField = f => false;
});
dynamic d = Mapper.Map<ExpandoObject>(user);
string json = Json((ExpandoObject)d);
```
The result is still
```json
{"UserName":"hello","<UserName>k__BackingField":"hello"}
```
If I change the automatic property  to field, everything is well, but I think it's not  a good solution.
I don't know why the filter is not working. How should I do to remove  the k__BackingField members?
Thanks.

lbargaoanu:


src/AutoMapper/Mappers/StringDictionaryMapper.cs
```diff
@@ -55,8 +55,8 @@ public bool IsMatch(TypePair context)
 
         private static TDestination Map<TDestination>(StringDictionary source, TDestination destination, ResolutionContext context)
         {
-            destination = destination == null ? context.Mapper.CreateObject<TDestination>() : destination;
-            var destTypeDetails = new TypeDetails(typeof(TDestination), _ => true, _ => true);
+            destination = destination == null ? context.Mapper.CreateObject<TDestination>() : destination;
+            var destTypeDetails = new TypeDetails(typeof(TDestination), context.ConfigurationProvider.ConfigurationExpression);
```
jbogard: Why can't we just use context.ConfigurationProvider here? I was hesitant to expose the configuration expression publicly, since that would be effectively mutating config.

lbargaoanu: Those interfaces didn't seem to match. I know you refactored them, but still, maybe you should take a second look :)

lbargaoanu: Of course I can try it too, I just took the easy way out.

lbargaoanu:
Unrelated, since many things changed in terms of public members, maybe we should try to replace IMappingExpression with IMappingExpression< object, object > so we can avoid the forwarding in MappingExpression.

jbogard:
@lbargaoanu My original thought was that if we ever did conventions, where we apply a mapping expression globally based on some convention, it would make more sense with that non-generic interface. Maybe not.

lbargaoanu:
I think it's more about syntax then anything. If the type parameters are both object, it's pretty much non generic :) It's not a lot of forwarding code, but it's a little annoying and we can easily forget to add the non generic corresponding method.

lbargaoanu:
Unless you have different thoughts about the interface thing, this is done.

jbogard:
Yeah I wanted to download it and take a closer look but ran out of time.

lbargaoanu:
Updated.

PR #1365, opened 2016-06-26
Use non prerelease versions of System-packages

Tasteful:
Use non prerelease versions of System-packages from https://www.myget.org/gallery/aspnetrelease

### DIFF
--- a/src/AutoMapper/project.json
+++ b/src/AutoMapper/project.json
@@ -17,25 +17,12 @@
   "frameworks": {
     "netstandard1.3": {
       "dependencies": {
-        "Microsoft.CSharp": "4.0.1-rc2-24027",
-        "NETStandard.Library": "1.5.0-rc2-24027",
-        "System.Linq.Expressions": "4.0.11-rc2-24027",
-        "System.Linq.Queryable": "4.0.1-rc2-24027",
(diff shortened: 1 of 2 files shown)


lbargaoanu:
This doesn't apply to netstandard1.1?

Tasteful:
@lbargaoanu Correct that it's not support netstandard 1.1. The dependency System.Collections.Specialized is only supporting the following TFM
```
.NETFramework 4.6: 4.6.0.0
.NETStandard 1.3: 1.3.0.0
MonoAndroid 1.0: 1.0.0.0
MonoTouch 1.0: 1.0.0.0
Xamarin.iOS 1.0: 1.0.0.0
Xamarin.Mac 2.0: 2.0.0.0
Xamarin.TVOS 1.0: 1.0.0.0
Xamarin.WatchOS 1.0: 1.0.0.0
```
https://www.myget.org/feed/aspnetrelease/package/nuget/System.Collections.Specialized/4.0.1
I'm not sure if this is the final-final package of the System.Collections.Specialized.

Tasteful:
@lbargaoanu It was I that read wrong. the System.Collections.Specialized was not used in netstandard1.1 so I added that back. (copy/paste misstake)

jbogard:
Thanks! RTM packages dropped today, so I just targeted those. Thanks for the PR!

PR #1373, opened 2016-06-27
Map from nullable source; fixes #1372

### LINKED ISSUE #1372: Value lost when mapping from Nullable<int> to Nullable<decimal>
When mapping from `int?` to `decimal?`, I get NULL even when the source is not null, 
as shown in the example below. This is using Automapper 5.0-beta-1. 

`public class C1{public int? X { get; set; }}`
`public class C2{public decimal? X { get; set; }}`
....
`var config = new MapperConfiguration(cfg => cfg.CreateMissingTypeMaps = true);`
`var mapper = config.CreateMapper();`
`var c2 = mapper.Map<C2>(new C1 {X = 99});`
`Assert.That(c2.X == 99); // fails, because c2.X = null`

I also tried the same in Automapper 4.2.1 but that resulted in a AutoMapperMappingException.

Also, the problem only occurs when also the source property is nullable.

lbargaoanu:


src/AutoMapper/project.json
```diff
@@ -14,34 +14,37 @@
   "buildOptions": {
     "keyFile": "..\\AutoMapper.snk"
   },
-  "frameworks": {
```
jbogard: Can you revert the project.json changes? I'm doing a PR to support RTM.

PR #1375, opened 2016-06-27
Remove the nullable check and avoid boxing

lbargaoanu:
The nullable check is not required anymore or we need more tests :)


jbogard:
Well that's odd. But nullable types are just odd anyway. It could be the stuff from doing mapper selection up front instead of runtime types, which nullable lies about.

PR #1376, opened 2016-06-27
Targeting preview 2 sdk

jbogard:


src/UnitTests/Bug/AssignableCollectionBug.cs
```diff
@@ -38,7 +38,7 @@ protected override void Because_of()
             [Fact]
             public void Should_copy_array()
             {
-                _dest.ImageData.ShouldBeSameAs(_source.ImageData);
+                _dest.ImageData.ShouldEqual(_source.ImageData);
```
lbargaoanu: This seems wrong. Why copy? It could be 1MB.

jbogard: Array lengths are immutable, but their contents are not. The original
version of this did an array copy, I don't know why I copied the reference
over. This makes it consistent with all we other enumerables.

lbargaoanu: Primitive types array are different I would say, they're just data. I wouldn't expect them to be copied.

PR #1400, opened 2016-06-29
Fixes #1380

### LINKED ISSUE #1380: Additional Type Checks still run after ConvertUsing in 5.0
When registering a custom type converter, I would expect no additional checks to be performed when validating the configuration.  Instead, an exception is thrown if a type exists that cannot be automatically converted:

	public class A
	{
	}
	public class B
	{
		public C Foo { get; set; }
	}
	public class C { }

	[TestMethod]
	public void ConvertUsing_IgnoresProperties()
	{
		{
			cfg.CreateMap<A, B>().ConvertUsing(x => new B { Foo = new C() });
		});
	}

This generates the following error:

> AutoMapper.AutoMapperConfigurationException: The following property on UnitTestProject8.C cannot be mapped: 
> 	
> Add a custom mapping expression, ignore, add a custom resolver, or modify the destination type UnitTestProject8.C.
> Context:
> 	Mapping from type System.String to UnitTestProject8.C

Observations:
* The error only occurs on properties which exist in TSrc and TDest
* The error only occurs if there is no mapping between the property types
* No exception is generated if trying to map using this configuration, only on AssertConfigurationIsValid

lbargaoanu:
Whitespace in UnitTests/ConfigurationValidation.cs. The tests are at the top. I was not sure if there are other similar cases.

### DIFF
--- a/src/AutoMapper/ConfigurationValidator.cs
+++ b/src/AutoMapper/ConfigurationValidator.cs
@@ -62,6 +62,10 @@ namespace AutoMapper
             if (typeMap != null)
             {
                 typeMapsChecked.Add(typeMap);
+                if(typeMap.CustomMapper != null || typeMap.TypeConverterType != null)
+                {
+                    return;
+                }
                 CheckPropertyMaps(typeMapsChecked, typeMap, context);
             }
             else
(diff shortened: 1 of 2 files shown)


jbogard:
After 5.0.1 drops I'm making a whitespace normalization commit, removing all the tabs with spaces.

PR #1401, opened 2016-06-29
IfNullVisitor memory leak

lbargaoanu:
the static visitor is a memory leak
replace the list with a HashSet, just in case there are many nodes


jbogard:
Ugh, good catch.

PR #1407, opened 2016-06-30
[WIP] Fix #1352

### LINKED ISSUE #1352: UseAsDataSource StackOverflow Error
This is giving me a stackoverflow error, but only when I have a collection navigation property.

`return dbSet.UseAsDataSource(mapper).For<TDTO>().AsExpandable().Where(filter).ToList();`

The filter is of type `Expression<Func<TDTO, bool>>`
Possible filter values that this happens with: 
```

filter = {f => (True AndAlso Invoke(f => (True AndAlso Invoke(Param_0 => Convert(Param_0.Cust.CustomerNumber).ToString().Contains("123"), f)), f))} 

filter = {f => (True AndAlso Invoke(f => ((((False OrElse Invoke(Param_0 => Convert(Param_0.Name1).ToString().Contains("som"), f)) OrElse Invoke(Param_0 => Convert(Param_0.Name2).ToString().Contains("som"), f)) OrElse Invoke(Param_0 => Convert(Param_0.PaytoName1).ToString().Contains("som"), f)) OrElse Invoke(Param_0 => Convert(Param_0.PaytoName2).ToString().Contains("som"), f)), f))}

```

Right now my mappings are just:

```
cfg.CreateMap<Cust, CustDTO>().ReverseMap();
cfg.CreateMap<Customer, CustomerDTO>().ReverseMap();
```

 
Previously I could get away with this because I didn't need the collection.

But now I have added another entity, which requires me to have that collection available in Cust. This Customers property in Cust is just an example of what is causing it.

 
Here are the relevant properties in the entities:

**namespace-Domain:**
 ```
public class Cust : IEntity   {
 
       [Key]
       public int CustomerID { get; set; }

       public string CustomerNumber { get; set; }
       public bool Status { get; set; } 
       public virtual ICollection<Customer> Customers { get; set; }
       } 

  public class Customer : IEntity
   {
       [Key]
       public int Id { get; set; }
 
       [ForeignKey("Cust")]
       public int CustomerId { get; set; }
       public virtual Cust Cust { get; set; }
       public bool Status { get; set; }
       public string Name1 { get; set; }
}
```
**namespace-DTO:**
 ```
public class CustDTO : IDTO
    {
        public int CustomerID { get; set; }
        public string CustomerNumber { get; set; }
        public bool Status { get; set; }

        public virtual ICollection<CustomerDTO> Customers { get; set; }
    }

public class CustomerDTO : IDTO
    {
        public int Id { get; set; }

        public int CustomerId { get; set; }
        public virtual CustDTO Cust { get; set; }
        public bool Status { get; set; }
        public string Name1 { get; set; }
}

```

Another note:
I'm not sure if this is relevant, but If I remove `.AsExpandable()` the error changes to (depending on the filter value stated above):

```
 'System.ArgumentException'
Property 'namespace-Domain.Cust Cust' is not defined for type 'namespace-DTO.CustomerDTO'
```
**or**
```
 'System.ArgumentException'
Property 'System.String Name1' is not defined for type 'namespace-DTO.CustomerDTO'
```
I wrote this a while ago, and am a little fuzzy on why I needed AsExpandable(), but it seems it was required to successfully map from Entity -> DTO... as pointed out by the incorrect namespace on the Cust object in the error.

TylerCarlson1:
At one point in 5.0 development you set AllowNullDestinationValues = true by default.
That caused the null reference exception in #1352.  It seems to fix it by taking that out and allowing the default constructor to work.  As far as I can tell this check needs to go, or you need to tell people to set AllowNullDestinationValues to true if using MaxDepth and Projection.

This is WIP because I don't fully know the effects of this change.  I don't know if you should return a default of return type instead instead of defaulting to a newing up of an object.  Either one is better than exception.

### DIFF
--- a/src/AutoMapper/QueryableExtensions/ExpressionBuilder.cs
+++ b/src/AutoMapper/QueryableExtensions/ExpressionBuilder.cs
@@ -97,10 +97,7 @@ namespace AutoMapper.QueryableExtensions
             var visitCount = typePairCount.AddOrUpdate(request, 0, (tp, i) => i + 1);
             if (typeMap.MaxDepth > 0 && visitCount >= typeMap.MaxDepth)
             {
-                if (_configurationProvider.AllowNullDestinationValues)
-                {
-                    return null;
-                }
+                // Don't null it cause problems
             }
             else
             {


TylerCarlson1:
@lbargaoanu Last edited ignoring property maps.  Maybe `AllowNullDestinationValues = false;` needs to happen if MaxDepth is used or default to false again.  It was at one point.

Issue #1414, opened 2016-06-30
Value Modifiers proposal


jbogard:
In the 4.x timeframe, I removed a feature called "Formatters". Formatters were a way for you to modify the value of a string, and they stacked on top of each other.

We were using it on a project and it was a horrible idea, BUT I think the general idea might be valid. Type converters globally apply to everything to go from Type A to B, and value resolvers work with individual properties, but nothing really allows you to selectively apply a modifier between resolution and assignment.

Our mapping today works like:

`source.Member = Resolve(dest)`

Where the resolving could be a value resolver, a MapFrom, a member chain, hardcoded value etc.

But there's nothing that lets you take the value from resolving and apply any transforms.

Maybe this is called a transform?

Anyway, the idea is similar to formatters. I killed them originally because they were horribly slow (now could be fast), and our team was horrrrrribly abusing them.

Transformers are like type converters, but can be applied to non-type-map contexts:
- Members
- Profiles
- Configuration

And at the Configuration level, we can apply conditions to them.

```
cfg.CreateMap<Source, Dest>()
    .ForMember(dest => dest.Name, opt => opt.Modify(srcVal => srcVal + " is handsome"));
cfg.CreateProfile("Foo", p => {
   p.Modify<string, string>(srcVal => srcVal + " is VERY handsome");
   p.CreateMap<InnerSource, InnerDest>();
});
cfg.Modify<string, string>(srcVal => srcVal + ". For realz");
cfg.Modify<string, string>(srcVal => srcVal, pm => pm.SourceMember.ContainsAttribute(typeof(SkipJunkAttribute)));
```

Modifiers would be applied first at the member, then profile, then configuration level. You can't opt-out, you'd need to segregate by profiles or add a condition.

Naming?
- Modifiers
- Transformers

TylerCarlson1:
I would think maybe what you want to do is make this based on `ExpressionVisitor` that only apply to the Profiles and nothing else.  I guess you could modify on Members, and configuration is just the default profile.

Then you can make the API use this to take `cfg.Modify<T1,T2>(Expression<Func<T1, T2, T2>>)` and make an `ExpressionAssignmentVisitor` that looks for those types in an assignable expression and then ReplaceParameters of assign from destination.  This example needs some tweaking in what you want to do but the idea is there.
Also for type map modifies this could use the same based code except check the base class the properties are coming from and see if they match too.  I don't like this idea because they can just use MapFrom to get the same result.

[...]

Issue #1390, opened 2016-06-28
ResolutionContext.Parent is missing!


mobius7777:
We have upgraded from Automapper 4.2.1 to 5.0 to get some of the bug fixes, as one of the bugs (1373 or 1372, I think) was biting us. So we have to upgrade to get this fix. Alas, there are too many changes! 

The big issue for us appears to be that the signature of ITypeConverter has radically changed and ResolutionContext is now missing Parent. 

We have a converter that converts from a DateTime to a DateTime, which does so to automatically adjust the datetime up or down for the user's TimeZone. That way, all DateTimes from the domain and converted to the user's TZ and all DateTimes originating from a non-domain ViewModel are coverted back to server time. 

Our 4.2.1 converter looked like this:

```
public class TimeZoneConverter : ITypeConverter<DateTime, DateTime>
    {
        public DateTime Convert(ResolutionContext source)
        {
            var sourceValue = (DateTime)source.SourceValue;
            string timezone = ApplicationCache.TimeZone;

            if (string.IsNullOrEmpty(timezone))
                return sourceValue;

            bool convertFrom = (source.Parent.DestinationType.Namespace.Contains("Domain") || ((source.Parent.Parent != null) && source.Parent.Parent.DestinationType.Namespace.Contains("Domain")));
            if (convertFrom)
            {
                //Convert the date from the user's time zone to the server/db time to copy back to domain
                return sourceValue.ConvertFromSpecifiedTimeZone(timezone);
            }
            else
            {
                //Convert the date to the specified timezone when returning from the domain
                return sourceValue.ConvertToSpecifiedTimeZone(timezone);
            }
        }
    }
```

I have manged to make the changes for 5.0 except for the missing Parent property. Here's what I have so far:

```
    public class TimeZoneConverter : ITypeConverter<DateTime, DateTime>
    {
        public DateTime Convert(DateTime source, ResolutionContext context)
        {
            string timezone = ApplicationCache.TimeZone;

            if (string.IsNullOrEmpty(timezone))
                return source;

            bool convertFrom = context.Parent.DestinationType.Namespace.Contains("Domain") || ((context.Parent.Parent != null) && context.Parent.Parent.DestinationType.Namespace.Contains("Domain")));
            if (convertFrom)
            {
                //Convert the date from the user's time zone to the server/db time to copy back to domain
                return source.ConvertFromSpecifiedTimeZone(timezone);
            }
            else
            {
[...]

PR #1420, opened 2016-07-01
[WIP] Transforms property maps based on a profile

TylerCarlson1:
Possible solution for #1414

Didn't implement version for ForMember because well, MapFrom does the same thing.  Also there's no condition, but I could possibly put one in.  It's kind of hard because it's expression based, but it's possible.

Also there's only one type, the destinationType, so if you use a string converter it will wrap around anything set to string, even if the source type is different.  Also since it's the last thing, it will wrap all possible scenarios of property map, without effecting how they work.

One BIG Issue is making sure these expression visitors only work where they are intended and not break functionality IObjectMapper expressions that are generated.

And later you can make up more functions to do even crazier things.  Even allow people to make quick hacks for issues without having to build a new release that would fix people's problems.
Like have the null checks actually not set the property if any of them are null instead of setting the property to null.  Or taking out the null checks completely if you want the exception to happen or you need the performance boost.

### DIFF
--- a/src/AutoMapper/IProfileExpression.cs
+++ b/src/AutoMapper/IProfileExpression.cs
@@ -1,3 +1,5 @@
+using System.Linq.Expressions;
+
 namespace AutoMapper
 {
     using System;
@@ -51,6 +53,14 @@ namespace AutoMapper
         /// <param name="memberList">Member list to validate</param>
         /// <returns>Mapping expression for more configuration options</returns>
         IMappingExpression CreateMap(Type sourceType, Type destinationType, MemberList memberList);
+
+        /// <summary>
+        /// Transforms PropertyMap with source and destination Type
+        /// </summary>
+        /// <typeparam name="TSource"></typeparam>
+        /// <typeparam name="TDestination"></typeparam>
+        /// <param name="transformExpression"></param>
+        void TransformPropertyMap<TSource>(Expression<Func<TSource, TSource>> transformExpression);
 
         /// <summary>
         /// Clear the list of recognized prefixes.
--- a/src/AutoMapper/IProfileExpression.cs
+++ b/src/AutoMapper/IProfileExpression.cs
@@ -63,6 +63,14 @@ namespace AutoMapper
         void TransformPropertyMap<TSource>(Expression<Func<TSource, TSource>> transformExpression);
 
         /// <summary>
+        /// Transforms PropertyMap with source and destination Type
+        /// </summary>
+        /// <typeparam name="TSource"></typeparam>
+        /// <typeparam name="TDestination"></typeparam>
+        /// <param name="transformExpression"></param>
+        void TransformPropertyMap<TSource>(Expression<Func<TSource, TSource>> transformExpression, Func<PropertyMap, bool> predicateFunc);
+
+        /// <summary>
         /// Clear the list of recognized prefixes.
         /// </summary>
         void ClearPrefixes();
--- a/src/AutoMapper/Profile.cs
+++ b/src/AutoMapper/Profile.cs
@@ -140,7 +140,12 @@ namespace AutoMapper
 
         public void TransformPropertyMap<TSource>(Expression<Func<TSource, TSource>> transformExpression)
         {
-            AfterBuildMapFuncVisitors.Add(new ReplaceAssignExpressionVisitor<TSource, TSource>(transformExpression));
+            AfterBuildMapFuncVisitors.Add(new ForPropertyMapExpressionVisitor(new ReplaceAssignExpressionVisitor<TSource, TSource>(transformExpression)));
+        }
+
+        public void TransformPropertyMap<TSource>(Expression<Func<TSource, TSource>> transformExpression, Func<PropertyMap, bool> predicateFunc)
+        {
+            AfterBuildMapFuncVisitors.Add(new ForPropertyMapExpressionVisitor(new ReplaceAssignExpressionVisitor<TSource, TSource>(transformExpression), predicateFunc));
         }
 
         private IMappingExpression<TSource, TDestination> CreateMappingExpression<TSource, TDestination>(MemberList memberList)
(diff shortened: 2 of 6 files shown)


src/AutoMapper/IProfileExpression.cs
```diff
@@ -53,6 +55,14 @@ public interface IProfileExpression
         IMappingExpression CreateMap(Type sourceType, Type destinationType, MemberList memberList);
 
         /// <summary>
+        /// Transforms PropertyMap with source and destination Type
+        /// </summary>
+        /// <typeparam name="TSource"></typeparam>
+        /// <typeparam name="TDestination"></typeparam>
+        /// <param name="transformExpression"></param>
+        void TransformPropertyMap<TSource>(Expression<Func<TSource, TSource>> transformExpression);
```
jbogard: Does this need to be an expression or can we do the same trick with conditions?

TylerCarlson1: I use `ReplaceParameters` so it has to be an expression.  You should be able to use `?.` or `??` or `? :` in the expressions.

I guess with conditions an example of what you want would be in order.  With more expression trickery I could get the PropertyMap but you already have code for that.

TylerCarlson1: Oh if you want the IfThen PR then no this can't do it I think.  What you would do is `ForAllPropertyMaps` for the condition to map it.  If you want a condition to make the modification, that needs to be another parameter passed over.  It's just a question of what you want passed over.  Want pre set rules based off source type, or want to pass something like PropertyMap, or MemberInfo over, so that they can make any condition they like?

src/AutoMapper/Profile.cs
```diff
@@ -136,6 +138,16 @@ public IMappingExpression CreateMap(Type sourceType, Type destinationType, Membe
             return map;
         }
 
+        public void TransformPropertyMap<TSource>(Expression<Func<TSource, TSource>> transformExpression)
```
lbargaoanu: This should simply call the overload below.

lbargaoanu:
Having both this and #1405 seems confusing.

TylerCarlson1:
@lbargaoanu Names can change.  Unfortunately I don't have a better name for this.

Also I made this because #1405 couldn't solve the problem #1390 had, where as this can.
Maybe call it AppendPropertyMapCode, AppendPropertyMapExpresison?  Since it's not modifying PropertyMap, it's modifying the final expression to get the value from source to set to the destination.

lbargaoanu:
It's not about the name. They both do something for each property map, with a filter. I would say everything could be done with #1405. You put in a ResolveUsing and off you go :) This one is more precise I guess, as you said. Like MapFrom after we have the destination value? Maybe merging the two or something. And anyway you need to have a simple usage story.

TylerCarlson1:
I think what #1390 said was resolve using wouldn't work. [This](https://github.com/AutoMapper/AutoMapper/issues/1390#issuecomment-229689839) says otherwise, because in #1405 you are modifying the PropertyMap, which doesn't always work depending on how the property map was configured.  If it was AutoGenerated, CustomExpression, ResolveUsing, or MapFrom it wouldn't all have a single implementation to work for all cases.

This one does because it doesn't edit PropertyMap.  It lets the PropertyMap do own thing then it modifies the final expression generated, and it shouldn't matter how it was made because you are no longer dealing with the PropertyMap object at this point.

lbargaoanu:
I see. Anyway, subtle :) Still confusing. AfterPropertyMap I guess. So maybe add this to IMemberConfigurationExpression.

TylerCarlson1:
Actually I think maybe AfterPropertiesMapped might be better so it can stay where it is.  Putting it in IMemberConfigurationExpression causes other issues, because at that point it's a ForMember level and this is meant to be at a Profile/ possibly TypeMap level as suggested in #1414.

Either way it's still confusing.  Still looking for the least-worst name for this.  :)

lbargaoanu:
Where it is is part of the problem :) (overlapping with #1405) Where it applies is flexible, defined by the filter. Per Profile is kind of clumsy, but obvious I guess. If it's useful, merging with #1405 through IMemberConfigurationExpression seems best to me.

TylerCarlson1:
Yea that could work too.  Although I don't like it because even though you can put it in ForAllPropertyMaps, you can also put it in ForMember. as well.  I don't think this should be available to ForMember as there already enough things you can do there that can give you the same functionality.

Also I have to change the logic because it's profile based currently, but that's not to bad.  And changing it would be less fragile to breaking if you change how property maps work.  The only thing you lose is flexibility to do more things, without having to change AutoMapper code.  It's a little hacky but if someone wanted it they could do it without having to wait for a new next release.

After MapExpression visitors are BRAND new, so anywhere we implement them is going to cause confusion in some form or another.  I guess whatever way they are going to go is fine for now.

Disclaimer:  I have a use case where I could use this implementation, similar to #1390.  I translate DateTime time zones in another place.  But I have no use currently for #1405.

@mobius7777 & @jbogard Any ideas on if the Modify logic should go in Profile or PropertyMap level?

lbargaoanu:
Being able to have things per profile is helpful if you already use them, like per module. But the point remains.

jbogard:
I like the idea of contextual type converters the best right now, though it
would be a breaking change. I wonder, could we add a method that isn't
CreateMap.ConvertUsing? Something top level?  RegisterConverter or
AddConverter etc.

TylerCarlson1:
That's completely doable.  The real question is do you REALLY want to break ITypeConverter in order to add conditions back into them.  Also is the conditions exclusively to PropertyMaps or does there need to be TypeMap conditions as well?  like if you map at the top level.

I would think RegisterConverter would be best name.  I don't know if you would put the conditions into ITypeConverter or have them pass over when you register.

Also If that were the case I wouldn't bother with the Expression Visitor stuff and just call Convert function, given it's not THAT big of a performance loss by calling a function.

TylerCarlson1:
Maybe that's best to redo ITypeConverter to put conditions back in.

At that point this PR isn't needed anymore.  Which is fine.

lbargaoanu:
Contextual type converters could be implemented with smth like I suggested that was not Expression based. You just put in the right filter.

PR #1429, opened 2016-07-01
Create missing map circular reference

### LINKED ISSUE #1422: StackOverflowException related to CreateMissingTypeMaps and AddConditionalObjectMapper
This issue was discussed in the #1398 but here is some background:

_I just upgraded a large app from 3.3.1 that has few thousands types, scattered across modules etc. In 3.3.1 I had lazy CreateMap custom functionality that was automatically creating maps for me. After the upgrade I had to remove that auto create map functionality and now I don't have many CreateMap calls. I tried to use CreateMissingTypeMaps and I get StackOverflowException, I have created configuration and all profiles are loaded, when I hit first Map I get the StackOverflowException._

Here is the [gist](https://gist.github.com/khorvat/70007d92dad3abd89420835a23ee7bdb) for this issue. In both cases I get StackOverflowException while this is working in 4.2.1 (I have updated the gist with the _PreserveReferences_ proposal)

I have tried to use @TylerCarlson1 [suggestion](https://github.com/AutoMapper/AutoMapper/issues/1398#issuecomment-229840694) and set _PreserveReferences_ in _cfg.ForAllMaps()_ and it's still returning _StackOverflowException_.

@TylerCarlson1 proposed a feature request to modify TypeMap after it's generated by AddConditionalObjectMapper/CreateMissingTypeMaps so could this be an option ?

TylerCarlson1:
Fixes #1422

### DIFF
--- a/src/AutoMapper/MapperConfiguration.cs
+++ b/src/AutoMapper/MapperConfiguration.cs
@@ -114,10 +114,10 @@ namespace AutoMapper
 
             _typeMapPlanCache.AddOrUpdate(typePair, typeMap, (tp, tm) =>
             {
-                if (tm?.Sealed == false)
-                    tm.Seal(_typeMapRegistry, this);
-                else if (typeMap?.Sealed != true)
-                    Task.Delay(1).Wait();
+                if (tm != null)
+                    lock (tm)
+                        if (tm.Sealed == false)
+                            tm.Seal(_typeMapRegistry, this);
                 return tm;
             });
 
--- a/src/AutoMapper/MapperConfiguration.cs
+++ b/src/AutoMapper/MapperConfiguration.cs
@@ -1,3 +1,5 @@
+using System.Threading;
+
 namespace AutoMapper
 {
     using System;
@@ -108,6 +110,16 @@ namespace AutoMapper
         public TypeMap ResolveTypeMap(TypePair typePair)
         {
             var typeMap = _typeMapPlanCache.GetOrAdd(typePair, _getTypeMap);
+
+            _typeMapPlanCache.AddOrUpdate(typePair, typeMap, (tp, tm) =>
+            {
+                if (tm?.Sealed == false)
+                    tm.Seal(_typeMapRegistry, this);
+                else if (typeMap?.Sealed != true)
+                    Thread.Sleep(1);
+                return tm;
+            });
+
             return typeMap;
         }
 
@@ -272,8 +284,6 @@ namespace AutoMapper
                 .Select(p => p.ConfigureConventionTypeMap(_typeMapRegistry, typePair))
                 .FirstOrDefault(t => t != null);
 
-            typeMap?.Seal(_typeMapRegistry, this);
-
             return typeMap;
         }
 
@@ -286,8 +296,6 @@ namespace AutoMapper
                 .Cast<IProfileConfiguration>()
                 .Select(p => p.ConfigureClosedGenericTypeMap(_typeMapRegistry, typePair, requestedTypes))
                 .FirstOrDefault(t => t != null);
-
-            typeMap?.Seal(_typeMapRegistry, this);
 
             return typeMap;
         }
--- a/src/AutoMapper/MapperConfiguration.cs
+++ b/src/AutoMapper/MapperConfiguration.cs
@@ -1,4 +1,5 @@
 using System.Threading;
+using System.Threading.Tasks;
 
 namespace AutoMapper
 {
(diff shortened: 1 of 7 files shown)


src/AutoMapper/MapperConfiguration.cs
```diff
@@ -108,6 +110,16 @@ public TypeMap ResolveTypeMap(Type sourceType, Type destinationType)
         public TypeMap ResolveTypeMap(TypePair typePair)
         {
             var typeMap = _typeMapPlanCache.GetOrAdd(typePair, _getTypeMap);
+
+            _typeMapPlanCache.AddOrUpdate(typePair, typeMap, (tp, tm) =>
+            {
+                if (tm?.Sealed == false)
+                    tm.Seal(_typeMapRegistry, this);
+                else if (typeMap?.Sealed != true)
+                    Thread.Sleep(1);
```
TylerCarlson1: This is needed to pass DynamicMapThreadingIssues.Should_Not_Fail test.
Not best solution to fix multi-threaded problem, but works.

Unless someone tries to go farther then the test does I don't see it as a big issue.

src/AutoMapper/MapperConfiguration.cs
```diff
@@ -116,7 +117,7 @@ public TypeMap ResolveTypeMap(TypePair typePair)
                 if (tm?.Sealed == false)
                     tm.Seal(_typeMapRegistry, this);
                 else if (typeMap?.Sealed != true)
-                    Thread.Sleep(1);
+                    Task.Delay(1).Wait();
```
TylerCarlson1: Didn't want to put another reference and this works with .NET standard as is.

src/AutoMapper/MapperConfiguration.cs
```diff
@@ -108,6 +111,16 @@ public TypeMap ResolveTypeMap(Type sourceType, Type destinationType)
         public TypeMap ResolveTypeMap(TypePair typePair)
         {
             var typeMap = _typeMapPlanCache.GetOrAdd(typePair, _getTypeMap);
+
+            _typeMapPlanCache.AddOrUpdate(typePair, typeMap, (tp, tm) =>
+            {
+                if (tm?.Sealed == false)
+                    tm.Seal(_typeMapRegistry, this);
+                else if (typeMap?.Sealed != true)
+                    Task.Delay(1).Wait();
```
jbogard: wat

TylerCarlson1: I can't seal the newly generated type maps as they are being created with a circular loop.  It tries to resolve the other type map which then resolves itself and over and over.  This resolves the type map, and then seals it after it's added to cache like regular TypeMaps are.  And to break the inifinite seal loop, I have to know it's in the process of being sealed and not try to re-seal it.

The delay it to make the dynamic mapping multi-threaded test pass

src/AutoMapper/MapperConfiguration.cs
lbargaoanu: This is just a hack :)

src/AutoMapper/MapperConfiguration.cs
lbargaoanu: That's what locking is for.

lbargaoanu:
@jbogard Maybe all this CreateMissingTypeMaps stuff is not worth the trouble. It made more sense back when you were creating maps on the fly. But now? Maybe remove it or try to support the simplest of cases. It's clear to me that people will abuse it no matter the guidance.

jbogard:
I think you're right. It was originally for types only known at runtime,
but honestly, exactly how often does that come up?

TylerCarlson1:
Conventions will still have this problem.  Doesn't matter if it's CreateMissingMaps, AddTypeConvention, OR GenericTypeMaps (Yes they will have the same issue).  The real problem is people have circular references and as you have said many times, the best way to deal with circular references is to get rid of them.  MaxDepth is there but never 100% works on everything AutoMapper does and seems to break with each iteration.

jbogard:
Can we do this without the task delay thing? Maybe locking the typemap during seal?

TylerCarlson1:
Maybe that might work, maybe not.  It seems weird to lock a resource that's already part of a thread safe operation that locks it out as well.  I know waiting for seal to equal true before continuing causes some tests to never finish.

lbargaoanu:
From MSDN
If you call GetOrAdd simultaneously on different threads, addValueFactory may be called multiple times, but its key/value pair might not be added to the dictionary for every call.

jbogard:
The locking operation on the collection only locks the collection, not the typemap itself. Another way to approach this would be to find what the deadlock comes from, and introduce some sort of soft lock (detecting a circular ref for example).

TylerCarlson1:
Yea I guess locking works.

Much better now.

PR #1430, opened 2016-07-01
Expando object bug

### LINKED ISSUE #1418: Can't map between List<ExpandoObject>s
When I try to create a map in 5.0.0 between two objects which contain properties of type List<ExpandoObject>, I receive an IndexOutOfBounds exception on the mapper Initialize: Index was outside the bounds of the array.

If I change the properties to dynamic, no exception is received.

TylerCarlson1:
Was counting ExpandoObject as dictionary type and using it as DictionaryMapper

Fixes #1418

### DIFF
--- a/src/AutoMapper/Configuration/PrimitiveExtensions.cs
+++ b/src/AutoMapper/Configuration/PrimitiveExtensions.cs
@@ -1,3 +1,5 @@
+using System.Dynamic;
+
 namespace AutoMapper.Configuration
 {
     using System;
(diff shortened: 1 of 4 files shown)


lbargaoanu:
So if you move the string dictionary mappers before the dictionary mapper you don't get the same effect? Seems more elegant.

TylerCarlson1:
Tried that first and like 16 tests failed.  Didn't look at it too closely why they failed, I just went for the solution as it is now to make it work.

PR #1434, opened 2016-07-04
Fixes #1417

### LINKED ISSUE #1417: Not all members are being mapped
After upgrading to 5.0.0 I have one model class (I've only tested this class so far) that has 4 members more than its dto class. Some members are other classes and some can be null or empty (if value types).

On mapping the model to the dto, only a couple of members are mapped. The remaining are all set to null.

On downgrading to 4.2.1, the mapping is successful. I've tried creating a gist for a repro but the gist works as expected with 5.0.0. However the test classes where not that complex as in my actual project.

As I cannot reproduce the error, would it be better to add Automapper into my project and debug the mapping?

lbargaoanu:
skip property mapping only when a matching constructor was really used for the destination. 
The initial implementation was kind of silly :blush:. But this one might be too much considering the purpose. So maybe forget the whole thing. It would mean that if you have a setter, it will be called, even if it was already mapped though the constructor.

### DIFF
--- a/src/UnitTests/Constructors.cs
+++ b/src/UnitTests/Constructors.cs
@@ -5,6 +5,175 @@ using Should;
 
 namespace AutoMapper.UnitTests.Constructors
 {
+    public class When_constructor_is_partial_match_with_value_type : AutoMapperSpecBase
+    {
+        GeoCoordinate _destination;
+
+        public class GeolocationDTO
+        {
+            public double Longitude { get; set; }
+            public double Latitude { get; set; }
+            public double? HorizontalAccuracy { get; set; }
+        }
+
+        public struct GeoCoordinate
+        {
+            public GeoCoordinate(double longitude, double latitude, double x)
+            {
+                Longitude = longitude;
+                Latitude = latitude;
+                HorizontalAccuracy = 0;
+            }
+            public double Longitude { get; set; }
+            public double Latitude { get; set; }
+            public double? HorizontalAccuracy { get; set; }
+        }
+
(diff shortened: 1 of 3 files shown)


jbogard:
We took away the setting via property because people complained about it. The idea was that a member could be mapped via ctor arg so would be skipped in the properties.

lbargaoanu:
I know. I was just saying that I don't think it's worth the extra complexity.

jbogard:
Is that what the original issue was? I couldn't tell as someone else brought up an issue.

lbargaoanu:
I think so. The only repro I got was about this. And the effect is indeed skipping property mapping. Of course I cannot be sure without a repro for the original issue, but I guess that's not coming anytime soon :).

src/UnitTests/Constructors.cs
```diff
@@ -5,6 +5,175 @@
...
+        {
+            public double Longitude { get; set; }
+            public double Latitude { get; set; }
+            public double? HorizontalAccuracy { get; set; }
+        }
+
+        public struct GeoCoordinate
+        {
+            public GeoCoordinate(double longitude, double latitude, double x)
+            {
```
jbogard: This looks like it should fail configuration validation. We have only one constructor, and there is an unmapped ctor arg.

lbargaoanu: It's a value type. It has a default constructor.

jbogard: Lol duh

lbargaoanu: But if you meant if it were a class and we would validate constructors, then yes, it should :)

PR #1450, opened 2016-07-05
Adding back in destinations for value resolvers; Added new member val…

jbogard:
…ue resolvers type.

This was necessary to be able to access destination values in a value resolver since they're gone from the resolution context.


TylerCarlson1:
So this should be able to replace ITypeConverter with ForAllPropertyMaps and IMemberValueResolver?

jbogard:
I dunno, it doesn't help if you've used MapFrom + TypeConverter.

TylerCarlson1:
True guess not.  ITypeConverter still needs to be redone I guess :)

PR #1452, opened 2016-07-05
Open generics allow any member configuration; removed a ton of dead c…

### LINKED ISSUE #1427: Calling ForMember with open generics throws NullRef
Repro: 

```
using AutoMapper;

namespace ConsoleApplication3 {
    class Program {
        static void Main(string[] args) {
            Mapper.Initialize(mapper => {
                mapper.CreateMap(typeof(Foo<>), typeof(Bar<>))
                .ForMember("C", to => to.MapFrom("A"))
                .ForMember("D", to => to.MapFrom("B"));
            });
        }
    }

    public class Foo<T> {
        public int A { get; set; }
        public int B { get; set; }
    }

    public class Bar<T> {
        public int C { get; set; }
        public int D { get; set; }
    }
}
```

If the property names are the same, i.e no need for ForMember calls it works as expected.
Automapper 5.0.0
.net 4.5.2

jbogard:
…ode; fixes #1427


TylerCarlson1:
I'm not sad to see the whole MemberAccess framework go.

jbogard:
@TylerCarlson1 lol no me either. I don't think this PR will work for things like open generic value resolvers, but hey, this is a start.

PR #1463, opened 2016-07-06
Map only the dynamic members necessary

jbogard:


jbogard:
@lbargaoanu I stole your test etc but just exposed the configuration directly rather than making the configuration provider implement the interface.

lbargaoanu:
:+1:

PR #1466, opened 2016-07-06
Validating constructor mapping as part of configuration validation

### LINKED ISSUE #1456: AutoMapper fails to map class with non-default constructor
When upgrading to version 5.0.0 (from nuget) I found this strange bug. 

I have two classes with one property each with different names. The target class has two constructors one default and one that take a value to fill the property. The parameter has the same name as the property. AutoMapper fail to map this this example. See unit test below. 

```
    [TestClass]
    public class AutoMapperBug
    {
        [TestMethod]
        public void Test()
        {
            Mapper.Initialize(config =>
            {
                config.CreateMap<ClassA, ClassB>().ForMember(dest => dest.PropertyB, opt => opt.MapFrom(src => src.PropertyA));
            });

            ClassA classA = new ClassA { PropertyA = "test" };
            ClassB classB = Mapper.Map<ClassB>(classA);
            Assert.AreEqual("test", classB.PropertyB);
        }
    }

    public class ClassA
    {
        public string PropertyA { get; set; }

    }

    public class ClassB
    {
        public ClassB()
        {
        }

        public ClassB(string propertyB)
        {
            PropertyB = propertyB;
        }

        public string PropertyB { get; set; }
    }
```

### LINKED ISSUE #1457: AutoMapper 5.0.0 introduces runtime errors that AssertConfigurationIsValid doesn't catch
I mentioned in #1455 that I have two issues since upgrading to 5.0.0. Here's the other one.

There are quite a lot of mappings in our project that are, by the standards of 5.0.0, pretty broken. For one thing, there are static calls to `Mapper.Map()` inside the configurations; those I'm working on removing.

But there are also mapping targets which have private constructors written explicitly to allow AutoMapper 2.2.1 to construct the objects and then map properties from some DTO. Many of these now throw runtime errors at mapping time with the pretty uninformative error message

> System.ArgumentException: MyProject.Foo.DestinationType needs to have a constructor with 0 args or only optional args

Since these errors aren't discovered by the configuration verification, we're pretty screwed; we don't have test coverage that excercises all possible mappings, so we won't find all of these until someone stumbles on them in production. Unless this can be resolved, this is pretty much a dealbreaker for our upgrade :(

jbogard:
Fixes #1456 
Fixes #1457

### DIFF
--- a/src/AutoMapper/AutoMapperConfigurationException.cs
+++ b/src/AutoMapper/AutoMapperConfigurationException.cs
@@ -14,11 +14,13 @@ namespace AutoMapper
         {
             public TypeMap TypeMap { get; }
             public string[] UnmappedPropertyNames { get; }
+            public bool CanConstruct { get; }
 
-            public TypeMapConfigErrors(TypeMap typeMap, string[] unmappedPropertyNames)
+            public TypeMapConfigErrors(TypeMap typeMap, string[] unmappedPropertyNames, bool canConstruct)
             {
                 TypeMap = typeMap;
                 UnmappedPropertyNames = unmappedPropertyNames;
+                CanConstruct = canConstruct;
             }
         }
 
@@ -78,7 +80,7 @@ namespace AutoMapper
                 {
                     var message =
                         new StringBuilder(
-                            "\nUnmapped members were found. Review the types and members below.\nAdd a custom mapping expression, ignore, add a custom resolver, or modify the source/destination type\n");
+                            "\nUnmapped members were found. Review the types and members below.\nAdd a custom mapping expression, ignore, add a custom resolver, or modify the source/destination type\nFor no matching constructor, add a no-arg ctor, add optional arguments, or map all of the constructor parameters");
 
                     foreach (var error in Errors)
                     {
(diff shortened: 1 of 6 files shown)


lbargaoanu:
What if they pass in the destination? :)

lbargaoanu:
Or any other case when AM doesn't create the destination object.

jbogard:
¯\_(ツ)_/¯

jbogard:
So I guess the other case would be UseDestinationValue?

jbogard:
The only way you could get around this really is to add an option to ignore constructor validation I think.

lbargaoanu:
Exactly :)

lbargaoanu:
But the default should be not to validate in my view.

lbargaoanu:
In the other PR I was thinking about [this](https://github.com/AutoMapper/AutoMapper/blob/781f0533f34564a6af3b8a8654cf14e9f2e4f73b/src/AutoMapper/Execution/TypeMapPlanBuilder.cs#L238). Forgot about UseDestinationValue :)

jbogard:
I made the assumption that normally people don't use UseDestinationValue or
mapping to an existing object. The default is to map to the ctors, so the
expression generated has that "throw" in it.

lbargaoanu:
I guess we'll see :)

PR #1485, opened 2016-07-07
CustomExpressions don't use dest type and have nullable logic

### LINKED ISSUE #1483: Failure mapping from short? to short
(I don't think this is a duplicate of [1481](https://github.com/AutoMapper/AutoMapper/issues/1481), this is going in the opposite direction.)

After upgrading from 5.0.0 to 5.0.1 (or 5.0.2), I am getting this error:

System.InvalidOperationException: Rewriting child expression from type 'System.Int16' to type 'System.Nullable`1[System.Int16]' is not allowed, because it would change the meaning of the operation. If this is intentional, override 'VisitUnary' and change it to allow this rewrite..

(As a side note, the issues I've run into during Seal have been somewhat difficult to diagnose, e.g. having to comment out mappings until I narrow in on the one that caused the problem. Is it possible to add more diagnostic information like source/dest type/member into the exceptions? I can create a separate issue for this if necessary.)

Here is an example mapping:
```
[TestClass]
public class UnitTest1
{
	static UnitTest1()
	{
		Mapper.Initialize(cfg =>
		{
			cfg.CreateMap<Source, Target>()
				.ForMember(t => t.X, opts => opts.MapFrom(s => s.X > 0 ? s.X : default(short?)));
		});
	}

	[TestMethod]
	public void IncludedMappings()
	{
		var src = new Source
		{
			X = 0
		};

		var dest = Mapper.Map<Source, Target>(src);

		Assert.AreEqual(src.X, dest.X);
	}
}

class Source
{
	public short X { get; set; }
}

class Target
{
	public short? X { get; set; }
}
```
Another interesting aside is that the exception is significantly different if mapping int instead of short:
```
System.InvalidOperationException: The binary operator GreaterThan is not defined for the types 'System.Nullable`1[System.Int32]' and 'System.Int32'..
    at System.Linq.Expressions.Expression.GetUserDefinedBinaryOperatorOrThrow(ExpressionType binaryType, String name, Expression left, Expression right, Boolean liftToNull)
   at System.Linq.Expressions.Expression.GreaterThan(Expression left, Expression right, Boolean liftToNull, MethodInfo method)
```

TylerCarlson1:
Fixes #1483 using Option 2 in https://github.com/AutoMapper/AutoMapper/issues/1483#issuecomment-231195506

### DIFF
--- a/src/UnitTests/NullBehavior.cs
+++ b/src/UnitTests/NullBehavior.cs
@@ -16,6 +16,7 @@ namespace AutoMapper.UnitTests
 				public ModelSubDto Sub { get; set; }
 				public int SubSomething { get; set; }
                 public int? SubSomethingNullDest { get; set; }
+                public int? NullableMapFrom { get; set; }
                 public string NullString { get; set; }
 			}
 
@@ -44,7 +45,8 @@ namespace AutoMapper.UnitTests
 		    protected override MapperConfiguration Configuration { get; } = new MapperConfiguration(cfg =>
 		    {
 		        cfg.AllowNullDestinationValues = false;
-		        cfg.CreateMap<ModelObject, ModelDto>();
+		        cfg.CreateMap<ModelObject, ModelDto>()
+                    .ForMember(d => d.NullableMapFrom, opt => opt.MapFrom(s => s.Sub.Something));
 		        cfg.CreateMap<ModelSubObject, ModelSubDto>();
 
 		    });
@@ -79,6 +81,12 @@ namespace AutoMapper.UnitTests
             public void Should_return_null_for_nullable_properties()
             {
                 _result.SubSomethingNullDest.ShouldBeNull();
+            }
+
+            [Fact]
+            public void Should_return_not_null_for_nullable_properties_that_are_user_defined()
+            {
+                _result.NullableMapFrom.ShouldEqual(0);
             }
 
             [Fact]
--- a/src/AutoMapper/ConstructorParameterMap.cs
+++ b/src/AutoMapper/ConstructorParameterMap.cs
@@ -39,7 +39,7 @@ namespace AutoMapper
             ParameterExpression ctxtParam)
         {
             if (CustomExpression != null)
-                return CustomExpression.ConvertReplaceParameters(srcParam).IfNotNull(DestinationType);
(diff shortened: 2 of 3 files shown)


jbogard:
Ah, I see what you mean now.

...test?

src/UnitTests/NullBehavior.cs
```diff
@@ -82,6 +84,12 @@ public void Should_return_null_for_nullable_properties()
             }
 
             [Fact]
+            public void Should_return_not_null_for_nullable_properties_that_are_user_defined()
+            {
+                _result.NullableMapFrom.ShouldEqual(0);
```
jbogard: Shouldn't this be null? If the member was a regular flattened one it would be.

TylerCarlson1: I refer to line 49.  Cause it was user defined from MapFrom, ala a CustomExpression we use the MapFrom's expression's return type which is int.

Yea it's not the best example.  The issue is if I put conditions `? :` into line 49 the seal would throw an exception.  So I'm just saying if user defined it it's possible they are there and therefore we can't assume the nullable return type will work.  So tests just say it doesn't null on user defined.  Again not the best of tests.

jbogard: So in this case I think the test is wrong, the result should be null. In
the original issue, it should be what was expected there too.

TylerCarlson1: OK then if that's the case you need a MUCH more complicated solution editing the NullExpressionVisitor then.  Just make every conditional option you have not throw an exception.

jbogard:
So I think the way we fix this (after my latest PR) is that the IfNotNull method needs to take the null expression instead of assuming Default(DestinationPropertyType).

TylerCarlson1:
Maybe I guess your PR works in more scenarios.  I just think there's eventually going to be others that are similar but won't work.

Given what you did now I would suggests not even bothering with the ExpressionVisitor and just do a try catch and return the default value type if any exception happens.  If it does they will see it in debugger and it won't happen in code.  Performance loss will be if they have a bunch of nulls failing tests, because of the catches, but they can just code the expression better themselves if they need to.  Which should be very rare.

And that way you don't have to worry about the nitty gritty of the expression possibilities.  I tried to wrap my head around it and I can't right now.

PR #1493, opened 2016-07-08
Enumerators

lbargaoanu:
Generate better enumeration code for concrete types. Avoid allocations and interface dispatch by using the enumerator pattern for collections and a for loop for arrays.

### DIFF
--- a/src/Benchmark/FlatteningMapper.cs
+++ b/src/Benchmark/FlatteningMapper.cs
@@ -232,13 +232,13 @@ namespace Benchmark.Flattening
 
             public Foo Foo1 { get; set; }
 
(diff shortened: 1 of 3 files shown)


src/Benchmark/FlatteningMapper.cs
```diff
@@ -67,7 +67,7 @@ public class Customer
             public Address Address { get; set; }
             public Address HomeAddress { get; set; }
             public Address[] Addresses { get; set; }
-            public ICollection<Address> WorkAddresses { get; set; }
+            public List<Address> WorkAddresses { get; set; }
```
lbargaoanu: Just to see results :) The optimization applies only for concrete types, the same as the compiler does it.

PR #1508, opened 2016-07-11
Only doing null fixing on custom member mappings when there are only …

### LINKED ISSUE #1505: Exception when configuring mappings (worked in 4.2.1)
Since upgrading to 5.0.2 it seems one of my mappings in throwing the following exception - 
`Expression of type 'System.Nullable`1[System.DateTime]' cannot be used for return type 'System.DateTime'`

The destination property is a nullable date time and the mapping is as follows -
`ForMember(Function(dest) dest.EndDate, Sub(opt) opt.MapFrom(Function(src) If(src.ExamType = ExamType.Makeup, src.ExamSchedules.Max(Function(m) m.Schedule.DtEnd), src.ExamSchedules.Where(Function(f) f.Schedule.BHide = False).Max(Function(m) m.Schedule.DtEnd)))).`

If I try to change the destination to just Date then the configuration doesn't throw an exception but I will get an exception later on when my code is run that uses the mapping because there are times where there are no values and it should be null.

Not sure if there is something I'm missing or whether this is a bug or already discussed.

jbogard:
…member accesses, otherwise doing try catches. Fixes #1505

### DIFF
--- a/src/UnitTests/NullBehavior.cs
+++ b/src/UnitTests/NullBehavior.cs
@@ -152,9 +152,15 @@ namespace AutoMapper.UnitTests
             }
 
             [Fact]
-            public void Should_return_not_null_for_nullable_properties_that_are_user_defined()
+            public void Should_return_null_for_nullable_properties_that_are_complex_map_froms()
             {
                 _result.SubExpressionName.ShouldEqual(null);
+            }
+
+            [Fact]
+            public void Should_return_null_for_nullable_properties_that_have_member_access_map_froms()
+            {
+                _result.NullableMapFrom.ShouldEqual(null);
             }
 
             [Fact]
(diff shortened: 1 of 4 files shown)


src/UnitTests/NullBehavior.cs
```diff
@@ -139,6 +152,12 @@ protected override void Because_of()
             }
 
             [Fact]
+            public void Should_return_not_null_for_nullable_properties_that_are_user_defined()
+            {
+                _result.SubExpressionName.ShouldEqual(null);
```
TylerCarlson1: ShouldNotEqual null?

jbogard: Eh, test name is bad.

PR #1512, opened 2016-07-12
Configuration faster

jbogard:
Made configuration faster, about 2-3x in a local project. @TylerCarlson1 I poked a bit around the convention stuff to precalculate some stuff that was getting calculated every time.

### DIFF
--- a/src/AutoMapper/Mappers/DynamicMappers.cs
+++ b/src/AutoMapper/Mappers/DynamicMappers.cs
@@ -18,7 +18,7 @@ namespace AutoMapper.Mappers
         {
             if (destination == null)
                 destination = ifNull();
-            var destinationTypeDetails = new TypeDetails(typeof(TDestination), context.ConfigurationProvider.Configuration);
+            var destinationTypeDetails = context.ConfigurationProvider.Configuration.CreateTypeDetails(typeof(TDestination));
             foreach (var member in destinationTypeDetails.PublicWriteAccessors)
             {
                 object sourceMemberValue;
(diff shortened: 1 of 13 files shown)


TylerCarlson1:
Yea looks good for the most part.  Have a hard time following Caching and Pre-Calculating config options but I assume copy paste refactor worked fine.

jbogard:
I thought about making the PrePost fix class do all the caching, but it was expensive to do the lookups vs. just putting the results on the TypeDetails. Then I also saw that we create TypeDetails (which is expensive) based on a profile, so why not cache them in the profile.

TylerCarlson1:
can you cache them in a dictionary that way the are cached for whatever part of Automapper uses them like you do with the expressions and typepair to maps?

jbogard:
Eh?

TylerCarlson1:
oh you are already doing it.  I saw CreateTypeDetails and forgot it cached.

src/AutoMapper/Mappers/DynamicMappers.cs
```diff
@@ -63,7 +63,7 @@ public class ToDynamicMapper : IObjectMapper
         {
             if (destination == null)
                 destination = ifNull();
-            var sourceTypeDetails = new TypeDetails(typeof(TSource), context.ConfigurationProvider.Configuration);
+            var sourceTypeDetails = context.ConfigurationProvider.Configuration.CreateTypeDetails(typeof(TSource));
```
lbargaoanu: Maybe expose configuration from the context directly.

PR #1514, opened 2016-07-13
Refactor from statics and parameters to instance methods and fields

lbargaoanu:
More readable code in my view.

### DIFF
--- a/src/AutoMapper/Execution/TypeMapPlanBuilder.cs
+++ b/src/AutoMapper/Execution/TypeMapPlanBuilder.cs
@@ -134,8 +134,7 @@ namespace AutoMapper.Execution
                     continue;
                 }
                 var property = TryPropertyMap(propertyMap);
-                //if(constructorMapping && _typeMap.ConstructorParameterMatches(propertyMap.DestinationProperty.Name))
-                if(_typeMap.ConstructorMap?.CanResolve == true && _typeMap.ConstructorParameterMatches(propertyMap.DestinationProperty.Name))
+                if(constructorMapping && _typeMap.ConstructorParameterMatches(propertyMap.DestinationProperty.Name))
                 {
                     property = IfThen(NotEqual(_initialDestination, Constant(null)), property);
                 }
--- a/src/AutoMapper/Execution/TypeMapPlanBuilder.cs
+++ b/src/AutoMapper/Execution/TypeMapPlanBuilder.cs
@@ -14,9 +14,9 @@ namespace AutoMapper.Execution
     public class TypeMapPlanBuilder
     {
         private static readonly Expression<Func<AutoMapperMappingException>> CtorExpression = () => new AutoMapperMappingException(null, null, default(TypePair), null, null);
-        private static Expression<Action<ResolutionContext>> IncTypeDepthInfo = ctxt => ctxt.IncrementTypeDepth(default(TypePair));
-        private static Expression<Action<ResolutionContext>> DecTypeDepthInfo = ctxt => ctxt.DecrementTypeDepth(default(TypePair));
-        private static Expression<Func<ResolutionContext, int>> GetTypeDepthInfo = ctxt => ctxt.GetTypeDepth(default(TypePair));
+        private static readonly Expression<Action<ResolutionContext>> IncTypeDepthInfo = ctxt => ctxt.IncrementTypeDepth(default(TypePair));
+        private static readonly Expression<Action<ResolutionContext>> DecTypeDepthInfo = ctxt => ctxt.DecrementTypeDepth(default(TypePair));
+        private static readonly Expression<Func<ResolutionContext, int>> GetTypeDepthInfo = ctxt => ctxt.GetTypeDepth(default(TypePair));
 
         readonly IConfigurationProvider configurationProvider;
         readonly TypeMap typeMap;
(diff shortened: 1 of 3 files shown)


src/AutoMapper/TypeMap.cs
```diff
@@ -263,7 +263,7 @@ public void Seal(TypeMapRegistry typeMapRegistry, IConfigurationProvider configu
                     .Union(_inheritedMaps)
                     .OrderBy(map => map.MappingOrder).ToArray();
 
-            MapExpression = TypeMapPlanBuilder.BuildMapperFunc(this, configurationProvider, typeMapRegistry);
+            MapExpression = new TypeMapPlanBuilder(configurationProvider, typeMapRegistry, this).CreateMapperLambda();
```
TylerCarlson1: I say put back TypeMapPlanBuilder.BuildMapperFunc and have it just implement this line.  Also make constructor and function call private, so that there's no way someone could accidentally use TypeMapPlanBuilder incorrectly.

src/AutoMapper/Execution/TypeMapPlanBuilder.cs
```diff
@@ -1,36 +1,57 @@
...
+        private static readonly Expression<Func<ResolutionContext, int>> GetTypeDepthInfo = ctxt => ctxt.GetTypeDepth(default(TypePair));
+
+        readonly IConfigurationProvider configurationProvider;
+        readonly TypeMap typeMap;
+        readonly TypeMapRegistry typeMapRegistry;
+        readonly ParameterExpression source;
+        readonly ParameterExpression initialDestination;
+        readonly ParameterExpression context;
+        bool constructorMapping;
+        ParameterExpression destination;
```
TylerCarlson1: Underscore Prefix?

lbargaoanu: For the fields you mean, yes.

src/AutoMapper/TypeMap.cs
lbargaoanu: Incorrectly how?

TylerCarlson1: They can make a new TypeMapPlanBuilder object, cache it and call it at same time on different threads.  Although looking at the code some more you might not have a problem in this scenario.

It would be a problem if the code wasn't thread safe, which is what I thought it might be given saving the parameters as fields and using them in function calls.

I don't expect people to access this class, but if they did you would want them to just build the function and not worry about the specifics.

jbogard:
I had contemplated moving this building to inside the TypeMap object, but I kinda thought it got too big. But everything inside the TypeMapPlanBuilder class now deals with a single TypeMap...what do you think?

TylerCarlson1:
Not sure what you are asking, but I think the expression stuff needs to stay isolated from everything else.

I thought that was your original idea was to push all the expression building logic into a single place away from TypeMap, PropertyMap, CtorParameterMap, ect.

src/AutoMapper/Execution/TypeMapPlanBuilder.cs
```diff
@@ -1,309 +1,280 @@
...
+        private static readonly Expression<Action<ResolutionContext>> DecTypeDepthInfo = ctxt => ctxt.DecrementTypeDepth(default(TypePair));
+        private static readonly Expression<Func<ResolutionContext, int>> GetTypeDepthInfo = ctxt => ctxt.GetTypeDepth(default(TypePair));
+
+        readonly IConfigurationProvider _configurationProvider;
+        readonly TypeMap _typeMap;
+        readonly TypeMapRegistry _typeMapRegistry;
+        readonly ParameterExpression _source;
+        readonly ParameterExpression _initialDestination;
+        readonly ParameterExpression _context;
+        bool _constructorMapping;
```
jbogard: Small thing - either these fields should be readonly, or parameters, but not mutable fields. It's not consistent that some are readonly fields, some are mutable, and some are parameters.

lbargaoanu:
About moving to TypeMap, although it makes sense, I think that expression code deserves its own class, beautiful as it is :)

src/AutoMapper/Execution/TypeMapPlanBuilder.cs
lbargaoanu: Yeah, something is somewhat fishy here :) I thought it's nicer code like this. I'll make it read-only fields and parameters and see how it looks.

lbargaoanu: Well, I've changed it and the constructor mapping thing looks weird, but it's your call.

jbogard: Go the other way then, make them all read only fields (sorry that's what I meant)

jbogard: The idea being that the fields are either set once, or you use a parameter as a choice for each field. But mutable fields gets hard to track when they're set.

lbargaoanu: So now there are only read-only fields. But the constructorMapping parameter is passed around an returned as out, a little weird. But if you prefer that to mutating a field, it's done.

lbargaoanu:
I think it's just a misunderstanding. Added a failing test. So the field was written once, but by default was false. And the parameter is actually written twice, first on method entry and then when we decide we're actually doing constructor mapping, So if I keep the old condition, we're good to go. I think :)

lbargaoanu:
Can we merge this? There are a few fixes that want to touch this file.

PR #1520, opened 2016-07-13
Fix #1518

TylerCarlson1:
Have to check for `IDictionary<TKey,TValue>` explicitly for instances that inherit dictionary but aren't generic.  Before would get `IEnumerable<KeyValuePair<TKey,TValue>>`

Issue #1518, opened 2016-07-13
Index was outside the bounds of the array.


niemyjski:
I'm still investigating this issue but I wanted to log it and get the ball rolling...

```csharp
            container.RegisterSingleton<IMapper>(() => {
                var profiles = container.GetAllInstances<Profile>();
                var config = new MapperConfiguration(cfg => {
                    cfg.ConstructServicesUsing(container.GetInstance);

                    foreach (var profile in profiles)
                        cfg.AddProfile(profile);
                }); // blows up here..
                return config.CreateMapper();
            });
```

I only have one mapping and this seems to be created okay..

```csharp

        public class ApiMappings : Profile {
            public ApiMappings() {
                CreateMap<UserDescription, EventUserDescription>();
                
                CreateMap<NewOrganization, Organization>();
                CreateMap<Organization, ViewOrganization>().AfterMap((o, vo) => {
                    vo.IsOverHourlyLimit = o.IsOverHourlyLimit();
                    vo.IsOverMonthlyLimit = o.IsOverMonthlyLimit();
                });

                CreateMap<StripeInvoice, InvoiceGridModel>().AfterMap((si, igm) => igm.Id = igm.Id.Substring(3));

                CreateMap<NewProject, Project>();
                CreateMap<Project, ViewProject>();

                CreateMap<NewToken, Token>().ForMember(m => m.Type, m => m.Ignore());
                CreateMap<Token, ViewToken>();
                
                CreateMap<User, ViewUser>();

                CreateMap<NewWebHook, WebHook>();
            }
        }
```

```
System.IndexOutOfRangeException occurred
  HResult=-2146233080
  Message=Index was outside the bounds of the array.
  Source=AutoMapper
  StackTrace:
       at AutoMapper.Mappers.CollectionMapperExtensions.MapKeyPairValueExpr(TypeMapRegistry typeMapRegistry, IConfigurationProvider configurationProvider, PropertyMap propertyMap, Type sourceType, Type destType, ParameterExpression itemParam, Expression contextParam)
  InnerException: 
```

``` 
[...]

PR #1522, opened 2016-07-14
ExpandoObject is a dictionary and should be mapped as such

lbargaoanu:
#1418 and #1518 are the same issue (as the numbers clearly suggest).
#1520 fixes both. 
#1430 is not needed.


jbogard:
Well where does that leave the dynamic mappers? Should they be removed?

lbargaoanu:
Nooooo :) Dynamic objects are different from Expando, they don't work like dictionaries. They can, but they don't have to. Remove them and you'll see failing tests.

Issue #1377, opened 2016-06-28
Inline sub map expressions


jbogard:
When a type map has a member that itself is a type map, today we go through a process of calling the ResolutionContext.Map method.

For most cases we probably don't need this. ResolutionContext.Map should be reserved when it needs to determine the mapping strategy at runtime.

I think we do this with a sort of two-pass mapping. We leave placeholder expressions in the tree, then the second pass fills in the placeholders with inlined calls for child maps, or with a regular context mapping call.

We could probably start simple, not worry about inheritance or collection stuff.

This is now where I see the bulk of time spent in the profiling.

TylerCarlson1:
I think inheritance would be the first, otherwise tests won't work.  In which case what you have to do is put all the conditions into a single SUPER expression that conditions all the inherited types and then compiles all the mapping expressions.

Funny thing is if you do this, then there's even more performance gain from something else.  At this point when you search for mapping function, you don't need to know the object types for Generic map calls because the function will take care of that for you.  And from there you don't need to make MapRequest or a new TypeMap saving time there.  Also it might make sense to go back to my `TypeMap<TSource,TDest>` static class idea, so you don't have to search a possibly big dictionary to find the expression.

TylerCarlson1:
Oh yea and the two pass sample isn't going to be any good I think because what you would want is to pass the inline expression over and replace parameters, but you can't do it cause there's the create map part.  If you were going to implement it without fixing inheritance problem all you would be doing is replacing ResolutionContext.Map with an expression doing the same thing, providing little performance gain.  And when you do resolve the inheritance problem you will remove all that expression code with a simple ReplaceParameters call.

That's what I think.  I could be overly complicating things as well.  Happened [before](https://github.com/AutoMapper/AutoMapper/pull/1353) :)

PR #1523, opened 2016-07-14
Reuse the default context even only when MaxDepth and PreserverRefer…

### LINKED ISSUE #1516: Multiple destination type maps for the same source type fails and uses the first type instead when PreserveReferences is specified.
If multiple CreateMap calls exist for the same type the first type is then retained for subsequent calls and will cause an invalid cast exception to occur. It would appear that **PreserveReferences** is holding on after the initial Map operation. See below for example. I have similar code working on 4.x but this now fails Unit Testing on 5.0.2. Removing PreserveChanges allows the map to work as expected. If a Dynamic EF6 entity is used then removing PreserveChanges results in a StackOverflow but that is already covered by a separate issue report.

Here's the exception the code below generates:
```
Exception thrown: 'System.InvalidCastException' in AutoMapper.dll

Additional information: Unable to cast object of type 'DtoOne' to type 'DtoThree'.
```

The code sample below,

Creates two maps,
```
        public void TestFailure()
        {
            try
            {
                var mapConfig = new MapperConfiguration(expression =>
                {
                    expression.CreateMap<EntityTwo, DtoTwo>().PreserveReferences();
                    expression.CreateMap<EntityOne, DtoOne>().PreserveReferences();

                    // If you add this map then it will attempt to convert from DtoTwo to DtoThree.
                    expression.CreateMap<EntityOne, DtoThree>().PreserveReferences();
                });

                var mapper = mapConfig.CreateMapper();

                // Now try it.
                var objects = new List<EntityOne>() { new EntityOne() {
                  Two  = new EntityTwo()
                {
                    Ones = new List<EntityOne>()
                    {
                        new EntityOne()
                        {
                            Two = new EntityTwo() {Ones = new List<EntityOne>()}
                        }
                    }

                } } };

                var res = objects.Select(mapper.Map<EntityOne, DtoOne>).ToList();

                // This should convert EntityOne to DtoThree but instead it attempts to convert from DtoOne to DtoThree thus is using the previously defined map.
                var res2 = objects.Select(mapper.Map<EntityOne, DtoThree>).ToList();
            }
            catch (Exception)
            {
                // Boom.
                throw;
            }
        }


        public class DtoOne
        {
            public DtoTwo Two { get; set; }
        }

        public class DtoTwo
        {
            public virtual ICollection<DtoOne> Ones { get; set; }
        }

        public class DtoThree
        {
            public int Id { get; set; }
        }

        public class EntityOne
        {
            public int Id { get; set; }
            public EntityTwo Two { get; set; }
        }

        public class EntityTwo
        {
            public virtual ICollection<EntityOne> Ones { get; set; }
        }
```

### LINKED ISSUE #1551: Mapping same object second time returns reference to the object mapped earlier
Hello, maybe im missing something but that's big problem for my code.
Lets assume that we have Parent and Child objects - they refers to each other like
`public Parent`
`{`
`            public string name { get; set; }`
`            public ICollection<Child> children { get; set; }`
`}`
`public Child`
`{`
`            public string name { get; set; }`
`            public Parent parent { get; set; }`
`}`
And corresponding models ParentModel and ChildModel
So i created map with .PreserveReferences() like
`CreateMap<Parent, ParentModel>().PreserveReferences();`
`CreateMap<Child, ChildModel>().PreserveReferences();`
to avoid StackOverflowException while mapping. The issue is if i will map in example some Parent object twice, even with different data i'll get reference to the same object:
`Parent _testParent.name = new Parent() { name = "x" };`
`ParentModel _testParentModel1= Mapper.Map<ParentModel >(_testParent);`
`Parent _testParent.name = new Parent() { name = "xxx" };`
`ParentModel _testParentModel2= Mapper.Map<ParentModel >(_testParent);`

After this _testParentModel2 is reference to _testParentModel1.
Is it possible to map same object multiple times and get two separated objects when mapping without `.PreserveReferences()` causes StackOverflowException ?

lbargaoanu:
…ences are not set.
The straightforward way would be to allocate it. This tries not to :) It's weird but it should work.
Fixes #1516. Fixes #1551.


jbogard:
Ahhhhhh ok, I like the idea here, but I don't want ThreadStatic in AutoMapper. It was there a long time ago, and just wound up confusing the hell out of me.

I know this helps the benchmarks, but in reality I don't think it'll make much of a difference, since no one really runs 1M maps all in a row.

lbargaoanu:
Not really about the benchmark. But if everybody starts using those 2, it would mean allocate a context for most calls. So allocate when they are set?

lbargaoanu:
There is also ThreadLocal, somewhat slower, but the same idea.

jbogard:
So what does this fix? Or is it an optimization?

lbargaoanu:
Fixes #1516.

lbargaoanu:
And enables sharing in a thread safe manner a "default context". Broken right now.

jbogard:
So I'm missing how the original issue is solved by holding on to one instance of a resolution context. We create just one resolution context per map call, is there something holding on to instances past the first map call?

jbogard:
Ohhhh I see it, the default context. I'd rather not cache that at all. Less state is better here I think. The default mapping options is OK to cache, because it only holds custom options they pass in and if they don't pass any in, you can use that. It's more of a null object pattern for the default options than the resolution context, which is intended to be distinct per call.

lbargaoanu:
Yes, but when neither MaxDepth nor PreserveReferences are set, the context is just a shell, generic. I think it makes sense to reuse it then. It's true that we would have to check the type map for those settings.

jbogard:
Can't we detect that then? In our overall mapping func for types, we can detect if a resolution context has been passed in, something like:

```
context = context ?? DefaultMappingContext
```

Statics I'm looking for things that don't change from map to map, are global to configuration (many of our caches are static and shouldn't be, for one). I would rather not cache it all than take on thread local stuff, that stuff is just too confusing.

lbargaoanu:
Yes, something like that. I'll give it a shot.

lbargaoanu:
Updated.

lbargaoanu:
Can we keep the markers in BenchEngine.cs? :) They don't affect anything and I find them useful when profiling allocations.

lbargaoanu:
@jbogard Can we merge this? I wanted to do smth about #1377 and it touches the same code.

jbogard:
Looks good!

PR #1530, opened 2016-07-15
Make constructor mapping a little more lenient

### LINKED ISSUE #1528: Invalid constructor mappings throw exceptions at seal-time
Given the following classes and mapping configuration, I get an `InvalidOperationException` (see stack trace below the code). If I comment out the constructor completely, I instead get a `NullReferenceException` (see second stack trace). Since the constructor parameters are specified with strings and not strongly typed, it would be nice if the configuration process was a little more lenient and could report these problems with better error messages.

```c#
[Fact]
public void ThrowSomeStuffAround()
{
    new MapperConfiguration(c =>
    {
        c.CreateMap<Source, DestinationA>()
            .ForCtorParam("bar", cfg => cfg.MapFrom(src => src.Foo));
    }).AssertConfigurationIsValid();
}

public class Source
{
}

public class DestinationA
{
    public DestinationA(string barf)
    {
        Foo = barf;
    }

    public string Foo { get; }
}
```

`InvalidOperationException`, thrown when no matching constructor parameter is found:

```
System.InvalidOperationException : Sequence contains no matching element
Stack Trace:
     at System.Linq.Enumerable.Single[TSource](IEnumerable`1 source, Func`2 predicate)
     at AutoMapper.CtorParamConfigurationExpression`1.Configure(TypeMap typeMap)
     at AutoMapper.Configuration.MappingExpression`2.Configure(IProfileConfiguration profile, TypeMap typeMap)
     at AutoMapper.Profile.BuildTypeMap(TypeMapRegistry typeMapRegistry, ITypeMapConfiguration config)
     at AutoMapper.Profile.AutoMapper.IProfileConfiguration.Register(TypeMapRegistry typeMapRegistry)
     at AutoMapper.MapperConfiguration.Seal(IConfiguration configuration)
     at AutoMapper.MapperConfiguration..ctor(MapperConfigurationExpression configurationExpression, IEnumerable`1 mappers)
     at AutoMapper.MapperConfiguration..ctor(Action`1 configure, IEnumerable`1 mappers)
     at AutoMapper.MapperConfiguration..ctor(Action`1 configure)
```

`NullReferenceException`, thrown when no constructor is found:

```
System.NullReferenceException : Object reference not set to an instance of an object.
Stack Trace:
     at AutoMapper.CtorParamConfigurationExpression`1.Configure(TypeMap typeMap)
     at AutoMapper.Configuration.MappingExpression`2.Configure(IProfileConfiguration profile, TypeMap typeMap)
     at AutoMapper.Profile.BuildTypeMap(TypeMapRegistry typeMapRegistry, ITypeMapConfiguration config)
     at AutoMapper.Profile.AutoMapper.IProfileConfiguration.Register(TypeMapRegistry typeMapRegistry)
     at AutoMapper.MapperConfiguration.Seal(IConfiguration configuration)
     at AutoMapper.MapperConfiguration..ctor(MapperConfigurationExpression configurationExpression, IEnumerable`1 mappers)
     at AutoMapper.MapperConfiguration..ctor(Action`1 configure, IEnumerable`1 mappers)
     at AutoMapper.MapperConfiguration..ctor(Action`1 configure)
```

I stumbled upon this because I wanted to create an extension on `IMappingExpression<TSrc, TDest>` called `Rename`, which would simply register both `ForMember` and `ForCtorParam` (assuming that the ctor param is named like the destination property but in camelCase), which of course blows up when using that extension method on types with no constructor.

tlycken:
Before, all attempts at mapping constructor parameters for classes that
did not have a constructor defined, or where no defined constructor had
a parameter with that name, would throw an exception with a pretty poor
uninformative exception message.

These scenarios are now considered valid mapping configurations, as long
as the stuff that's needed to actually perform a mapping is also
configured.

Fixes #1528

### DIFF
--- a/src/UnitTests/IMappingExpression/ForCtorParam.cs
+++ b/src/UnitTests/IMappingExpression/ForCtorParam.cs
@@ -1,5 +1,4 @@
 ﻿using System;
-using System.Linq.Expressions;
 using Xunit;
 using Should;
 
@@ -20,6 +19,11 @@ namespace AutoMapper.UnitTests
             }
 
             public int Value1 { get; }
+        }
+
+        public class DestWithNoConstructor
+        {
+            public int Value1 { get; set; }
         }
 
         protected override MapperConfiguration Configuration { get; } = new MapperConfiguration(cfg =>
@@ -50,6 +54,53 @@ namespace AutoMapper.UnitTests
             var dest = mapper.Map<Source, Dest>(new Source { Value = 5 });
 
             dest.Value1.ShouldEqual(8);
+        }
+
+        [Fact]
+        public void Should_ignore_nonexistent_parameter()
+        {
+            var config = new MapperConfiguration(cfg =>
+            {
+                cfg.CreateMap<Source, Dest>()
+                    .ForCtorParam("thing", opt => opt.MapFrom(src => src.Value))
+                    .ForCtorParam("think", opt => opt.MapFrom(src => src.Value));
+            });
+
+            config.AssertConfigurationIsValid();
+
+            var result = config.CreateMapper().Map<Dest>(new Source { Value = 42 });
+
+            result.Value1.ShouldEqual(42);
+        }
+
+        [Fact]
+        public void Should_ignore_when_no_constructor_is_present()
+        {
+            var config = new MapperConfiguration(cfg =>
+            {
+                cfg.CreateMap<Source, DestWithNoConstructor>()
+                    .ForMember(dest => dest.Value1, opt => opt.MapFrom(src => src.Value))
+                    .ForCtorParam("thing", opt => opt.MapFrom(src => src.Value));
+            });
+
+            config.AssertConfigurationIsValid();
+
+            var result = config.CreateMapper().Map<DestWithNoConstructor>(new Source { Value = 17 });
+
+            result.Value1.ShouldEqual(17);
+        }
+
+        [Fact]
+        public void Should_not_pass_config_validation_when_parameter_is_mispelt()
+        {
+            var config = new MapperConfiguration(cfg =>
+            {
+                cfg.CreateMap<Source, Dest>()
+                    .ForCtorParam("think", opt => opt.MapFrom(src => src.Value));
+            });
+
+            Action configValidation = () => config.AssertConfigurationIsValid();
+            configValidation.ShouldThrow<AutoMapperConfigurationException>();
         }
     }
 }
--- a/src/UnitTests/IMappingExpression/ForCtorParam.cs
+++ b/src/UnitTests/IMappingExpression/ForCtorParam.cs
@@ -57,62 +57,51 @@ namespace AutoMapper.UnitTests
         }
 
         [Fact]
-        public void Should_ignore_nonexistent_parameter()
+        public void Should_throw_on_nonexistent_parameter()
         {
-            var config = new MapperConfiguration(cfg =>
+            Action configuration = () => new MapperConfiguration(cfg =>
             {
                 cfg.CreateMap<Source, Dest>()
                     .ForCtorParam("thing", opt => opt.MapFrom(src => src.Value))
                     .ForCtorParam("think", opt => opt.MapFrom(src => src.Value));
             });
-
-            config.AssertConfigurationIsValid();
-
-            var result = config.CreateMapper().Map<Dest>(new Source { Value = 42 });
-
-            result.Value1.ShouldEqual(42);
+            configuration.ShouldThrow<AutoMapperConfigurationException>(exception =>
+            {
+                exception.Message.ShouldContain("does not have a constructor with a parameter named 'think'.", StringComparison.InvariantCulture);
+                exception.Message.ShouldContain(typeof(Dest).FullName, StringComparison.InvariantCulture);
+            });
         }
 
         [Fact]
-        public void Should_ignore_when_no_constructor_is_present()
+        public void Should_throw_when_no_constructor_is_present()
         {
-            var config = new MapperConfiguration(cfg =>
+            Action configuration = () => new MapperConfiguration(cfg =>
             {
                 cfg.CreateMap<Source, DestWithNoConstructor>()
                     .ForMember(dest => dest.Value1, opt => opt.MapFrom(src => src.Value))
                     .ForCtorParam("thing", opt => opt.MapFrom(src => src.Value));
             });
 
-            config.AssertConfigurationIsValid();
-
-            var result = config.CreateMapper().Map<DestWithNoConstructor>(new Source { Value = 17 });
-
-            result.Value1.ShouldEqual(17);
+            configuration.ShouldThrow<AutoMapperConfigurationException>(exception =>
+            {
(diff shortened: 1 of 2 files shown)


src/UnitTests/IMappingExpression/ForCtorParam.cs
```diff
@@ -51,5 +55,52 @@ public void Should_resolve_using_custom_func()
...
         }
+
+        [Fact]
+        public void Should_ignore_nonexistent_parameter()
+        {
+            var config = new MapperConfiguration(cfg =>
+            {
+                cfg.CreateMap<Source, Dest>()
+                    .ForCtorParam("thing", opt => opt.MapFrom(src => src.Value))
+                    .ForCtorParam("think", opt => opt.MapFrom(src => src.Value));
```
lbargaoanu: This should throw. I see the other mapping, but if the parameter doesn't exist, this should be reported somehow.

src/UnitTests/IMappingExpression/ForCtorParam.cs
```diff
@@ -51,5 +55,52 @@ public void Should_resolve_using_custom_func()
...
+
+            config.AssertConfigurationIsValid();
+
+            var result = config.CreateMapper().Map<DestWithNoConstructor>(new Source { Value = 17 });
+
+            result.Value1.ShouldEqual(17);
+        }
+
+        [Fact]
+        public void Should_not_pass_config_validation_when_parameter_is_mispelt()
```
lbargaoanu: You misspelt :)

src/UnitTests/IMappingExpression/ForCtorParam.cs
tlycken: The point of this PR is exactly that it shouldn't matter, as long as there is still enough information for a valid mapping configuration.

Whether that behavior is desirable to more people than myself is of course debatable, but this is the behavior I want.

src/UnitTests/IMappingExpression/ForCtorParam.cs
```diff
@@ -51,5 +55,52 @@ public void Should_resolve_using_custom_func()
...
+        {
+            var config = new MapperConfiguration(cfg =>
+            {
+                cfg.CreateMap<Source, Dest>()
+                    .ForCtorParam("think", opt => opt.MapFrom(src => src.Value));
+            });
+
+            Action configValidation = () => config.AssertConfigurationIsValid();
+            configValidation.ShouldThrow<AutoMapperConfigurationException>();
```
lbargaoanu: There should be a check that the exception is about that parameter. The message should contain the offending name and map.

src/UnitTests/IMappingExpression/ForCtorParam.cs
tlycken: Haha, that's sort of like iron... ;)

src/UnitTests/IMappingExpression/ForCtorParam.cs
lbargaoanu: And what happens when the mapping is called?

src/UnitTests/IMappingExpression/ForCtorParam.cs
lbargaoanu: I understand, I just said I disagree. What if there are more parameters?

src/UnitTests/IMappingExpression/ForCtorParam.cs
tlycken: When the mapping is called, an `ArgumentException` is thrown with the message "Dest needs to have a constructor with 0 args or only optional args", i.e. the same behavior as if the setup had been just `cfg.CreateMap<Source, Dest>()`.

lbargaoanu: A test is worth a thousand words :)

tlycken: How do I get access to the thrown exception to verify the message contents?

(The message is just "No available constructor", but I believe there might be a separate issue about making those messages more informative.)

lbargaoanu: There is a parameter to ShouldThrow that receives the exception.

src/UnitTests/IMappingExpression/ForCtorParam.cs
tlycken: Do you mean something like `public Dest(string foo, int bar)`? The same thing happens - the mismatching configuration is ignored, and the rest work just as if it wasn't there. The mapping works if all parameters are mapped, and doesn't if they're not. Or was that not what you were asking?

lbargaoanu: Why allow something that is clearly wrong? That's bound to create confusion eventually. With multiple parameters, I thought I overrode that one, but I misspelt, but it still matched somehow and no error anywhere.

tlycken: My primary use case for this behavior is outlined as a final remark on #1528.

lbargaoanu: Yes, but in the general case, that's a problem. Garbage in, garbage out.

tlycken: Well, I'd much rather have the configuration validator somehow report this, then, than throw an exception configuration-time. And I have no idea how to hook into that structure :P

lbargaoanu: [Here](https://github.com/AutoMapper/AutoMapper/blob/a279afeed13f2564606221bd9fa46ae32e691f56/src/AutoMapper/TypeMap.cs#L149).

jbogard:
Yeah, this should throw an AutoMapperConfigurationException. I assume that you intend your configuration to actually work, and since ctor params are string-only, it's room for error.

tlycken:
OK, I guess I'll have to solve my use case some other way :)

I won't be able to do anything more about this now, though - I'm leaving my desk in a few minutes, and will be afk for about four weeks. If you feel like taking the code in this PR, change it to behave the way you want, feel free. Otherwise I'll try to remember getting back on this mid-August.

tlycken:
In the interest of improving this, I've updated the behavior to what I gather you want it to be.

I still think it would make sense to defer throwing the exceptions to validation time. There are lots of other ways I could create mappings that are invalid, but that won't throw exceptions at config time; I can either validate them using the validation helpers on the configuration object, or - if I never run that utility - I can get runtime errors. I don't see what makes constructor maps so different from that, other than that the errors might be easier to make because there is no way to get static typing.

tlycken:
The test error seems unrelated; I'll rebase to see if it goes away (seems to be related to versioning, so it could be just that it's been a while since this PR was initiated) but if not I have no idea.

jbogard:
Ahhh I think the AppVeyor builds haven't updated to the latest NuGet.

tlycken:
OK, I'll just stand by then. Ping me if you need anything else to move this forward :smile:

lbargaoanu:
@jbogard This looks good to me.

tlycken:
@jbogard fun to have this merged just after coming home from listening to your presentations at Øredev :)

PR #1537, opened 2016-07-18
Convert using to string

### LINKED ISSUE #1529: Possible regression: AutoMapper 5.0.2 calls ToString() where 4.x didn't
During my attempt to update to 5.0.2 I came across what may be a regression. 

A combination of `cfg.CreateMap<T, string>().ProjectUsing(...)` and `T` being a base type of the member that is actually mapped causes the problem: AutoMapper 5 calls `ToString()` instead of the configured projection. AutoMapper 4, on the other hand, uses the projection.

The following test illustrates the problem:

    [Fact]
    public void Custom_string_mapping_is_used()
    {
        var config = new MapperConfiguration(
            cfg =>
            {
                cfg.CreateMap<IId, string>().ProjectUsing(id => id.Serialize());
            });

        var output = mapper.Map<Destination>(new Source { TheId = new Id { Prefix = "p", Value = "v" } });

        Assert.Equal("p_v", output.TheId);
    }

    class Source
    {
        public Id TheId { get; set; }
    }

    class Destination
    {
        public string TheId { get; set; }
    }

    interface IId
    {
        string Serialize();
    }

    class Id : IId
    {
        public string Prefix { get; set; }

        public string Serialize()
        {
            return Prefix + "_" + Value;
        }
    }

### Observations

- With AutoMapper 5, `output.TheId` is the FQN of the `Id` class, indicating that AutoMapper calls `ToString()`.
- When the interface `IId` is removed and the class `Id` is used directly, the mapping behaves as expected. This is however not flexible enough for my case, because I have many implementations of that interface.
- It doesn't matter whether `ProjectUsing` or `ConvertUsing` is used.

lbargaoanu:
fixes #1529
use ResolveTypeMap instead of GetTypeMap
special case CreateMissingTypeMaps to avoid the overhead on the normal path

### DIFF
--- a/src/AutoMapper/MapperConfiguration.cs
+++ b/src/AutoMapper/MapperConfiguration.cs
@@ -109,39 +109,43 @@ namespace AutoMapper
         public TypeMap ResolveTypeMap(TypePair typePair)
         {
             var typeMap = _typeMapPlanCache.GetOrAdd(typePair, _getTypeMap);
-
-            _typeMapPlanCache.AddOrUpdate(typePair, typeMap, (tp, tm) =>
+            if(Configuration.CreateMissingTypeMaps && typeMap != null && typeMap.MapExpression == null)
             {
-                if (tm != null)
-                    lock (tm)
-                        if (tm.Sealed == false)
-                            tm.Seal(_typeMapRegistry, this);
-                return tm;
-            });
-
+                lock(typeMap)
+                {
+                    typeMap.Seal(_typeMapRegistry, this);
+                }
+            }
             return typeMap;
         }
 
-        private TypeMap GetTypeMap(TypePair pair)
+        private TypeMap GetTypeMap(TypePair initialTypes)
         {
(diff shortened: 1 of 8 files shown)


src/AutoMapper/MapperConfiguration.cs
```diff
@@ -109,39 +109,43 @@ public TypeMap ResolveTypeMap(Type sourceType, Type destinationType)
...
                     return typeMap;
+                }
+                if(!CoveredByObjectMap(initialTypes))
+                {
+                    typeMap = FindConventionTypeMapFor(types);
+                    if(typeMap != null)
+                    {
+                        return typeMap;
+                    }
+                }
```
jbogard: btw does this make more sense now? I broke it out instead of a ginormous conditional

lbargaoanu: Of course :)

PR #1538, opened 2016-07-18
Map to existing object

### LINKED ISSUE #1486: Problem with mapping DTOs onto their POCO class in 5.02
I've been using the following code successfully with Automapper to do entity updates in my service layer:
```
        public async Task<UserDto> UpdateAsync(UserDto dto)
        {
            var existingAccount = await _repo.Get()
                .Where(u => u.Id == dto.Id)
                .Include(u => u.Permissions)
                .FirstOrDefaultAsync();

            if (existingAccount != null)
            {
                Mapper.Map(dto, existingAccount);

                await _repo.SaveChangesAsync();
                return Mapper.Map<UserDto>(existingAccount);
            }
            return null;
        }
```
Basically I get the entity from the database, map the DTO onto it and then save the changes

I've just updated to 5.02 which fixed some issues I had but now the code above is throwing the following error when I call save changes:

> "The instance of entity type 'UserPermissions' cannot be tracked because another instance of this type with the same key is already being tracked. When adding new entities, for most key types a unique temporary key value will be created if no key is set (i.e. if the key property is assigned the default value for its type). If you are explicitly setting key values for new entities, ensure they do not collide with existing entities or temporary values generated for other new entities. When attaching existing entities, ensure that only one entity instance with a given key value is attached to the context."

All my update methods seem to work fine unless they have a one to one relationship with another entity. For example User's Permissions and Customer's Licence

I'm not sure if there is an intended change that broke this workflow or if its a bug...

lbargaoanu:
if a destination was passed, map to destination subobjects by default, no UseDestinationValue necessary
fixes #1486

### DIFF
--- a/src/AutoMapper/Execution/TypeMapPlanBuilder.cs
+++ b/src/AutoMapper/Execution/TypeMapPlanBuilder.cs
@@ -284,9 +284,22 @@ namespace AutoMapper.Execution
                 getter = destMember;
             }
 
-            var destValueExpr = propertyMap.UseDestinationValue
-                ? getter
-                : Default(propertyMap.DestinationPropertyType);
+            Expression destValueExpr;
+            if(propertyMap.UseDestinationValue)
+            {
+                destValueExpr = getter;
+            }
+            else
+            {
+                if(_initialDestination.Type.IsValueType())
+                {
+                    destValueExpr = Default(propertyMap.DestinationPropertyType);
+                }
+                else
+                {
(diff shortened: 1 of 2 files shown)


jbogard:
Someone at some point had a GH issue where the destination getter was always accessed, and they had some kind of issue where if the getter was accessed without a value it would throw. Does this PR mean that the getter would always get used? Or only if they actually passed in a value?

lbargaoanu:
If there is a value in the initial call, for the parent, the getter is accessed.

PR #1554, opened 2016-07-22
Add a null guard instead of rewriting each node

### LINKED ISSUE #1550: Update to 5.0.2 broke existing maps
After updating automapper I'm getting
```
Expression of type 'System.Nullable`1[System.DateTime]' cannot be used for return type 'System.DateTime'
```

### LINKED ISSUE #1552: 5.0.2 Can't create map where source is DateTime.Date
- AutoMapper 5.0.2
- Visual Studio 2015
- .Net Framework 4.6

When trying to create a mapping for a nullable DateTime when the source property is a DateTime.Date type, it throws an exception:

> System.ArgumentException : Property 'System.DateTime Date' is not defined for type 'System.Nullable`1[System.DateTime]'

This worked fine in version 4.2.x.

Sample code showing how to reproduce the issue and a work-around:
`using System;
using NUnit.Framework;

namespace AutoMapper.Tests
{
    public class Employee
    {
        public DateTime? FiredDate { get; set; }
        public DateTime HiredDate { get; set; }
    }

    public class HumanResourcesService
    {
        private static readonly IMapper Mapper;

        static HumanResourcesService()
        {
            var config = new MapperConfiguration(m =>
            {
                // works fine (well, unit test fails, but it doesn't blow up creating the map)
                //m.CreateMap<Employee, Employee>().ForMember(d => d.FiredDate, o => o.MapFrom(s => s.HiredDate));

                // System.ArgumentException : Property 'System.DateTime Date' is not defined for type 'System.Nullable`1[System.DateTime]'
                //m.CreateMap<Employee, Employee>().ForMember(d => d.FiredDate, o => o.MapFrom(s => s.HiredDate.Date));

                // work-around
                m.CreateMap<Employee, Employee>()
                    .ForMember(d => d.FiredDate, o => o.Ignore())
                    .AfterMap((s, d) => d.FiredDate = s.HiredDate.Date);
            });

            Mapper = config.CreateMapper();
        }

        public Employee Fire(Employee badEmployee)
        {
            return Mapper.Map<Employee>(badEmployee);
        }
    }

    [TestFixture]
    public class HumanResourcesServiceTest
    {
        [Test]
        public void ATest()
        {
            var employee = new Employee() { HiredDate = DateTime.Now };

            var humanResourcesService = new HumanResourcesService();
            var firedEmployee = humanResourcesService.Fire(employee);

            Assert.That(employee.HiredDate.Date, Is.EqualTo(firedEmployee.FiredDate.Value));
        }
    }
}
`

Also, no idea why my sample code formatting above keeps getting biffed.  Sorry 'bout that.

lbargaoanu:
Fixes #1552. Closes #1550.
I think the root cause of the nullable issues was the fact that the node type was changed in flight, so to speak, without knowing it's safe to do so. It should be safe for the initial expression.
@jbogard This assumes the member accesses visited are really executed, so when any is null, return null. I think this is correct after the new checks you made recently. Of course it could be made more complicated to detect whether the members were really accessed, but with those checks in place, I see no point. If you want to get rid of the checks, the story changes.

### DIFF
--- a/src/AutoMapper/Execution/DelegateFactory.cs
+++ b/src/AutoMapper/Execution/DelegateFactory.cs
@@ -51,17 +51,6 @@ namespace AutoMapper.Execution
             return lambda;
         }
 
-        public static Expression IfNotNullExpression(MemberExpression member, Type destinationType)
-        {
-            var returnType = destinationType.IsNullableType() && destinationType.GetTypeOfNullable() == member.Type
-                ? destinationType
-                : member.Type;
-            if (member.Expression != null && !member.Expression.Type.IsValueType())
-                return Condition(Equal(member.Expression, Default(member.Expression.Type)),
-                Default(returnType), ExpressionExtensions.ToType(member, returnType));
-            return member;
-        }
-
         public LateBoundCtor CreateCtor(Type type)
         {
             var ctor = _ctorCache.GetOrAdd(type, _generateConstructor);
--- a/src/AutoMapper/ExpressionExtensions.cs
+++ b/src/AutoMapper/ExpressionExtensions.cs
@@ -125,9 +125,10 @@ namespace AutoMapper
                     node.NodeType == ExpressionType.MemberAccess 
                     || (node.NodeType == ExpressionType.Call && ((MethodCallExpression)node).Arguments.Count == 0));
             }
-            if (node != null && node.NodeType == ExpressionType.Parameter)
-                return new IfNotNullVisitor(destinationType).Visit(expression);
-
+            if(node != null && node.NodeType == ExpressionType.Parameter)
+            {
+                return new IfNotNullVisitor().VisitRoot(expression, destinationType);
+            }
             return expression;
         }
 
@@ -142,22 +143,25 @@ namespace AutoMapper
 
         internal class IfNotNullVisitor : ExpressionVisitor
         {
-            private readonly Type _destinationType;
-            private readonly HashSet<MemberExpression> _alreadyUpdated = new HashSet<MemberExpression>();
+            private Expression nullConditions = Constant(false);
 
-            public IfNotNullVisitor(Type destinationType)
+            public Expression VisitRoot(Expression node, Type destinationType)
             {
-                _destinationType = destinationType;
+                var returnType = Nullable.GetUnderlyingType(destinationType) == node.Type ? destinationType : node.Type;
+                var expression = base.Visit(node);
+                var checkNull = Condition(nullConditions, Default(returnType), ToType(expression, returnType));
+                return checkNull;
             }
 
             protected override Expression VisitMember(MemberExpression node)
             {
-                if(_alreadyUpdated.Contains(node))
+                var returnNode = base.VisitMember(node);
+                if(node.Expression == null || node.Expression.Type.IsValueType())
                 {
-                    return base.VisitMember(node);
+                    return returnNode;
                 }
-                _alreadyUpdated.Add(node);
-                return Visit(DelegateFactory.IfNotNullExpression(node, _destinationType));
+                nullConditions = Or(nullConditions, Equal(node.Expression, Constant(null, node.Expression.Type)));
+                return returnNode;
             }
         }
 
--- a/src/UnitTests/Bug/NullableDateTime.cs
+++ b/src/UnitTests/Bug/NullableDateTime.cs
@@ -1,10 +1,73 @@
 ﻿using System;
+using System.Linq;
 using Should;
 using AutoMapper;
 using Xunit;
 
 namespace AutoMapper.UnitTests.Bug
 {
+    public class NullableDateTimeMapFromArray : AutoMapperSpecBase
+    {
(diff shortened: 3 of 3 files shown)


jbogard:
Resulting expression:

```
(src, dest, ctxt) =>
{
    if (src == null)
    {
        null;
    }
    else
    {
        var destination = dest ?? new CustomerDTO();
        try
        {
            destination.AddressCity = ((false | (src == null)) | (src.Address == null)) ? null : src.Address.City;
        }
        catch (Exception ex)
        {
            throw new AutoMapperMappingException("Error mapping types.", ex, AutoMapper.TypePair, AutoMapper.TypeMap, AutoMapper.PropertyMap);
        }

        try
        {
            destination.WorkAddresses = if ((false | (src == null)) ? null : src.WorkAddresses == null)
            {
                return ((dest == null) ? null : destination.WorkAddresses != null)
                    ? (dest == null) ? null : destination.WorkAddresses
                    : new List<AddressDTO>();
            }

            var d = ((dest == null) ? null : destination.WorkAddresses != null)
                ? (dest == null) ? null : destination.WorkAddresses
                : new List<AddressDTO>();
            var enumerator = false | (src == null)) ? null : src.WorkAddresses.GetEnumerator(;
            while (true)
            {
                if (enumerator.MoveNext())
                {
                    var item = enumerator.Current;
                    d.Add(ctxt.Map(item, null));
                }
                else
                {
                    break;
                }
            }

            return d;;
        }
        catch (Exception ex)
        {
            throw new AutoMapperMappingException("Error mapping types.", ex, AutoMapper.TypePair, AutoMapper.TypeMap, AutoMapper.PropertyMap);
        }

        try
        {
            destination.Addresses = if ((false | (src == null)) ? null : src.Addresses == null)
            {
                new AddressDTO[0];
            }
            else
            {
                var count = false | (src == null)) ? null : src.Addresses.Count(;
                var destinationArray = new AddressDTO[count];
                var destinationArrayIndex = 0;
                var sourceArrayIndex = 0;
                while (true)
                {
                    if (sourceArrayIndex < (false | (src == null)) ? null : src.Addresses.Length)
                    {
                        var item = (false | (src == null)) ? null : src.Addresses[sourceArrayIndex];
                        destinationArray[destinationArrayIndex++] = ctxt.Map(item, null);
                        sourceArrayIndex++;
                    }
                    else
                    {
                        break;
                    }
                }

                return destinationArray;
            };
        }
        catch (Exception ex)
        {
            throw new AutoMapperMappingException("Error mapping types.", ex, AutoMapper.TypePair, AutoMapper.TypeMap, AutoMapper.PropertyMap);
        }

        try
        {
            destination.HomeAddress = ctxt.Map((false | (src == null)) ? null : src.HomeAddress, (dest == null) ? null : destination.HomeAddress);
        }
        catch (Exception ex)
        {
            throw new AutoMapperMappingException("Error mapping types.", ex, AutoMapper.TypePair, AutoMapper.TypeMap, AutoMapper.PropertyMap);
        }

        try
        {
            destination.Address = (false | (src == null)) ? null : src.Address;
        }
        catch (Exception ex)
        {
            throw new AutoMapperMappingException("Error mapping types.", ex, AutoMapper.TypePair, AutoMapper.TypeMap, AutoMapper.PropertyMap);
        }

        try
        {
            destination.Name = (false | (src == null)) ? null : src.Name;
        }
        catch (Exception ex)
        {
            throw new AutoMapperMappingException("Error mapping types.", ex, AutoMapper.TypePair, AutoMapper.TypeMap, AutoMapper.PropertyMap);
        }

        try
        {
            destination.Id = (false | (src == null)) ? default(int) : src.Id;
        }
        catch (Exception ex)
        {
            throw new AutoMapperMappingException("Error mapping types.", ex, AutoMapper.TypePair, AutoMapper.TypeMap, AutoMapper.PropertyMap);
        }

        return destination;
    }
}
```

```
(src, dest, ctxt) =>
{
    if (src == null)
    {
        null;
    }
    else
    {
        var destination = dest ?? new AddressDTO();
        try
        {
            destination.Country = (false | (src == null)) ? null : src.Country;
        }
        catch (Exception ex)
        {
            throw new AutoMapperMappingException("Error mapping types.", ex, AutoMapper.TypePair, AutoMapper.TypeMap, AutoMapper.PropertyMap);
        }

        try
        {
            destination.City = (false | (src == null)) ? null : src.City;
        }
        catch (Exception ex)
        {
            throw new AutoMapperMappingException("Error mapping types.", ex, AutoMapper.TypePair, AutoMapper.TypeMap, AutoMapper.PropertyMap);
        }

        try
        {
            destination.Id = (false | (src == null)) ? default(int) : src.Id;
        }
        catch (Exception ex)
        {
            throw new AutoMapperMappingException("Error mapping types.", ex, AutoMapper.TypePair, AutoMapper.TypeMap, AutoMapper.PropertyMap);
        }

        return destination;
    }
}
```

jbogard:
That's from the BasicFlattening test, which is the same flattening test from our benchmark.

Issue #1357, opened 2016-06-23
AfterMap()/BeforeMap() getting invoked multiple times


brittonbeckham:
I'm currently on 4.1.1.

I have had this happen on two separate projects in two separate occasions. The first time I thought I was going nuts and just found a different way to do what I needed, but now it is happening once again and. I've have spend time looking through the Github release notes and have not seen and closed Issues that resolve `AfterMap()` problems, but I would like to ask here to see if this was a known issue for 4.1.1 or not. I can recreate this issue every time I use `AfterMap()` on any of my maps. My current work around is to put the `AfterMap()` logic in `ConstructUsing()`. This is happening with both `AfterMap ()` and `BeforeMap()`

Is this even a known issue?
Is this resolved in 4.2.0+ ?

brittonbeckham:
Here is my code for reference; `AfterMap()` is getting invoked 10 times on this Map:

```c#
Mapper.CreateMap<StorefrontDealCacheItem, LiveStorefrontDeal>()
	.ForMember(dest => dest.StorefrontDealStatus, opt => opt.ResolveUsing(src => StorefrontDealHelpers.CalculateStorefrontDealStatus(src.StartDate, src.EndDate)))
	.AfterMap((src, dest) => {

		var likeCountTask = _cahcedDealValueAgent.GetDealLikeCount(src.DealId);
		var soldQuantitiesTask = _cahcedDealValueAgent.GetDealSoldQuantity(src.DealId);
		var optionsTask = _cahcedDealValueAgent.GetOptionsAndVariantSoldQuantities(new[] { src.DealId });
		var sellerShipTimesTask = _cahcedDealValueAgent.GetSellerShipTime(src.SellerAccountId);

		//get all cached values async
		Task.WaitAll(soldQuantitiesTask, optionsTask, sellerShipTimesTask);

		var likeCount = likeCountTask.Result;
		var soldQuantity = soldQuantitiesTask.Result;
		var optionsCacheItem = optionsTask.Result.OptionsCacheItems.GetValueOrDefault(src.DealId);
		var sellerShipTimeDto = sellerShipTimesTask.Result;

		dest.LikeCount = likeCount;
		dest.IsLikelySellOut = StorefrontDealHelpers.IsLikelySellOut(src.Quantity, likeCount);
		dest.SoldQuantity = GetSoldQuantityWithAb(src, soldQuantity).Result;
[...]

PR #1555, opened 2016-07-22
Don't allow duplicate After/Before Map registrations

lbargaoanu:
Because it's easy &trade;
@jbogard Just close it if you think this needs more consideration.


lbargaoanu:
Fixes #1357.

jbogard:
Nah, this works for me. Anyways, if you're doing AfterMap combined with inheritance, that already seems complicated.

PR #1573, opened 2016-07-29
Base class for simple object mappers that don't want to use expressions.

lbargaoanu:
It's true that anybody can write this, but I thought it might be nice to have something built in.

### DIFF
--- a/src/AutoMapper/IObjectMapper.cs
+++ b/src/AutoMapper/IObjectMapper.cs
@@ -1,11 +1,15 @@
+using System;
 using System.Linq.Expressions;
+using System.Reflection;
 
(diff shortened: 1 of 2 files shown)


jbogard:
What's the difference between this and a type converter?

lbargaoanu:
IsMatch is more flexible. But anyway, what's the difference in general? I wasn't sure whether to add the extra parameters so I didn't.

lbargaoanu:
I think it didn't feel like an interface with all those parameters, but they are input, so I guess they do no harm. The main point was that probably no one would implement the interface as it is. But if it still feels redundant, just close it.

PR #1577, opened 2016-08-01
Resolve and then map if needed

lbargaoanu:
I started from that test...

### DIFF
--- a/src/AutoMapper/ConstructorMap.cs
+++ b/src/AutoMapper/ConstructorMap.cs
@@ -53,6 +53,7 @@ namespace AutoMapper
         }
 
         public Expression BuildExpression(
+            TypeMapRegistry typeMapRegistry,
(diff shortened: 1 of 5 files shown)


lbargaoanu:
I think we should pass the TypeMapPlan builder around, instead all those parameters. We would just call methods on that object instead of the static method with many parameters.

lbargaoanu:
I did it for constructor mapping. Better I think :)

jbogard:
Started from which test?

lbargaoanu:
When_renaming_class_constructor_parameter in Constructors.cs

lbargaoanu:
@jbogard Can we merge this? Most fixes want to touch TypeMapPlanBuilder. If you like the idea of refactoring the mappers interface, I can do that in another PR.

Issue #1596, opened 2016-08-04
NUll Reference Exception with 5.0.2.0


jogibear9988:
I've the following code:

Mapper.Initialize(cfg => {
                    cfg.CreateMap(source.GetType(), destination.GetType());
                });

and I got this null ref exception (used reflector for debugging):
<img width="911" alt="bildschirmfoto 2016-08-05 um 00 42 37" src="https://cloud.githubusercontent.com/assets/364896/17420819/8e535374-5aa5-11e6-8711-9643ff0ca491.png">

jbogard:
Why are you calling CreateMap with runtime objects?

jogibear9988:
?? I did so in the old Version, where it worked. I know them firstly at runtime, not at build time.

lbargaoanu:
A repro would help. Make a gist that we can execute and see fail.

jogibear9988:
This Example will cause the error: (you need Castle.Core(3.3.3) and AutoMapper(5.0.2))


    public class aa
    {
        public string b1 { get; set; }
    }
    
    class Program
    {
        static void Main(string[] args)
        {
            var a1 = new aa();
            var proxy = new ProxyGenerator().CreateClassProxy(a1.GetType(), new[]{typeof (INotifyPropertyChanged)}, new NotifyPropertyChangedInterceptor(a1.GetType().FullName));

            Mapper.Initialize(cfg => {
                cfg.CreateMap(a1.GetType(), proxy.GetType());
            });
        }

        public class NotifyPropertyChangedInterceptor : IInterceptor
        {
            private readonly string typeName;
            private PropertyChangedEventHandler subscribers = delegate { };

            public NotifyPropertyChangedInterceptor(string typeName)
            {
                this.typeName = typeName;
            }

            public void Intercept(IInvocation invocation)
            {
            }
        }
    }

jogibear9988:
Fixed with: https://github.com/AutoMapper/AutoMapper/pull/1597

lbargaoanu:
Well yes, but I think the problem is with Castle, not with AutoMapper. A parameter name cannot be null.

lbargaoanu:
A workaround is to disable constructor mapping.

jogibear9988:
As you see it can be, but with my pull req it's fixed.

lbargaoanu:
No, it cannot :) A parameter has a name. I think the bug is that it returns null, not that the check is missing. But I guess it doesn't matter much.

[...]

PR #1597, opened 2016-08-05
bugfix empty name for example generated by Castle Dnyamic Proxy

jogibear9988:


jogibear9988:
I've only changed line 92, but I created the edit via github, maybe cause of that the other lines are changed

jogibear9988:
will this get merged?

lbargaoanu:
There is no need to explicitly check for null. Using the static Equals would do.

jbogard:
There's no test here, is it possible to write a test to show how this behavior should work? Also, I'm not entirely sure this addresses your issue, since you're calling CreateMap with runtime values.

jogibear9988:
i've tested it, with my null check it works

jogibear9988:
look at issue #1596, you could use my castle example for a test

PR #1601, opened 2016-08-08
Map from null interface

### LINKED ISSUE #1598: Mapping from Interface to concrete class no longer works as expected
In Automapper 4.2.1 mapping from an Interface to a concrete class worked so that if a `null` property defined as an Interface would map to a concrete class as `null`.

In Automapper 5.0.2 a null interface property maps to a default constructed concrete class. 

This is a breaking issue for my application. I have a lot of interface properties on source classes mapping to concrete class properties in the destination class. 

**Expectation:**
I would expect that null properties on the source would map to null properties on the destination.

**To test:**
[Gist](https://gist.github.com/jgilm/f73fb5ac2ea96aa0d7d8257ca1035f69)
- Make new console app, paste in gist
- install Automapper 4.2.1 package and run console app. Test passes. 
- install Automapper 5.0.2 package and run console app. Test fails. 

_Although test is a console app, issue also occurs in a UWP app._

4.2.1 output:
image

5.0.2 output:
image

lbargaoanu:
Fixes #1598.

### DIFF
--- a/src/AutoMapper/Execution/DelegateFactory.cs
+++ b/src/AutoMapper/Execution/DelegateFactory.cs
@@ -66,8 +66,7 @@ namespace AutoMapper.Execution
 
         public static Expression GenerateConstructorExpression(Type type)
         {
-            //handle valuetypes
-            if (!type.IsClass())
+            if(type.IsValueType())
             {
                 return Convert(New(type), typeof(object));
             }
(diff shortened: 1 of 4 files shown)


lbargaoanu:
Now, if the source is abstract and the value is not mapped, you will get null even with allow null false. So maybe have that null logic from the map in the assignable mapper? Or just wait for someone to complain.

jbogard:
Yeah, let's wait for someone to complain.

PR #1602, opened 2016-08-08
Don't fail when the parameter name is null

### LINKED ISSUE #1596: NUll Reference Exception with 5.0.2.0

lbargaoanu:
Pedantic? :)
There is no need to explicitly check for null. Fixes #1596. Closes #1597.

### DIFF
--- a/src/AutoMapper/TypeMap.cs
+++ b/src/AutoMapper/TypeMap.cs
@@ -89,7 +89,7 @@ namespace AutoMapper
 
         public bool ConstructorParameterMatches(string destinationPropertyName)
         {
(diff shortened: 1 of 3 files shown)


jbogard:
lol probably. How can we have a test for this?

lbargaoanu:
Too much work I would say. You'd have to dynamically generate a type with a constructor that has a parameter without name. It hurts just to say it :)

lbargaoanu:
Or reference Castle :)

lbargaoanu:
Or hack a little test :) Now, _this_ is pedantic...

PR #1621, opened 2016-08-16
TypePair ignore EntityFramework proxy types when getting type from source instance

DaveOMac:
When specifying a source instance to be mapped to the IMapper.Map the source.GetType() takes precedence over the specified sourceType. 

If source is an instance of an EntityFramework proxy object, CreateMissingTypeMaps is set to true, and there is a defined map for the POCO the proxy is sub classing; the custom map is not used as the sourceType is the EF proxy not the POCO type.

### DIFF
--- a/src/AutoMapper/TypeExtensions.cs
+++ b/src/AutoMapper/TypeExtensions.cs
@@ -253,5 +253,27 @@ namespace AutoMapper
         {
             return type.GetRuntimeField(name);
         }
+
+        public static bool IsEntityFrameworkProxyType(this Type type)
+        {
+            const string entityFrameworkDynamicProxiesNamespace = "System.Data.Entity.DynamicProxies";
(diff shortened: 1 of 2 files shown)


TylerCarlson1:
instead of  `cfg.CreateMissingTypeMaps = true;` you can use
`cfg.AddConditionalObjectMapper().Where((s, d) => s.Namespace != "System.Data.Entity.DynamicProxies");` and dynamic proxies won't be mapped.

DaveOMac:
@TylerCarlson1 thanks that's awesome!

Modified slightly to:

`cfg.AddConditionalObjectMapper().Conventions.Add(typePair => typePair.SourceType.Namespace != "System.Data.Entity.DynamicProxies");`

for AutoMapper 5.1.1

I must admit I did not know what AddConditionalObjectMapper() did or how it behaved before making this pull request will close.

*note to self: read more documentation*

PR #1625, opened 2016-08-17
[WIP] Inheritance baked into Expressions

TylerCarlson1:
No longer need map requests because conditions are handled in the expression generated.

Idea is to take out conditions for include when doing collections.
Also to remove the need for GetType to make mappings faster in general, but there's tests that fail if I try.


TylerCarlson1:
It still needs some work for the collection using the expression.  The mapper function does the wrapping and casting of the MappingExpression now, but what I might need to do is push this functionality into the  TypeMap itself to generate this expression, and then it can be used in CollectionMapping instead of having to call Mapper.Map and speed it up.

TylerCarlson1:
@jbogard Weird it didn't build.  And it seems like the other PR's are failing builds at the very beginning.

Don't know why version number 5.2.0-alpha.* is invalid, but I got it until I updated VS.

jbogard:
Yeah, I'm trying out SemVer 2.0 that is available in the latest NuGet.


Hmmmmm

jbogard:
Merge conflicts :(

TylerCarlson1:
Still WIP as I need to re-order things to make Inheritance get the performance gain when mapping collections.

TylerCarlson1:
Got it working with passing tests, but it looks like for normal mappings performance took a hit.

Also I have to change type maps where source/destination differ from generics, because it will compile every map right now and is super slow.  Maybe even cache them in MaperFuncs class when compile and the map cache plan will use actual object types vs Generics types.  There will be more type pairs to compile, but once they are set they don't need to be re-compiled.  This would probably be best for things like Proxies on EF that would have to go through the invoke type map and cast process.

Or maybe I'm wrong about what it actually does and is efficient right now.

PR #1629, opened 2016-08-17
Only a global AllowNullCollections

### LINKED ISSUE #1618: Profile.AllowNullCollections ignored?
The `Profile.AllowNullCollections` property seems to be being missed in 5.0.2 and beyond. Moving the code out of the `Configure` method into a constructor results in the same behavior.

```
class Program
{
    {
        var mapper = new MapperConfiguration(cfg =>
        {
            cfg.AddProfile(new TestMappingProfile());
        }).CreateMapper();

        var isNull = mapper.Map<Source, Destination>(new Source()).Ints == null;
        // "true" in 4.2.1
        // "false" in 5.0.2+
    }
}

class TestMappingProfile : Profile
{
    {
        AllowNullCollections = true;
        CreateMap<Source, Destination>();
    }
}

class Source
{
    public IEnumerable<int> Ints { get; set; }
}

class Destination
{
    public IEnumerable<int> Ints { get; set; }
}
```

lbargaoanu:
Closes #1618.
And if this makes sense, AllowNullDestinationValues too. Or just wait for someone to complain.

### DIFF
--- a/src/AutoMapper/Configuration/MapperConfigurationExpression.cs
+++ b/src/AutoMapper/Configuration/MapperConfigurationExpression.cs
@@ -15,6 +15,7 @@ namespace AutoMapper.Configuration
             _profiles.Add(this);
         }
 
+        public bool AllowNullCollections { get; set; }
         public IEnumerable<Profile> Profiles => _profiles;
         public Func<Type, object> ServiceCtor { get; private set; } = ObjectCreator.CreateObject;
 
--- a/src/AutoMapper/IConfiguration.cs
+++ b/src/AutoMapper/IConfiguration.cs
@@ -5,6 +5,7 @@ namespace AutoMapper
 
     public interface IConfiguration : IProfileConfiguration
     {
+        bool AllowNullCollections { get; set; }
         Func<Type, object> ServiceCtor { get; }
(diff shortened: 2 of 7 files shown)


jbogard:
Regarding the comment on #1618 I wonder if it would be better to do the thing where we apply the global configuration only if the profile configuration isn't set?

lbargaoanu:
I don't know. Having these things per profile too seems complicated and error prone to me. We had similar issues before.

jbogard:
What's the point of profiles then? Just naming conventions?

TylerCarlson1:
I like the flexibility of it, even if you could possibly miss it.  And who knows if someone next version will go HEY where did my Profile settings go?  I needed that for ____.

lbargaoanu:
Profiles are used to split the configuration per module. Then you gather it all in one place. If you look at it this way, I think global settings make sense. You can have multiple configurations if you want.

lbargaoanu:
But if some things work per profile, naming conventions for example, then I guess all should work per profile :) (and in the same way) Because it used to be the only way to separate configuration. Right now separate configurations seem to me a better way to do that, but I guess that would break things.

PR #1632, opened 2016-08-18
Just clear the destination everytime we don't create it ourselves

### LINKED ISSUE #1630: Mapping onto existing object does no longer clear existing items from List [5.1.1]
```
class SomethingDto {}
class ADto {
    public List<SomethingDto> Somethings { get; } = new List<SomethingDto> { new SomethingDto() };
}

class Something {}
class ACls {
    public List<Something> Somethings { get; protected set; } = new List<Something> { new Something() };
}
```

```
Mapper.CreateMap<ADto, ACls>();
Mapper.CreateMap<SomethingDto, Something>();
```

```
var existing = new ACLs();
Mapper.Map(new ADto(), existing);
```

Observed: `existing.Somethings` will now have two Something instances.
Expected: `existing.Somethings` will now have one Something instance.

Workaround: `.BeforeMap((_, dst) => dst.Somethings.Clear());`
However, that does not make for a happy panda :)

### LINKED ISSUE #1645: Dictionary mappings no longer handle non-empty dictionary destinations in 5.1.1
Two objects have dictionary properties. When a source object is mapped onto a destination object, the dictionary values should be replaced. Unfortunately, it seems that `Insert` is being called on the destination object dictionary without clearing it first, which results in an exception. That's (apparently) how the 4.x codeline works; maybe this is just a breaking change in 5.x?

Repro:

```csharp
public class Source
{
  public IDictionary<string, string> Values { get; private set; } = new Dictionary<string, string>();
}

public class Destination
{
  public IDictionary<string, object> Values { get; private set; } = new Dictionary<string, object>();
}

[Fact]
public void DictionaryTest()
{
  var source = new Source();
  source.Values["first"] = "1";
  source.Values["second"] = "2";

  var dest = new Destination();
  dest.Values["first"] = "one";

  var mapper = new MapperConfiguration(cfg => cfg.CreateMap<Source, Destination>()).CreateMapper();
  mapper.Map(source, dest);

  Assert.Equal("1", dest.Values["first"]);
  Assert.Equal("2", dest.Values["second"]);
}
```

The `mapper.Map` line fails in 5.1.1 with "An item with the same key has already been added." That's because the "first" key already exists in the destination dictionary, which should be cleared(?) first.

I've found a workaround by making a generic merging dictionary mapper...

```csharp
public class DictionaryConverter<TSourceKey, TSourceValue, TDestKey, TDestValue> :
  ITypeConverter<IDictionary<TSourceKey, TSourceValue>, IDictionary<TDestKey, TDestValue>>
{
  public IDictionary<TDestKey, TDestValue> Convert(IDictionary<TSourceKey, TSourceValue> source, IDictionary<TDestKey, TDestValue> destination, ResolutionContext context)
  {
    if (source == null)
    {
      return null;
    }

    var valuesToMerge = new Dictionary<TDestKey, TDestValue>();
    foreach (var sourceKey in source.Keys)
    {
      var destKey = context.Mapper.Map<TDestKey>(sourceKey);
      var destValue = context.Mapper.Map<TDestValue>(source[sourceKey]);
      valuesToMerge[destKey] = destValue;
    }

    if (destination != null)
    {
      destination.Clear();
    }

    return valuesToMerge;
  }
}
```

I can then register it as an open generic:

```csharp
var mapper = new MapperConfiguration(cfg =>
{
  cfg.CreateMap(typeof(IDictionary<,>), typeof(IDictionary<,>))
     .ConvertUsing(typeof(MergingDictionaryConverter<,,,>));
}).CreateMapper();
```

But doing this sort of hoop-jumping makes me feel like I'm missing something obvious. Maybe I was just taking advantage of an inadvertent/incorrect behavior before? Or am I actually missing something? (Or is this a bug?)

lbargaoanu:
Fixes #1630. Fixes #1645.

### DIFF
--- a/src/AutoMapper/Mappers/CollectionMapper.cs
+++ b/src/AutoMapper/Mappers/CollectionMapper.cs
@@ -23,56 +23,25 @@ namespace AutoMapper.Mappers
             var newExpression = Variable(newExpressionValue.Type, "collectionDestination");
             var sourceElementType = TypeHelper.GetElementType(sourceExpression.Type);
             var itemParam = Parameter(sourceElementType, "item");
+
             var itemExpr = mapItem(typeMapRegistry, configurationProvider, propertyMap, sourceExpression.Type, passedDestination.Type, itemParam, contextExpression);
 
-            var blockExpressions = new List<Expression>();
-            var blockVariables = new List<ParameterExpression> { newExpression, passedDestination };
-            Expression destination;
             var destinationElementType = itemExpr.Type;
             var destinationCollectionType = typeof(ICollection<>).MakeGenericType(destinationElementType);
             var clearMethod = destinationCollectionType.GetDeclaredMethod("Clear");
-            if(passedDestination.Type.IsCollectionType())
-            {
-                if(propertyMap == null)
-                {
-                    destination = newExpression;
-                    blockExpressions.Add(IfThenElse(NotEqual(passedDestination, Constant(null)),
-                        Call(passedDestination, clearMethod),
-                        Assign(destination, newExpression)
-                        ));
-                }
-                else if(propertyMap.UseDestinationValue)
-                {
-                    destination = passedDestination;
-                    blockExpressions.Add(Call(passedDestination, clearMethod));
-                }
-                else
-                {
-                    destination = newExpression;
-                }
-            }
-            else
-            {
-                destination = newExpression;
-            }
-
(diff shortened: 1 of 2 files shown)


sickboy:
@lbargaoanu nice, would be great if this is included in a release soon. My head explodes trying to reason about where in my use cases I might be getting duplicate entries now..

lbargaoanu:
@jbogard Can we merge this? Because there is another issue that wants to touch the same code.

PR #1635, opened 2016-08-19
Call the condition after setting the source property value because it…

### LINKED ISSUE #1634: Conditions do not get either Source or Destination values
For the given code:
`map.ForAllMembers(source =>
            {                                                              
                source.Condition((src, dest, srcVal, destVal) =>
                {                   
                    if (srcVal == null)
                        return !(destVal == null);
                    return !srcVal.Equals(destVal);
                });
            });`

"srcVal" and "destVal" are always the default value for their respective property. I am therefore unable to do conditional mappings using this "ForAllMembers" context as i have no concrete way of being able to retrieve and compare what my Source and Destination values are.
automapper issue

Note: I have walked through this including when it is for the "Colour" property. Both srcVal and destVal are the default for that type (it is a struct, so all internal values are default 0)

lbargaoanu:
… needs it
Fixes #1634.

### DIFF
--- a/src/AutoMapper/Execution/TypeMapPlanBuilder.cs
+++ b/src/AutoMapper/Execution/TypeMapPlanBuilder.cs
@@ -362,15 +362,6 @@ namespace AutoMapper.Execution
                 }
             }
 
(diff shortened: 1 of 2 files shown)


hmvs:
Could you please merge it?

PR #1638, opened 2016-08-19
Decimal to nullable decimal

lbargaoanu:
I believe my initial fix for a similar problem was a global fix for a local issue so I reverted it. Again the root problem is changing the type of the expression. I guess this applies everywhere IfNotNull is called, but I thought the problem is subtle enough to wait.


lbargaoanu:
If it's not clear for the test, when you map a nullable, the decimal map matches and if you pass null, it blows.

PR #1649, opened 2016-08-26
Convert the null substitute to what's needed for projections

lbargaoanu:


lbargaoanu:
I don't know what VisitConditional is for. There is no test for it. Delete? :)

jbogard:
Delete!

Issue #1664, opened 2016-09-01
Differentiation between type and nullable type


RobinHerbots:
@jbogard,

I upgraded from v4.2.1 to v5.1.1  and noticed the following change in the mapping behavior.

See code example

```
//just some snippet as example to get the idea
...
Mapper.Initialize(cfg => {
            cfg.CreateMap<bool?, int>().ConvertUsing(s => { return 5; });
            cfg.CreateMap<bool, int>().ConvertUsing(s => { return 1; });
});
...

bool testBool = true;
Mapper.Map<int>(testBool);

bool? testBool2 = true;
Mapper.Map<int>(testBool2);
```

In version 4.2.1 the above will call the nullable version in case a nullable bool is passed and the non-nullable when the non-nullable bool is passed. 

In version 5.1.1 in both cases the nullable version is called.

Is this intentional or is this a bug?  When this is intentional, what is the motivation to it?

jbogard:
Ha, no motivation, other than I hate nullables.

lbargaoanu:
What I see is that 4.2.1 doesn't distinguish between these cases either, and it makes sense, because
    
    ((bool?)true).GetType() == true.GetType()

lbargaoanu:
Anyway, what are you trying to do? Maybe there is a better way.

RobinHerbots:
@jbogard , 

Seeing at the result below for automapper 5.1.1 I think you LOVE nullables  ;-)

@lbargaoanu , 

Ok, I should have better prepared my example ...

Have a look below.  The same code but with different versions of automapper.  Also the result for Automapper 4.2.1 is a surprise to me.


```
using System;
using System.Collections.Generic;
using System.Linq;
using System.Runtime.Remoting.Messaging;
using System.Text;
using System.Threading.Tasks;
using AutoMapper;

namespace ConsoleApplication2
{
    class Program
    {
        static void Main(string[] args)
        {
            Mapper.Initialize(cfg =>
            {
                cfg.CreateMap<bool, string>().ConvertUsing(s => s ? "normal bool mapper" : "normal bool mapper"); //just to make it compile in version 4.2.1
                cfg.CreateMap<bool?, string>().ConvertUsing(s => s.HasValue ? "nullable bool mapper" : "nullable bool mapper"); //just to make it compile in version 4.2.1
            });


            bool testBool = true;
            Console.WriteLine("bool = true " + Mapper.Map<bool, string>(testBool));

            bool? testBool2 = true;
[...]

PR #1669, opened 2016-09-05
derived maps fix

RobinHerbots:
Also see issue #1664

In the MapperConfiguration.cs, we register the different typemaps.  For nullables there is a derived typemap created.  But it also overwrites an existing typemapping when present. 

This shouldn't happen as when the mapping already exist it is probably the intention of the dev to have it there.  So, I added a check to create the derived map only if it doesn't exist.

Robin

### DIFF
--- a/src/AutoMapper/MapperConfiguration.cs
+++ b/src/AutoMapper/MapperConfiguration.cs
@@ -247,10 +247,13 @@ namespace AutoMapper
             }
             foreach (var redirectedType in redirectedTypes)
             {
-                var derivedMap = FindTypeMapFor(redirectedType.Item2);
-                if (derivedMap != null)
+                if (!_typeMapPlanCache.ContainsKey(redirectedType.Item1))
                 {
-                    _typeMapPlanCache[redirectedType.Item1] = derivedMap;
+                    var derivedMap = FindTypeMapFor(redirectedType.Item2);
+                    if (derivedMap != null)
+                    {
+                        _typeMapPlanCache[redirectedType.Item1] = derivedMap;
+                    }
                 }
             }
             foreach (var derivedMap in derivedMaps.Where(derivedMap => !_typeMapPlanCache.ContainsKey(derivedMap.Item1)))


lbargaoanu:
Not the right fix. Redirected types are unrelated to the issue.

RobinHerbots:
@lbargaoanu , 

I would say have a try yourself and see the behavior change!  The mappingconfiguration gets altered by the derived types and so modifying behavior overall.

lbargaoanu:
Maybe. But that's just an accident.

RobinHerbots:
@lbargaoanu , 

When you look at the code example in the issue.  You can see that it is intentional.  I defined a mapping for bool already.  I don't want the mapping be overwritten by a nullable derived typemap.  This is exactly what happens in the current version and what is prevented by the small patch.  Althought still not the perfect scenario but at least a better as it is now.

RobinHerbots:
@lbargaoanu , 

Below a code example which gives a stackoverflow in version 5.1.1, which is offcourse solved with the pr  ;-)  
```
using System;
using AutoMapper;

namespace ConsoleApplication2
{
    class Program
    {
        {
            Mapper.Initialize(cfg =>
            {
                cfg.CreateMap<bool, string>().ConvertUsing(s => s ? "yes" : "no");
                cfg.CreateMap<bool?, string>().ConvertUsing(s => s.HasValue ? Mapper.Map<bool, string>(s.Value) : null);
            });

            bool test = true;
            Console.WriteLine("bool mapper" + Mapper.Map<bool, string>(test));
            Console.WriteLine("bool? mapper" + Mapper.Map<bool?, string>(test));

            Console.ReadKey();
        }
    }
}

```

lbargaoanu:
Or you can simply delete the nullable converter and get the same result.

RobinHerbots:
@lbargaoanu,

But that is what makes a library like automapper robust or fragile.

lbargaoanu:
Or stuck in the mud... Nobody likes to change their code, but you lose something to gain something. The Dude abides :)

RobinHerbots:
@lbargaoanu , 

Ok, my last try, ... 

explain me what do you gain/lose in context of the change of the pr?

PR #1672, opened 2016-09-06
Not needed code - nullables

lbargaoanu:
While testing nullable stuff I saw this code. Maybe it just needs a test.


RobinHerbots:
@lbargaoanu , 

Hmm, by removing that you loose the convenience that the mapper for the non-nullable equivalent is autogenerated.

RobinHerbots:
@lbargaoanu, 

Ok, 
... and it will solve #1669 and #1664 !

jbogard:
So that came from 294d64bfa36bdbe142d5e5ebfa235860cf863859 but there's no failing test. I can only assume that there *was* a test failing and I added this code to make the test pass. But now the test passes without it (and I don't like that solution of automatically adding type maps - that's bad).

lbargaoanu:
I fully agree :)

RobinHerbots:
@jbogard ,  @lbargaoanu , 

Thx!

PR #1677, opened 2016-09-13
Lazily create convert expressions

### LINKED ISSUE #1655: UWP 5.2.2 + AutoMapper 5.1.1 = Insanely long builds
With UWP 5.1.0 and AutoMapper 4.2.1 my release build time is 11 mins.
When bumping UWP 5.2.2 and AutoMapper 5.1.1 my release built time goes to 40 mins.

It looks like AutoMapper 5.1.1 has a dependency on UWP 5.2.2, otherwise I get the following
```Some packages are not compatible with UAP,Version=v10.0 (win10-arm-aot).
System.Reflection.Emit.ILGeneration 4.0.1 provides a compile-time reference assembly for System.Reflection.Emit.ILGeneration on UAP,Version=v10.0, but there is no run-time assembly compatible with win10-x64-aot.```

lbargaoanu:
I think it's cleaner this way. I guess the regular app will not use that many conversions.
Closes #1655.


lbargaoanu:
Our tests don't use any :) So there is no test for the ConvertMapper. And actually many of the conversions just throw.

lbargaoanu:
Removed DateTime and Char. This causes boxing with the TypeConverterMapper, but I wouldn't worry about it :)

TylerCarlson1:
I thought there was something special that everything was explicit for this type mapping,  Like silverlight or windows phone or something couldn't use reflection to get convert methods or something along these lines.  So you had to explicitly name each and every one.

Don't know if this is still applicable after .Net Standard change from old PCL style AutoMapper had before.

PR #1688, opened 2016-09-19
ResolveUsing overload with ResolutionContext for ctor params

jakubfijalkowski:
We already have the ability to resolve constructor parameters using `Func<TSource, object>`, but we can't access `ResolutionContext` there (as it is possible with property mappings). There is sufficient infrastructure for this, so the only change that was required was to provide correct overload of `ResolveUsing`.


jbogard:
Looks good!

jakubfijalkowski:
Great! Thanks!

PR #1692, opened 2016-09-21
Fixed memory leaks

### LINKED ISSUE #1691: ConcurrentBag causes memory leak
After recently upgrading from 4.2 to 5.1 our (FitNesse) test suite started running out of memory. In this testsuite the bootstrapper, and therefore our Automapper config, of our application is executed several (read thousands of) times.
I've tracked the problem to TypeMap's ConcurrentBag<PropertyMap>. The ConcurrentBag class has some issues with memory leaks (http://stackoverflow.com/questions/10850443/is-there-a-memory-leak-in-the-concurrentbagt-implementation).
In short, when initializing Automapper (both using static and instance api) multiple times on the same thread, all typemaps and propertymaps will be kept alive in a ThreadLocal until the thread is killed.

Anybody else experience this and/or have a proper solution?
Are there any way to properly dispose and clear the Automapper configuration?

I've temporarily fixed it by using the static API and only initializing Automapper if I detect that it is not initialized.

kopfsick:
Fixed memory leaks caused when initializing Automapper multiple times on
the same thread/appdomain.

Made Cache in SourceToDestinationNameMapperAttributesMember non-static.
Removed usage of ConcurrentBag in TypeMap, it now uses List<T>.

Closes #1691

### DIFF
--- a/src/AutoMapper/Configuration/Conventions/SourceToDestinationNameMapperAttributesMember.cs
+++ b/src/AutoMapper/Configuration/Conventions/SourceToDestinationNameMapperAttributesMember.cs
@@ -8,7 +8,7 @@ namespace AutoMapper.Configuration.Conventions
 
     public class SourceToDestinationNameMapperAttributesMember : ISourceToDestinationNameMapper
     {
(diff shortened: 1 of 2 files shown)


lbargaoanu:
Add "Closes #1691." to you comment so the issue will close when @jbogard merges this.

kopfsick:
Updated. Sorry, didn't know about that.

lbargaoanu:
No worries, Tips and tricks :)

Issue #279, opened 2012-12-20
byte array to byte array is very slow.


Sjaaky:
Mapping a byte[] property to a byte[] property is terribly slow. It is typically 50 times slower than a plain Array.Copy(). It seems automapper is handling byte[] as an IEnumerable<byte> and maps every byte individually.

One of the mapped classes (tested with Data = new byte[1000])
    public class Document
    {
        public int Id { get; set; }
        public byte[] Data { get; set; }
    }

Adding the following map solves the issue:
    Mapper.CreateMap<byte[], byte[]>().ConstructUsing(src =>
    {
        byte[] dest = new byte[src.Length];
        Array.Copy(src, dest, src.Length);
        return dest;
    });

I tested this also with arrays of char, short, int, long, ushort, uint, ulong. So the problem probably exists with arrays of all primitive types.
Could you add  these special cases to the automapper?

I came across this message https://groups.google.com/forum/?fromgroups=#!topic/automapper-users/bCKicMmVgHY . It suggests it only occurs on a 64bit machine. I've tested it using a 64bit machine, testing on a 32 bit machine would take me more effort so I'd rather not.

PR #1714, opened 2016-10-05
Always map array contents & remove the primitive array mapper.

lbargaoanu:
I don't think PrimitiveArrayMapper is worth keeping as an optimization (presumably that was the intent).


jbogard:
Yeah, that was the intent. There was an issue a long time ago where someone was mapping a really large byte array, and Array.Copy was blazing fast.

jbogard:
I found it! #279 but I don't know, given the numbers, how much was just old AutoMapper slowness

lbargaoanu:
I've done some ad-hoc microbenchmarks and yes, it's faster, but I think for large arrays you should just avoid copying them in the first place. And for smaller ones, it's just pennies.

lbargaoanu:
I think my point here is clear. The ArrayMapper does the right thing fast enough, the PrimitiveArrayMapper is faster because it doesn't really map, it just copies. But if you want to keep it, I should revert that part and leave the fix in.

jbogard:
So in the large array case, what SHOULD be the behavior? Suppose someone has something like an image stored as a byte array, or a document. When I tested with a large byte array (1M items), there was a huge time difference between the two.

lbargaoanu:
All I see is that the ArrayMapper is twice as slow. But it maps :) So if you have a map, it will call it, as with any other type. This should be the default behavior in my view. Users should special case them in they need reference assignment (most likely, I believe) or Array.Copy.

jbogard:
Sounds good.

lbargaoanu:
For future reference. The way to special case them would be AfterMap or a type converter (ResolveUsing or MapFrom wouldn't work). Right? :)

PR #1718, opened 2016-10-06
When we map a nullable, we should map the underlying type unless an e…

### LINKED ISSUE #1717: Mapping from nullable enum stopped working in AutoMapper 5
Mapping from nullable enum doesn't work anymore for general enum mapping (using Enum class).
Worked fine in 4.2.1.

`static void Main(string[] args)`
`{`
`    Mapper.Initialize(cfg =>`
`        {`
`            cfg.CreateMap<TestClass, TestResult>();`
`            cfg.CreateMap<Enum, string>().ConvertUsing((Enum src) => "Test mapping");`
`        });`
``
`    var result = Mapper.Instance.Map<TestResult>(new TestClass { Color = ConsoleColor.Black });`
``
`    Debug.Assert(result.Color == "Test mapping", result.Color);`
`}`
``
`public class TestClass`
`{`
`    public ConsoleColor? Color { get; set; }`
`}`
``
`public class TestResult`
`{`
`    public string Color { get; set; }`
`}`

lbargaoanu:
…xplicit map exists
The ConvertMapper can call ToString without boxing.
Fixes #1717.

### DIFF
--- a/src/AutoMapper/Mappers/MapperRegistry.cs
+++ b/src/AutoMapper/Mappers/MapperRegistry.cs
@@ -6,6 +6,7 @@ namespace AutoMapper.Mappers
     {
         private static readonly IObjectMapper[] _initialMappers =
         {
+            new NullableSourceMapper(),
             new ExpressionMapper(), 
             new FlagsEnumMapper(),
             new StringToEnumMapper(), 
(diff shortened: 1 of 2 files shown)


jbogard:
The order of the mapper registries is getting weird. I'm starting to think it's not the best way to control mapping, too many bugs around moving things up and down in that list.

lbargaoanu:
It's not obvious what each does exactly and what the best order would be.

hakant:
Is this also supposed to fix the "string to nullable enum" mapping issue? I think starting from V5 AutoMapper can't map from string to nullable enums anymore.

Example:

```csharp
    class Source
    {
        public string State { get; set; }
    }

    class Destination
    {
        public State? State { get; set; }
    }

    enum State
    {
        Off = 0,
        On = 1
    }

    class Program
    {
        {
            {
            });

            var source = new Source
            {
                State = "On"
            };

            // Following line throws "AutoMapper.AutoMapperMappingException: Error mapping types."
            // Property:
            // State--->System.ArgumentException: Type provided must be an Enum.
            // Parameter name: enumType

            var destination = mapper.Map<Destination>(source);
        }
    }
```

lbargaoanu:
No, please open another issue.

PR #1734, opened 2016-10-14
Compile once

lbargaoanu:


src/AutoMapper/MapperConfiguration.cs
```diff
@@ -415,4 +414,21 @@ private static LambdaExpression TryCatch(LambdaExpression mapExpression, Paramet
             }
         }
     }
-}
+
+    public struct LockingConcurrentDictionary<TKey, TValue>
```
jbogard: Should this be public?

lbargaoanu: It can be useful in general.

jbogard: Useful for who? Asking because I've made a few of these internal and broke people when I've moved/changed them. Like reflection extension methods that people started using because they were public.

lbargaoanu: Done.

lbargaoanu:
@jbogard [Other people](https://blog.oneunicorn.com/2016/11/09/internal-code-in-ef-core-1-1/) see it my way too. But I understand :)

jbogard:
We can move some types into an internal namespace in that case. The only exception is probably the extension methods, those are too easy to just start using and depending on.

lbargaoanu:
I don't know. I mean you'd have to explicitly add a using for that internal namespace, they are not visible by default.

SaltyDH:
Not sure if I should continue the discussion here on the pull request or back over in my original feedback issue, but I would definitely like to voice a +1 at the very least for simplifying the experience of building custom object mappers. I like the approach of exposing the internal namespaces as described in the EF blog link above. As a developer, I would fully expect the following when utilising something in an "Internal" namespace:

1. The code is without general support - submitting a ticket / issue would generally require proving the internal functionality breaks when used through the standard API surface area.
2. I'm responsible for how I interact with this code, my own tests proving it works correctly for my use cases. (Really that's ways the case, just that I would put a larger emphasis on testing this code in isolation with some behavioural tests.
3. The methods may come and go on a whim and signatures may change without notice.

Having to explicitly add namespaces to my custom modules is a minor inconvenience and given the built in support in VS2017 and tooling like R# etc, this is very much a passing thought these days - IMHO of course :)

lbargaoanu:
Please move this to the new issue.

PR #1740, opened 2016-10-17
Mapping to/from object doesn't seem useful

### LINKED ISSUE #1739: Issue With Mapping Dynamic Nested Elements
I've created an EF query similar to the following sample using expressions:

```
Expression<Func<MyModel, dynamic>> columns = x => new
{
        x.EncounterId,
        x.PatientId,
        x.Date,
        x.Time,
        Patient = x.PatientEntity
}
```

**PatientEntity** is an entity framework entity class with the following implementation:

```
public partial class PatientEntity
{
        public int PatientId { get; set; }        
        public string FullName { get; set; }
}
```

I have a DTO class similar to the **PatientEntity** class with the following implementation:

```
public class PatientModel
{
        public int PatientId { get; set; }        
        public string FullName { get; set; }
}
```

I also have a container class with the following implementation:

```
public class DashboardModel
{
        public int EncounterId { get; set; }
        public int PatientId { get; set; }
        public DateTime Date { get; set; }
        public TimeSpan Time { get; set; }
        public PatientModel Patient { get; set; }
        public decimal RemainingBalance { get; set; }
}
```

I used to map these two using the following code. But I updated AutoMapper from 5.0.2 to 5.1.1 and the following code doesn't map the **Patient** property anymore (It returns null).

```
var dataModel = await Service.GetListAsync(columns);
var model = Mapper.Map<List<DashboardModel>>(dataModel);
```

I guess something changes in the configuration of mapper that prevents the mapper to map nested properties of a property. Anyway, how would I resolve this? 
**Thanks.**

lbargaoanu:
Fixes #1739.

### DIFF
--- a/src/AutoMapper/Profile.cs
+++ b/src/AutoMapper/Profile.cs
@@ -17,7 +17,7 @@ namespace AutoMapper
     /// </summary>
     public abstract class Profile : IProfileExpression, IProfileConfiguration
     {
-        private readonly ConditionalObjectMapper _mapMissingTypes = new ConditionalObjectMapper {Conventions = {tp => true}};
+        private readonly ConditionalObjectMapper _mapMissingTypes = new ConditionalObjectMapper {Conventions = { tp => tp.SourceType != typeof(object) && tp.DestinationType != typeof(object) }};
         private readonly List<string> _globalIgnore = new List<string>();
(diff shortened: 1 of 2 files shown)


jbogard:
Weird we didn't have this explicitly tested for.

PR #1753, opened 2016-10-24
Expose the destination member

lbargaoanu:
It was needed a couple of times.

### DIFF
--- a/src/AutoMapper/Configuration/MemberConfigurationExpression.cs
+++ b/src/AutoMapper/Configuration/MemberConfigurationExpression.cs
@@ -13,7 +13,7 @@ namespace AutoMapper.Configuration
         MemberInfo DestinationMember { get; }
     }
 
-    public class MemberConfigurationExpression<TSource, TDestination, TMember> : IMemberConfigurationExpression<TSource, TDestination, TMember>
+    public class MemberConfigurationExpression<TSource, TDestination, TMember> : IMemberConfigurationExpression<TSource, TDestination, TMember>, IMemberConfiguration
     {
         private readonly MemberInfo _destinationMember;
         private readonly Type _sourceType;
--- a/src/AutoMapper/Configuration/MemberConfigurationExpression.cs
+++ b/src/AutoMapper/Configuration/MemberConfigurationExpression.cs
@@ -13,7 +13,7 @@ namespace AutoMapper.Configuration
         MemberInfo DestinationMember { get; }
     }
 
-    public class MemberConfigurationExpression<TSource, TDestination, TMember> : IMemberConfigurationExpression<TSource, TDestination, TMember>, IMemberConfiguration
+    public class MemberConfigurationExpression<TSource, TDestination, TMember> : IMemberConfigurationExpression<TSource, TDestination, TMember>
     {
         private readonly MemberInfo _destinationMember;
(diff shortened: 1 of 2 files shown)


Itamaram:
Piggybacking off this, what would be the best way to access the destination member in a ValueResolver?

(Use case is mapping a property out a dictionary on source object)

lbargaoanu:
You need to pass it yourself to the constructor of your resolver. Or use a lambda resolver.

Itamaram:
(Please let me know if I should start my own thread, or take this conversation elsewhere and I will, but it seems relevant enough so I'm keeping it here)

I'm not entirely sure how I can achieve what I'm after with either one of your suggestions.

Back in AutoMapper 3.2.1, my resolver looked like this:
```
public class CustomResolver : IValueResolver
{
    public ResolutionResult Resolve(ResolutionResult source)
    {
        var value = ((Foo) source.Context.SourceValue).Dictionary[source.Context.MemberName];
        return source.New(value);
    }
}
```
And my map included several lines of the form:
`.ForMember(x => x.SomeProp, x => x.ResolveUsing<CustomResolver>())`

In AutoMapper 5,
When I call `ForMember`, my second argument (`Action<IMemberConfigurationExpression>`), lacks the context of the first argument, so regardless of whether I have an inline func, or I instantiate a resolver, I am unable to work out the member name.

A potential workaround is to pass the func twice eg:
`.ForMember(x => x.SomeProp, x => x.ResolveUsing(() => new CustomResolver(x => x.SomeProp))`
But I would really like to avoid this repetition, as well as the the manual instantiation of the resolver.

jbogard:
You need IMemberValueResolver. I split the interfaces between ones that just need the source/dest parent objects, and the ones that need source/dest parent objects AND members.

Itamaram:
I'm afraid I do not follow. `IMemberValueResolver` exposes an additional field, `TSourceMember sourceMember`, and that's it. I still have no access to the member name, just its type.

lbargaoanu:
I'm not saying naming is easy, but surely you can do better than x :) A better name would be options and that might cause you to ask about its type. And that is IMemberConfigurationExpression which has a member DestinationMember.

Itamaram:
Thanks @lbargaoanu, this indeed resolves the first half of the issue (I'm really not sure how I've missed it, I was staring at the interface for a fair while...)

There still exists a case though in which it is needed for a resolver to be instantiated by the IoC container AND have access to the abovementioned `DestinationMember` property.

Is this currently not possible? Or am I still misunderstanding?

lbargaoanu:
You need to set the destination member inside that resolver instance. So get the instance from the container, set the destination and then pass it to Resolve. I don't see an easier way right now.

lbargaoanu:
Well, a lot of people use AM, and more than a few wanted smth else added to the ResolutionContext. This didn't go well...

jbogard:
@Itamaram because it's suuuuuper expensive to keep track of that stuff at every point *just in case*. If we wanted that, we'd need to change the overload of IValueResolver or IMemberValueResolver to pass that stuff in. That way we only allocate/use it when people use value resolvers. Before we'd set it all the time, for every member, everywhere. Very expensive.

Itamaram:
I understand. Thank you for your time and assistance!

PR #1765, opened 2016-10-30
Fix #1482

### LINKED ISSUE #1482: UseAsDataSource Translate constants with PropertyMap changes.
Request from #1461 

I'm thinking check Binary Operations and if one side has a constant and other has map change, then change the constant in the same way the property map changed.

Same for function calls that have the base change and update the constants that had a matching type.
Ex:  Equals() call vs using == operator.

Maybe more scenarios need to be accounted for but it's a start.

The ExpressionMapper ExpressionVisitor logic is going to get a lot more complex :)

TylerCarlson1:
Just fixed the ToString issue.  I think in the long run IObjectMapper, IExpressionBuilder, and these expression translation classes should all be in a single class or same file since they do the same kinds of things but in different Expression Mapping Scenarios.

Assign, ToString, and possibly Map can be applied to all 3 expression translation scenarios

Slices not Layers I guess?

### DIFF
--- a/src/AutoMapper/Mappers/ExpressionMapper.cs
+++ b/src/AutoMapper/Mappers/ExpressionMapper.cs
@@ -144,6 +144,10 @@ namespace AutoMapper.Mappers
                 var newLeft = base.Visit(node.Left);
                 var newRight = base.Visit(node.Right);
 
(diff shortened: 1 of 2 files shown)


jbogard:
Fixes #1482

PR #1766, opened 2016-10-30
Inheritance baked into expression v2

TylerCarlson1:
Re-did #1625 as there were some refactorings that degraded performance.

Added benchmark to show include mapping performance as first commit.

### DIFF
--- a/src/AutoMapper/Execution/TypeMapPlanBuilder.cs
+++ b/src/AutoMapper/Execution/TypeMapPlanBuilder.cs
@@ -571,22 +571,8 @@ namespace AutoMapper.Execution
             var typeMap = configurationProvider.ResolveTypeMap(typePair);
             if(typeMap != null)
             {
-                if(!typeMap.HasDerivedTypesToInclude())
-                {
-                    typeMap.Seal(typeMapRegistry, configurationProvider);
-                    if(typeMap.MapExpression != null)
-                    {
-                        return typeMap.MapExpression.ConvertReplaceParameters(sourceParameter, destinationParameter, contextParameter);
-                    }
-                    else
-                    {
-                        return ContextMap(typePair, sourceParameter, contextParameter, destinationParameter);
-                    }
-                }
-                else
-                {
-                    return ContextMap(typePair, sourceParameter, contextParameter, destinationParameter);
-                }
+                return GetMapperExpression(configurationProvider, typeMapRegistry, typeMap)?.ConvertReplaceParameters(sourceParameter, destinationParameter, contextParameter)
+                       ?? ContextMap(typePair, sourceParameter, contextParameter, destinationParameter);
             }
             var match = configurationProvider.GetMappers().FirstOrDefault(m => m.IsMatch(typePair));
             if(match != null)
@@ -595,6 +581,13 @@ namespace AutoMapper.Execution
                 return ToType(mapperExpression, typePair.DestinationType);
             }
             return ContextMap(typePair, sourceParameter, contextParameter, destinationParameter);
+        }
+
+        private static readonly MethodInfo GetMapperExpressionMethod = typeof(IConfigurationProvider).GetDeclaredMethod("GetMapperExpression");
+        private static LambdaExpression GetMapperExpression(IConfigurationProvider configurationProvider, TypeMapRegistry typeMapRegistry, TypeMap typeMap)
+        {
+            typeMap.Seal(typeMapRegistry, configurationProvider);
+            return GetMapperExpressionMethod.MakeGenericMethod(typeMap.SourceType, typeMap.DestinationType).Invoke(configurationProvider, new object[] { typeMap.Types }) as LambdaExpression;
         }
 
         private static Expression ContextMap(TypePair typePair, Expression sourceParameter, Expression contextParameter, Expression destinationParameter)
--- a/src/AutoMapper/MapperConfiguration.cs
+++ b/src/AutoMapper/MapperConfiguration.cs
@@ -66,17 +66,58 @@ namespace AutoMapper
         {
             var key = new TypePair(typeof(TSource), typeof(TDestination));
             var mapRequest = new MapRequest(key, types);
-            return (Func<TSource, TDestination, ResolutionContext, TDestination>)GetMapperFunc(mapRequest);
+            var mapperFuncs = GetMapperFuncs(mapRequest);
+            return GetSubMap<TSource, TDestination>(mapperFuncs)?.Compile() ?? (Func<TSource, TDestination, ResolutionContext, TDestination>)mapperFuncs.Typed;
+        }
+
+        public Expression<Func<TSource, TDestination, ResolutionContext, TDestination>> GetMapperExpression<TSource, TDestination>(TypePair typePair)
+        {
+            var key = new TypePair(typeof(TSource), typeof(TDestination));
+            var mapRequest = new MapRequest(key, typePair);
+            var mapperFuncs = GetMapperFuncs(mapRequest);
+            return GetSubMap<TSource, TDestination>(mapperFuncs) ?? (Expression<Func<TSource, TDestination, ResolutionContext, TDestination>>)mapperFuncs.TypedExpression;
+        }
+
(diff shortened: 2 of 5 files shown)


TylerCarlson1:
Ops forgot a manual include map to test against.

Anyways the IncludeMap benchmark's expressions looks like this now, where instead of calling map and resolving mapping contexts it calls the function directly.

I think the test would be a lot better if there were more than 3 items to show the performance gain of not having to waste time calling Map for every item.

```
(src, dest, ctxt) =>
{
    if (src == null)
    {
        null;
    }
    else
    {
        var typeMapDestination = dest ?? new Destination();
        try
        {
            var resolvedValue = (false || (src == null)) ? null : src.List;
            var propertyValue = (var passedDestination = (dest == null) ? null : typeMapDestination.List;
            if (passedDestination != null)
            {
                var collectionDestination = passedDestination;
                collectionDestination.Clear();
            }
            else
            {
                collectionDestination = new List<DestinationBase>();
            }

            return if (resolvedValue == null)
            {
                return collectionDestination;
            }

            var enumerator = resolvedValue.GetEnumerator();
            while (true)
            {
                if (enumerator.MoveNext())
                {
                    var item = (Base)enumerator.Current;
                    collectionDestination.Add(
                        if (item is BaseInherit1)
                        {
                            return (DestinationBase)if ((item as BaseInherit1) == null)
                            {
                                null;
                            }
                            else
                            {
                                var typeMapDestination = (null as DestinationInherit1) ?? new DestinationInherit1();
                                return typeMapDestination;
                            }
                        }

                        if (item is BaseInherit2)
                        {
                            return (DestinationBase)if ((item as BaseInherit2) == null)
                            {
                                null;
                            }
                            else
                            {
                                var typeMapDestination = (null as DestinationInherit2) ?? new DestinationInherit2();
                                return typeMapDestination;
                            }
                        }

                        if (item == null)
                        {
                            null;
                        }
                        else
                        {
                            var typeMapDestination = null ?? new DestinationBase();
                            return typeMapDestination;
                        });
                }
                else
                {
                    break;
                }
            }

            return collectionDestination);
            typeMapDestination.List = propertyValue;
        }
        catch
        {
            throw new AutoMapperMappingException("Error mapping types.", ex, AutoMapper.TypePair, AutoMapper.TypeMap, AutoMapper.PropertyMap);
        }

        return typeMapDestination;
    }
}
```

src/AutoMapper/MapperConfiguration.cs
```diff
@@ -428,7 +480,20 @@ public LockingConcurrentDictionary(Func<TKey, TValue> valueFactory)
...
+            catch (InvalidOperationException e) when(e.Message == "ValueFactory attempted to access the Value property of this instance.")
+            {
+                return default(TValue);
+            }
+        }
+
+        public TValue AddOrUpdate(TKey key)
+        {
+            LockingConcurrentDictionary<TKey, TValue> tmpThis = this;
+            return tmpThis._dictionary.AddOrUpdate(key, tmpThis._valueFactory, (tp, mf) => tmpThis._valueFactory(tp)).Value;
```
lbargaoanu: This is weird and allocates a delegate on each call, unlike GetOrAdd above.

src/AutoMapper/Execution/TypeMapPlanBuilder.cs
```diff
@@ -562,44 +562,37 @@ public Expression MapExpression(TypePair typePair, Expression sourceParameter, P
...
             }
-            return ContextMap(typePair, sourceParameter, contextParameter, destinationParameter);
+            return ContextMap(typePair, sourceParameter, contextParameter, destinationParameter);
+        }
+
+        private static readonly MethodInfo GetMapperExpressionMethod = typeof(IConfigurationProvider).GetDeclaredMethod("GetMapperExpression");
+        private static LambdaExpression GetMapperExpression(IConfigurationProvider configurationProvider, TypeMapRegistry typeMapRegistry, TypeMap typeMap)
+        {
+            typeMap.Seal(typeMapRegistry, configurationProvider);
+            return GetMapperExpressionMethod.MakeGenericMethod(typeMap.SourceType, typeMap.DestinationType).Invoke(configurationProvider, new object[] { typeMap.Types }) as LambdaExpression;
```
lbargaoanu: GetMapperExpression could be non generic and receive the types as objects and then you don't need reflection.

src/AutoMapper/MapperConfiguration.cs
```diff
@@ -66,25 +66,66 @@ public MapperConfiguration(Action<IMapperConfigurationExpression> configure, IEn
         {
             var key = new TypePair(typeof(TSource), typeof(TDestination));
             var mapRequest = new MapRequest(key, types);
-            return (Func<TSource, TDestination, ResolutionContext, TDestination>)GetMapperFunc(mapRequest);
+            var mapperFuncs = GetMapperFuncs(mapRequest);
+            return GetSubMap<TSource, TDestination>(mapperFuncs)?.Compile() ?? (Func<TSource, TDestination, ResolutionContext, TDestination>)mapperFuncs.Typed;
```
lbargaoanu: I think this sub map this should be inside the MapperFunc. This way is not even clear when the compile happens.

src/AutoMapper/MapperConfiguration.cs
```diff
@@ -66,25 +66,66 @@ public MapperConfiguration(Action<IMapperConfigurationExpression> configure, IEn
...
+            var mapperFunc = _mapPlanCache.GetOrAdd(mapRequest);
+            if (mapperFunc.TypedExpression == null)
+                mapperFunc = _mapPlanCache.AddOrUpdate(mapRequest);
+            return mapperFunc;
         }
 
         public Delegate GetMapperFunc(MapRequest mapRequest)
         {
-            return _mapPlanCache.GetOrAdd(mapRequest).Typed;
+            return _mapPlanCache.GetOrAdd(mapRequest).Typed ??
```
lbargaoanu: GetOrAdd and then AddOrUpdate? :) Again, maybe encapsulating inside the MapperFunc would make it more clear.

TylerCarlson1: I have to try to get it twice.  Cause the first time might fail because of trying to get the included mappings might not work because of circular references, or recursion.  But after that the second time trying to resolve it will work.

src/AutoMapper/MapperConfiguration.cs
TylerCarlson1: Yea the thing wouldn't let me compile otherwise.  This is what Resharper did to make it compile.

src/AutoMapper/MapperConfiguration.cs
```diff
@@ -428,7 +480,20 @@ public LockingConcurrentDictionary(Func<TKey, TValue> valueFactory)
 
         public TValue GetOrAdd(TKey key)
         {
-            return _dictionary.GetOrAdd(key, _valueFactory).Value;
+            try
+            {
+                return _dictionary.GetOrAdd(key, _valueFactory).Value;
+            }
+            catch (InvalidOperationException e) when(e.Message == "ValueFactory attempted to access the Value property of this instance.")
```
TylerCarlson1: Yea the circular reference causes and exception with trying to get a Value from Lazy object in LockingDictionary.  Cause you are trying to get a value, while you are already trying to actively resolve said value.  The reason I don't ignore just this exception is because there's a test that expects to throw InvalidOperationException, so I filtered by exception message.

src/AutoMapper/MapperConfiguration.cs
lbargaoanu: Yes, I get that, but do you have to go through the dictionary twice? Can't this double check be encapsulated?

lbargaoanu:
An optional constructor parameter would work.

jbogard:
I wonder, instead of trying to compile twice, you can include a placeholder Expression type in the expression tree. Then you can make two passes, one for the initial expression creation, then another to go replace those placeholders in the expression tree with the real thing.

TylerCarlson1:
@jbogard Don't know.  Probably, you've done circular reference before.

Have to go back and re-look at it again.

Also set this PR to edit for maintainers, so you can change this branch directly instead of PRing a PR.

Issue #1796, opened 2016-11-16
Apply Root Configuration to Profiles


jbogard:
Currently, root configuration is separate from profiles. This includes:
- AllowNullCollections
- AllowNullDestinationValues
- ConstructorMapping
- ShouldMapField/Property
- CreateMissingTypeMaps
- Naming conventions
- Global ignores
- Extension methods
- Conventions

To fix this, we need to segregate the profile configuration entity from the runtime profile configuration. To do so, I'm creating a new class, ProfileMap, to be the analog to TypeMap and PropertyMap to represent the final configuration for a profile.

This also means that when we build a ProfileMap, I can examine the root configuration and anything set in your profile, to build the final picture. In order to do so, many of the defaults need to be removed in the Profile class, in addition to making some of the values nullable so that I can detect if they're set or not.

The final picture is a bit cleaner, with the ProfileMap class containing behavior of "how do I create a TypeMap", and the Profile class containing behavior of "how do I gather Profile configuration".

jbogard:
See also #1618 and #1712

jbogard:
Fixed with #1797

PR #1797, opened 2016-11-16
Global configuration

jbogard:
See #1796 for more information

### DIFF
--- a/src/AutoMapper/Configuration/Conventions/DefaultMember.cs
+++ b/src/AutoMapper/Configuration/Conventions/DefaultMember.cs
@@ -11,7 +11,7 @@ namespace AutoMapper.Configuration.Conventions
     {
         public IParentSourceToDestinationNameMapper NameMapper { get; set; }
 
-        public bool MapDestinationPropertyToSource(IProfileConfiguration options, TypeDetails sourceType, Type destType, Type destMemberType, string nameToSearch, LinkedList<MemberInfo> resolvers, IMemberConfiguration parent = null)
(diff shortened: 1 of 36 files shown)


jbogard:
@TylerCarlson1 @lbargaoanu this adds the ProfileMap companion to TypeMap and PropertyMap. I'm not sure how I missed this one in the config overhaul, but it was quite a bit harder to break apart than the others.

lbargaoanu:
That's a lot :) And it makes configuration more complex. But it makes sense.

jbogard:
Yeah, it definitely adds more classes. I think this makes it consistent with the rest of the config API - you have config gatherers:

- Profile : IProfileExpression, IProfileConfiguration
- MappingExpression : IMappingExpression, ITypeMapConfiguration
- MemberConfigurationExpression : IMemberConfigurationExpression, IMemberConfiguration

Then configuration holders:

- ProfileMap
- TypeMap
- PropertyMap

And a very explicit boundary - MapperConfiguration can create the root ProfileMap and sub profiles. ProfileMap can create type maps, TypeMaps can create PropertyMaps, delegating to the IXyzConfiguration to fill in the details. Finally something that reads these configuration models to build the plan of execution.

More moving pieces but hopefully it helps bring the Profile class more consistent with the rest of the config API.

TylerCarlson1:
My main concern is does mapper configuration conventions add on with sub profile's mapping.  Like if in a sub profile I want to remove flattening or other default member conventions, for whatever reasoning when mapping properties, and I have it in the mapping configuration, will that re-add it back into the sub profile?

As far as I can tell it doesn't look like it, but I haven't looked over the code extensively.

jbogard:
No, I left the conventions alone as they currently are. You'd want to be able to add/remove them, so for everything you do in the root config, you'd want to undo in a profile.

Issue #1685, opened 2016-09-18
MethodAccessException: Windows Phone 8.1 (WinRT)


adamlysak:
I have PCL where I using AutoMapper 5.1.1 with code:

```
var mapperConfig = new MapperConfiguration(cfg =>
{
    cfg.AddProfile<TileProfile>();
    cfg.AddProfile<PictureProfile>();
    ...
});

var mapper = mapperConfig.CreateMapper();
```

Which working well with WinForms but when use my library with Windows Phone 8.1 (WinRT) then I get exception in **MapperConfiguration** constructor:

```
System.MethodAccessException was unhandled by user code
  HResult=-2146233072
  Message=Attempt by security transparent method 'AutoMapper.Profile..ctor()' to access security critical method 'System.Collections.Concurrent.ConcurrentDictionary`2<System.__Canon,System.__Canon>..ctor()' failed.
  Source=AutoMapper
  StackTrace:
       at AutoMapper.Profile..ctor()
       at AutoMapper.Configuration.MapperConfigurationExpression..ctor()
       at AutoMapper.MapperConfiguration.Build(Action`1 configure)
       at AutoMapper.MapperConfiguration..ctor(Action`1 configure)
  InnerException: 
```

I also create blank Windows Phone 8.1 project for test with only Automapper nuget package and this code:

```
...
public MainPage()
{
    this.InitializeComponent();

    var mapperConfig = new MapperConfiguration(cfg => {});
}
...

```

with same result. Do anyone some experience with this issue? 
It looks like to me as bug in Automapper.

Thanks

Version: 
Automapper 5.1.1, portable-net45+win8+wpa81

lbargaoanu:
Try googling that error.

AKorpysz:
I've the same issue.
Unfortunately answers google answers are not working.
Could someone share any idea how to fix that error ?
Thank in advance

adamlysak:
I fix the issue by recompiling Automapper from source and removing attribute:
`[assembly: AllowPartiallyTrustedCallers]`
from the AssemblyInfo.cs

jbogard:
Seriously?!?

adamlysak:
Yes, really

jbogard:
Well that's an easy fix then. I can just include that for full .NET

lbargaoanu:
@jbogard There's nothing wrong with that attribute (and it helps with performance, remember? :)). They just need to set smth in their app.

[...]

PR #1812, opened 2016-11-23
Applying partially trusted callers only for full .NET

jbogard:


jbogard:
Closes #1685

lbargaoanu:
So for example .NET Core loses performance for Windows Phone? :)

jbogard:
Yes but it's offset because .NET Core is webscale

jbogard:
I'm not sure the perf hits with the partial trust piece shows up in other areas. It's not a thing in .NET Core, doesn't seem to work in WinRT, and I don't think it's a thing in Xamarin either. From what I can see, the partial trust perf hit is only a thing in full .NET.

But it's hard to test tbqh

lbargaoanu:
Yeah, maybe :)

lbargaoanu:
Some ad-hoc testing suggests you're right. I see no overhead in .NET Core.

PR #1822, opened 2016-11-28
Just look for IMemberValueResolver; fixes #1813

### LINKED ISSUE #1813: Interfaces derived from IMemberValueResolver<,,,> break BuildValueResolverFunc
If I specify my own interface:

`public interface IMyValueResolver<TSrcMember,TDestMember> : IMemberValueResolver<object,object,TSrcMember,TDestMember> {
  ...
}`
I cannot supply an implementing type to ResolveUsing().  `BuildValueResolverFunc ` throws an exception while building the expressions due to the interface types not having the same number of generic parameters.

This is caused by the below, which in my case is finding both the interface that I have defined myself as well as the generic implementation. While this may not be the best way for me to do things, the below should most likely be checking that the generic argument count matches when selecting the correct type

```
var iResolverType =
                        resolverType.GetTypeInfo()
                            .ImplementedInterfaces.First(t => t.ImplementsGenericInterface(typeof(IMemberValueResolver<,,,>)));
```

lbargaoanu:


jbogard:
Fixes #1813

lbargaoanu:
I didn't forget this time :) There is a commit message with fixes #1813. It's true that I used to put it in a PR comment, but I guess I'm not consistent.

jbogard:
No i just added it so that I could click it

lbargaoanu:
:)

lbargaoanu:
It works in the commit message too, it's a link.

PR #1834, opened 2016-12-03
Always null check

### LINKED ISSUE #1833: Auto-flattening null reference exception
I have the following basic model structure:

	public class OrderModel
	{
		public VendorModel Vendor { get; set; }
		public string Number { get; set; }
	}

	public class VendorModel
	{
		public CurrencyModel Currency { get; set; }
	}

	public class CurrencyModel
	{
	}

	public class OrderDTO
	{
		public CurrencyDTO VendorCurrency { get; set; }
		public string VendorName { get; set; }
		public string Number { get; set; }
	}

	public class CurrencyDTO
	{
	}

With this code as a test:

	var config = new MapperConfiguration(c =>
	{
		c.CreateMap<OrderModel, OrderDTO>();
	});

	var orderModel = new OrderModel()
	{
		Number = "1",
		Vendor = null
	};

	var dto = mapper.Map<OrderDTO>(orderModel);

And it's throwing a NullReferenceException on the Map statement, and I think it's coming from the fact that it's trying to map OrderDTO.VendorCurrency -> Name from Vendor -> Currency -> Name based on the naming of the properties, and I guess it's not doing null checks with the multiple levels.  For instance, if I create a Vendor in the OrderModel (but not a Currency), it works:

	var orderModel = new OrderModel()
	{
		Number = "1",
		Vendor = new VendorModel() { Name = "vendor", Currency = null }
	};

TylerCarlson1:
Fixes #1833

### DIFF
--- a/src/AutoMapper/Execution/TypeMapPlanBuilder.cs
+++ b/src/AutoMapper/Execution/TypeMapPlanBuilder.cs
@@ -486,10 +486,7 @@ namespace AutoMapper.Execution
                                 : (Expression)Call(inner, (MethodInfo)getter)
                             : MakeMemberAccess(getter.IsStatic() ? null : inner, getter)
                         );
-                    if(destinationPropertyType == valueResolverFunc.Type || _configurationProvider.ResolveTypeMap(valueResolverFunc.Type, destinationPropertyType) == null)
-                    {
-                        valueResolverFunc = valueResolverFunc.IfNotNull(destinationPropertyType);
-                    }
+                    valueResolverFunc = valueResolverFunc.IfNotNull(destinationPropertyType);
                 }
             }
             else if(propertyMap.SourceMember != null)
--- a/src/UnitTests/BasicFlattening.cs
+++ b/src/UnitTests/BasicFlattening.cs
@@ -149,4 +149,61 @@ namespace AutoMapper.UnitTests
             _destination.ParentDataValue.ShouldBeNull();
         }
     }
+
+    public class NullTypeMapFlattening : AutoMapperSpecBase
+    {
+        private OrderDTO _dto;
+
+        public class OrderModel
+        {
+            public VendorModel Vendor { get; set; }
+            public string Number { get; set; }
+        }
+
+        public class VendorModel
+        {
+            public string Name { get; set; }
+            public CurrencyModel Currency { get; set; }
+        }
+
+        public class CurrencyModel
+        {
+            public string Name { get; set; }
+        }
+
+
+        public class OrderDTO
+        {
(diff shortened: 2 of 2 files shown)


lbargaoanu:
The idea was to be able to override the built in null check. So maybe check for a type converter, not just for a map? It seems wrong to simply hard-code this.

TylerCarlson1:
This is right after aggregating the MemberInfo list to make the expression.  This is just `a.b.c.d.e.f` as far as expressions go.  This should be null checked always regardless of how the resolved value gets assigned

If the end method types use a type converter, you still need to null check all the properties that get there before you call the type converter, otherwise you'll just get null exception again.

lbargaoanu:
Some other way then. The problem is that you're making an implicit conversion and you have no way to override it. And this implicit conversion caused more than a few issues in the past. Is this now safer? I doubt it.

TylerCarlson1:
It's an implicit conversion on an implicit rule (Flattening).  If you look [here](https://github.com/TylerCarlson1/AutoMapper/blob/1e3a8138b0e426256e53f0a218d7d2eef7b2b752/src/AutoMapper/Execution/TypeMapPlanBuilder.cs#L457) you'll see that if you use CustomExpression which is MapFrom you still always IfNotNull it.
Whether you get 'a.b.c' from flattening or explicitly set mapping for it, it should always do null checks regardless.

And yes you can't change it, because this is a safety feature for expressions.  Because there's no good exception for this to know where in code you failed.  For Resolvers and other non-expression resolvers you have no ability to add that null check, but you can debug and get exception messages that point to user code.

Also if you could please explain the problems this might cause, or has caused in the past, because as far as I understand it now there's no issue with this.  I could be missing something critical somewhere else in code but right now I see it only isolated to resolving the source value.

lbargaoanu:
The original test passes somehow, maybe nobody will notice.

TylerCarlson1:
Notice what?  Which test are you talking about?  The NullFlattening test?

lbargaoanu:
No, this initially had nothing to do with flattening (check the history). Changing the type of the resolved value caused a type converter for the initial type pair not no match. I think now it matches because of the NullableSourceMapper. So if you change int to int?, you then map int? to int? and a type converter int to int? will match. So in this case I think the behavior is preserved.

TylerCarlson1:
Oh Ok I see what you mean now.  I can look into that scenario more.

Yes it didn't fail any tests, but those nulls are always tricky.

lbargaoanu:
@jbogard I think we should merge this, TypeMapPlanBuilder is usually in demand.

PR #1845, opened 2016-12-07
Fix TypePair so it respects the types passed in to it in the static api (for EntityFramework it uses the Proxy type no matter what you send in)

dzins:


dzins:
As an alternative, could you put code in to check if the type is an EF proxy type and if so, use the base type to look up the map?
if (type.BaseType != null && type.Namespace == "System.Data.Entity.DynamicProxies")
           {
               return type.BaseType;
           }
return type;

dzins:
Or just add an injection point to transform the type?

dzins:
Or add a configuration option to always use the types passed in instead of the run-time type?

TylerCarlson1:
Soooo.  What's the exact problem this is causing for you?  The update is causing tests to fail, so it's not viable as is.

Also could you be using CreateMissingTypeMaps by any chance?

jbogard:
No there's a direct conflict between whether or not to use the passed in
types or the runtime types to resolve a type map.

TylerCarlson1:
It should only try to map on type pairs it can find mappers for.  If you are using `CreateMissingTypeMaps = true;` then every type can be resolved and will use the proxies.
If you instead use
```
AddConditionalObjectMapper().Where((s, d) => s.Namespace != "System.Data.Entity.DynamicProxies" && d.Namespace != "System.Data.Entity.DynamicProxies");
```

It will always ignore proxy types as a means to map, but still map everything else

dzins:
Yes, we are using CreateMissingTypeMaps = true; but without that it would just error on all of the Entity Framework Proxied objects.  The problem is, we have maps for ex:
            this.CreateMap<CustomerOrder, CartModel>()
                .ForMember(x => x.OrderDate, y => y.MapFrom(x => x.Id == Guid.Empty ? null : (DateTime?)x.OrderDate.LocalDateTime))
                .ForMember(x => x.BillTo, y => y.Ignore())
                .ForMember(x => x.ShipTo, y => y.Ignore())

Entity Framework (with LazyLoading enabled) creates a Proxy class that inherits from our CustomerOrder.  So we want (anyone who is using Entity Framework and allows lazy loading / proxy creation) to say, AutoMapper, map this object, but use the type I pass in instead of getting the type from the object, which the Mapper.Map(source, destination, sourceType, destinationType) method (and Mapper.Map<TFrom, TTo>(source)) would seem to imply you can do.  We don't want it to skip the mapping, we need it to do the mapping, but using the type map for the CustomerOrder to CartModel, not the CustomerOrderProxy_XXXXXX type that EF creates.

We upgraded from AutoMapper 4.0.4 and we did not have problems with the EF Proxies then.

lbargaoanu:
Set CreateMissingTypeMaps to false. Then make a gist that we can execute and see fail. A lot of people use EF with AM, maybe you're doing smth different :)

TylerCarlson1:
If you replace `CreateMissingtypeMaps=true` with `AddConditionalObjectMapper().Where((s, d) => s.Namespace != "System.Data.Entity.DynamicProxies" && d.Namespace != "System.Data.Entity.DynamicProxies");` it should work.

All your existing mappings will work, and when you make your own it won't use the proxy type to map and will use your mapping to map.

This is exactly what you said in https://github.com/AutoMapper/AutoMapper/pull/1845#issuecomment-265588891 will do.  Conditionally ignore proxy types when mapping.

dzins:
This works!:
AddConditionalObjectMapper().Where((s, d) => s.Namespace != "System.Data.Entity.DynamicProxies" && d.Namespace != "System.Data.Entity.DynamicProxies");

So that doesn't just tell automapper to skip mapping those objects, it tells automapper something like when looking up the type maps for these objects, skip this type?

This is exactly what I was looking for, thank you so much!!

dzins:
I suggest that you add that as an example in your doc for this.  It doesn't seem obvious to me, but that it does work is fantastic

TylerCarlson1:
It's here https://github.com/AutoMapper/AutoMapper/wiki/Conventions but since AutoMapper isn't dependent fully on EF it's kinda weird putting that example into the Wiki.  I think it's more of a stack overflow answer as it stands now.  Because you shouldn't be using CreateMissingTypeMaps when your code knows the types.

And yes you are correct about skipping map types.
`AddConditionalObjectMapper().Where((s, d) => true)` is the same thing as`CreateMissingTypeMaps=true`

PR #1848, opened 2016-12-09
Avoid static state

### LINKED ISSUE #1847: InvalidOperationException : ValueFactory attempted to access the Value property of this instance.
This issue was introduced to us after updating to version 5.2.0.

We have a test helper which creates our mapper and adds profiles for us via reflection (we also tried to wrap it with a lock).

```
private static IMapper MapAutoMapper()
{
	lock (new object())
	{
		var profiles = from assembly in AppDomain.CurrentDomain.GetAssemblies().Where(x => x.FullName.StartsWith("ProjectName."))
					   from profileType in assembly.GetExportedTypes()
					   where typeof(Profile).IsAssignableFrom(profileType) && !profileType.GetTypeInfo().IsAbstract
					   select (Profile)Activator.CreateInstance(profileType);

		{
			foreach (var profile in profiles)
			{
				cfg.AddProfile(profile);
			}
		});

		return config.CreateMapper();
	}
}
```

This often throws the exception down below on our build server (TeamCity). This should be related to a multi-threading issue as Lazy<> is involved which is not thread-safe, right?

```
System.InvalidOperationException : ValueFactory attempted to access the Value property of this instance.
   at System.Lazy`1.CreateValue()
   at System.Lazy`1.LazyInitValue()
   at AutoMapper.Mappers.ConvertMapper.MapExpression(TypeMapRegistry typeMapRegistry, IConfigurationProvider configurationProvider, PropertyMap propertyMap, Expression sourceExpression, Expression destExpression, Expression contextExpression)
   at AutoMapper.Execution.TypeMapPlanBuilder.MapExpression(TypeMapRegistry typeMapRegistry, IConfigurationProvider configurationProvider, TypePair typePair, Expression sourceParameter, Expression contextParameter, PropertyMap propertyMap, Expression destinationParameter)
   at AutoMapper.Execution.TypeMapPlanBuilder.CreatePropertyMapFunc(PropertyMap propertyMap)
   at AutoMapper.Execution.TypeMapPlanBuilder.TryPropertyMap(PropertyMap propertyMap)
   at AutoMapper.Execution.TypeMapPlanBuilder.CreateAssignmentFunc(Expression destinationFunc, Boolean constructorMapping)
   at AutoMapper.Execution.TypeMapPlanBuilder.CreateMapperLambda()
   at AutoMapper.TypeMap.Seal(TypeMapRegistry typeMapRegistry, IConfigurationProvider configurationProvider)
   at AutoMapper.MapperConfiguration.Seal()
   at AutoMapper.MapperConfiguration..ctor(MapperConfigurationExpression configurationExpression, IEnumerable`1 mappers)
   at AutoMapper.MapperConfiguration..ctor(Action`1 configure)
   at Shared.TestHelper.WithFakes.MapAutoMapper() in C:\Something\Something\Project\Path\WithFakes.cs:line xx
```

Any ideas?

lbargaoanu:
Closes #1847.

### DIFF
--- a/src/AutoMapper/Mappers/MapperRegistry.cs
+++ b/src/AutoMapper/Mappers/MapperRegistry.cs
@@ -4,7 +4,7 @@ namespace AutoMapper.Mappers
 {
     internal static class MapperRegistry
     {
-        public static IEnumerable<IObjectMapper> Mappers() => new IObjectMapper[]
+        public static IList<IObjectMapper> Mappers() => new List<IObjectMapper>
         {
             new NullableSourceMapper(),
             new ExpressionMapper(), 
--- a/src/AutoMapper/Mappers/MapperRegistry.cs
+++ b/src/AutoMapper/Mappers/MapperRegistry.cs
@@ -4,7 +4,7 @@ namespace AutoMapper.Mappers
 {
     public static class MapperRegistry
     {
-        public static IObjectMapper[] Mappers() => new IObjectMapper[]
+        public static IEnumerable<IObjectMapper> Mappers => new IObjectMapper[]
         {
             new NullableSourceMapper(),
             new ExpressionMapper(), 
--- a/src/AutoMapper/Mappers/MapperRegistry.cs
+++ b/src/AutoMapper/Mappers/MapperRegistry.cs
@@ -2,9 +2,9 @@ using System.Collections.Generic;
 
 namespace AutoMapper.Mappers
 {
-    public static class MapperRegistry
+    internal static class MapperRegistry
     {
-        public static IEnumerable<IObjectMapper> Mappers => new IObjectMapper[]
+        public static IEnumerable<IObjectMapper> Mappers() => new IObjectMapper[]
         {
             new NullableSourceMapper(),
             new ExpressionMapper(), 
--- a/src/AutoMapper/Mappers/MapperRegistry.cs
+++ b/src/AutoMapper/Mappers/MapperRegistry.cs
@@ -4,7 +4,7 @@ namespace AutoMapper.Mappers
 {
     public static class MapperRegistry
     {
-        private static readonly IObjectMapper[] _initialMappers =
+        public static IObjectMapper[] Mappers() => new IObjectMapper[]
         {
             new NullableSourceMapper(),
             new ExpressionMapper(), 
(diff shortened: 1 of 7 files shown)


TylerCarlson1:
This will break functionality in AutoMapper.Collections because it has the collection profile add 2 object mappers in specific locations in the mapper configuration list.

I understand this is technically better implementation in that the list resets on every mapper configuration created, but I'm curious is this a problem for people.

lbargaoanu:
See the issue. You just have to build the list yourself.

TylerCarlson1:
I don't think #1847 has anything to do with this.  Could be wrong though.

TylerCarlson1:
Ok I see what you mean now.  But alternatively can you make the mappers themselves thread safe as an alternative?  That way there's no need to new up new classes every time?

@jbogard what do you think?

lbargaoanu:
There is no need to share them. They are created only when a new configuration is created.

TylerCarlson1:
Only thing I can think of is if you make ConvertMapper thread safe, you save on having to re-initialize Convert functions for unit tests, saving time generating expressions over and over again for each test.

lbargaoanu:
Actually they are lazy now.

TylerCarlson1:
They are lazy but if you are using a lot of converters in your mappings, you will have to lazy load them each test.  You won't load all of them (hopefully), but if you do then for each test you will have to redo it each time.

How much time this will take and if this is a priority or not I don't know.

lbargaoanu:
I don't see a difference in our tests. And anyway, we're not going to optimize for that now :)

lbargaoanu:
@jbogard Can we merge this? I want to direct people to the MyGet build.

jbogard:
@lbargaoanu I left some questions, as this is a breaking change.

lbargaoanu:
Well, yes, but it's simple enough to pass the mappers. We can't keep a static property...

lbargaoanu:
Well, we could with ThreadStatic, but I suppose that's out of the question :)

jbogard:
In that case, if it's going to break, I'd rather just break it in a consistent way. This is a bit of a one-step-forwards-two-steps-back approach. I'll amend your PR...

jbogard:
I like it. @TylerCarlson1 the only downside still is that you can't access the mapper registry from a profile, that would be nice. But let's look at that in a follow-up PR. @lbargaoanu is this good to go?

lbargaoanu:
Yes.

src/AutoMapper/Mappers/MapperRegistry.cs
```diff
@@ -4,7 +4,7 @@ namespace AutoMapper.Mappers
 {
     public static class MapperRegistry
     {
-        private static readonly IObjectMapper[] _initialMappers =
+        public static IObjectMapper[] Mappers() => new IObjectMapper[]
         {
```
jbogard: Ha! I was trying out the Review feature in GitHub. Honestly don't understand the point of it.

lbargaoanu: Neither do I :)

Issue #1836, opened 2016-12-05
Mapping Dynamic objects with a List of  Dynamic objects


HubertoKusters:
Hi,

I have two Entity classes:

        private class RouteEntity
        {
            public int Id { get; set; }
            public string Name { get; set; }
            public string Description { get; set; }
            public DateTime CreationDate { get; set; }
            public string CreatedBy { get; set; }
            public DateTime ModifyDate { get; set; }
            public string ModifiedBy { get; set; }
            public string RowVersion { get; set; } 
            // Rem: Will be byte[], but is not jet supported by AutoMapper
        }

        private class CarrierModeEntity
        {
            public int Id { get; set; }
            public string Name { get; set; }
            public string Description { get; set; }
            public DateTime CreationDate { get; set; }
            public string CreatedBy { get; set; }
            public DateTime ModifyDate { get; set; }
            public string ModifiedBy { get; set; }
            public string RowVersion { get; set; } 
            // Rem: Will be byte[], but is not jet supported by AutoMapper
            public List<RouteEntity>  Routes { get; set; }
        }

And two Domain Classes:

        private class RouteDomain : Expando
        {
            public string Name { get; set; }
            public string Description { get; set; }
        }

        private class CarrierModeDomain : Expando
        {
            public string Name { get; set; }
            public string Description { get; set; }
            public List<RouteDomain> Routes { get; set; }
        }

The Domain classes inherit from Rick Strahl's Expando object (https://github.com/RickStrahl/Expando), which in essence is a DynamicObject with some sugar. 

In a test I create the following instances for the Entity classes:

            var routeEntity1 = new RouteEntity
            {
                Name = "ROUTE1",
                Description = "Route 1",
                Id = 100,
                CreatedBy = "unittest",
                CreationDate = DateTime.UtcNow,
                ModifiedBy = "unittest",
                ModifyDate = DateTime.UtcNow,
                RowVersion = Convert.ToBase64String(
                                new byte[] {0x00, 0x00, 0x00, 0x00, 0xfa, 0x23, 0x45, 0xf3}),
            };

            var routeEntity2 = new RouteEntity
            {
                Name = "ROUTE2",
                Description = "Route 2",
                Id = 101,
                CreatedBy = "unittest",
                CreationDate = DateTime.UtcNow,
                ModifiedBy = "unittest",
                ModifyDate = DateTime.UtcNow,
                RowVersion = Convert.ToBase64String(
[...]

PR #1850, opened 2016-12-13
Mapping dynamic objects with lists of dynamic objects

HubertoKusters:
Added the ability to map dynamic objects with lists of dynamic objects to class instances  and back again.
Addresses issue #1836

### DIFF
--- a/src/AutoMapper/Mappers/DynamicMappers.cs
+++ b/src/AutoMapper/Mappers/DynamicMappers.cs
@@ -1,42 +1,66 @@
 using System;
-using System.Collections.Generic;
-using System.Dynamic;
 using System.Linq;
 using System.Linq.Expressions;
 using System.Reflection;
 using System.Runtime.CompilerServices;
+using AutoMapper.Configuration;
 using Microsoft.CSharp.RuntimeBinder;
 using Binder = Microsoft.CSharp.RuntimeBinder.Binder;
 
 namespace AutoMapper.Mappers
 {
-    using Execution;
-
     public class FromDynamicMapper : IObjectMapper
     {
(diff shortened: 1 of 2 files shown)


lbargaoanu:
I'm sorry, but this code is arbitrary. I'm sure it solves your particular case, but it doesn't really make sense in the general case.

HubertoKusters:
What is necessary in your view to make it not so arbitrary? I try to understand what is wrong with this solution?

jbogard:
In that it only does a collection. Perhaps instead it should defer to the context Mapper?

lbargaoanu:
This is a corner case anyway, you can simpy replace the dynamic mappers with your versions in the mappers collection.

HubertoKusters:
I agree with @jbogard that it only handles a collection. I think it is reasonable to assume that the handling of arrays and dictionaries should be included as well. I did not do that jet, because I was not sure if my solution is the way to go. But I had to give some implementation to get this started.

I do not agree with @lbargaoanu that it is an edge case. The dynamic mapper is capable of mapping A to X and B to Y. But it is not capable to map A (containing a list of B) to X (containing a list of Y) and that is short-coming I try to implement. 

I know that you one can implement a custom mapper and insert it in the mappers collection. But doing so would require duplicating a lot of your utility extensions, which I automatically have to my disposal when implementing this within AutoMapper.

That being said I totally agree that a well-used library like the AutoMapper should not be polluted with code the has no place in it. If you think that this is the case, I will retract the pull-request. ;)

lbargaoanu:
AM can map it if you type it object or if you don't mix dynamic with typed members. So of course it's an edge case :) And it's not obvious why AM would prefer dynamic instead of the type you specify. I'm sure some people would find that strange.

PR #1854, opened 2016-12-16
Preserve references bug

### LINKED ISSUE #1794: Performance issues and PreserveReferences not working correctly
We have been using AutoMapper 4.2.1 in our project to map our EF database model to DTO's which are 1-1 mapping. So our model has references in that way that child has reference to parent, same as parent has reference to children. Beside 1 to many relationship, we also have many to many relationship. After we upgraded AutoMapper to 5.1.1, we refactored code to use instance of IMapper which we create by calling config.CreateMappings() so no more usage of static methods from AutoMapper. We are experiencing stackoverflow exceptions. We applied PreserveReferences on each mapping in configuration, in order to get same behavior as with version 4.2.1 but we still get Stackoverflow in some cases. We also tried applying MaxDepth which helped in one case, but in another case we still get stackoverflow. My question is, if PreserveReferences should enable AutoMapper to track mapped objects using local hashtable, why doesn't it work in v5.1.1 if I enabled PreserveReferences  for every mapping configuration?

Edit: Another issue we noticed is that mapping performance is much slower with either PreserveReferences   or MaxDepth(1) enabled then with AutoMapper 4.2.1

lbargaoanu:
I'm not sure about that check, no tests fail :)
Fixes #1794.

### DIFF
--- a/src/AutoMapper/Execution/TypeMapPlanBuilder.cs
+++ b/src/AutoMapper/Execution/TypeMapPlanBuilder.cs
@@ -213,11 +213,7 @@ namespace AutoMapper.Execution
                 var condition = Condition(
                     AndAlso(
                         NotEqual(_source, Constant(null)),
-                        AndAlso(
-                            Equal(_initialDestination, Constant(null)),
-                            Call(Property(_context, "InstanceCache"),
-                                typeof(Dictionary<object, object>).GetDeclaredMethod("ContainsKey"), _source)
-                            )),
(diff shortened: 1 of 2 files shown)


lbargaoanu:
I guess the idea was that when a destination is passed, there's no reason to check the cache, because you have to return the passed destination. I believe now we do exactly that, but through the cache. The fix is simple enough, I'd rather leave it as it is.

PR #1858, opened 2016-12-19
Make the constructor parameter resolver type safe; fixes #1857

### LINKED ISSUE #1857: Exception when mapping null to `Nullable` for ctor param
Consider this:
```csharp
public class Dest
{
    public int? Value1 { get; }

    public Dest(int? thing)
    {
        this.Value1 = thing;
    }
}

public class Source
{
    public int? Value { get; set; }
}

public static void Main(string[] args)
{
    Mapper.Initialize(ctx =>
        ctx.CreateMap<Source, Dest>()
            .ForCtorParam("thing", c => c.ResolveUsing(s => s.Value)));

    Mapper.Map<Dest>(new Source { Value = 1 }); // This works
    Mapper.Map<Dest>(new Source { Value = null }); // This throws "Missing type map Object -> Nullable`1"
}
```
The reason for this is that `ICtorParamConfigurationExpression` forces the result of `s => s.Value` to be `object` and when building the expression for the ctor parameter ([here](https://github.com/AutoMapper/AutoMapper/blob/c968f34e58eed73e7cb40e8e52849408693eddb6/src/AutoMapper/ConstructorParameterMap.cs#L43)), it tries to convert `object` to `Nullable` (using configuration, as there is no matching `IObjectMapper`). This works when the result is non-null (`AssignableMapper` may be used when performing the mapping, because runtime types are known), but not when the value is null (even though it is convertible to `Nullable`, but the mapper does not know it).

Of course, I am aware that `MapFrom` should be used (which works as expected) in this example, but it is not sufficient in some scenarios (e.g. using `ResolutionContext` to get value from the `Items` collection).

I see a couple of fixes:
 * Create `IObjectMapper` specifically for this purpose - i.e. mapping `object` to `Nullable` when source is `null` and using other mapper otherwise (just like `NullableSourceMapper`),
 * Provide `ICtorParamConfigurationExpression.MapFrom<TDest>(Expression<Func<TSource, ResolutionContext, TDest>> resolver)` and make another `if` in [ConstructorParameterMap.ResolveSource](https://github.com/AutoMapper/AutoMapper/blob/c968f34e58eed73e7cb40e8e52849408693eddb6/src/AutoMapper/ConstructorParameterMap.cs#L46-L69),
 * Switch `ICtorParamConfigurationExpression.Resolve` to use `Expression`s - rather bad (will break things),
 * Handle `null -> Nullable` conversion right in [ConstructorParameterMap.CreateExpression](https://github.com/AutoMapper/AutoMapper/blob/c968f34e58eed73e7cb40e8e52849408693eddb6/src/AutoMapper/ConstructorParameterMap.cs#L41-L43) - bad, too.

I can prepare the fix, but I really don't know which way is the most appropriate.

lbargaoanu:


jbogard:
Not sure how we missed this one...

PR #1872, opened 2016-12-28
Match on requested types if runtime types don't match; fixes #1865

### LINKED ISSUE #1865: Mapping nullable does not use nullable map when called with a non-null value
Hi,

After upgrading to 5.2.0 I am having issues with mapping nullables: A map configured for a nullable type does not seem to be used if map is called with a non-null value, even if the nullable type is explicitly stated in the map method. We have several cases where we have a nullable enum mapped to a regular enum with special handling of the null and some transformation of other values. For these all non-null values are mapped using the standard mapping. The below simplified example shows it also fails for nullable ints. This issue seems to have been introduced between 5.1.1 and 5.2.0. Is this change of behavior intentional, and will I have to split my mapping into two?

```C#
[TestClass]
public class NullableMappingTests
{
	public enum GreekLetters
	{
		Alpha = 11,
		Beta = 12,
		Gamma = 13
	}

	[TestMethod]
	public void Convert_using_is_not_called_when_mapping_nullable_ints()
	{
		var mapper = new MapperConfiguration(cfg => cfg
				.CreateMap<int?, GreekLetters>()
				.ConvertUsing(n => n == null ? GreekLetters.Beta : GreekLetters.Gamma))
			.CreateMapper();

		Assert.AreEqual(GreekLetters.Beta, mapper.Map<int?, GreekLetters>(null), "This passes");
		Assert.AreEqual(GreekLetters.Gamma, mapper.Map<int?, GreekLetters>(42), "This Fails");
	}
}
```

lbargaoanu:
Maybe this would make sense in other places too, but the cases I saw before were similar.

### DIFF
--- a/src/AutoMapper/MapperConfiguration.cs
+++ b/src/AutoMapper/MapperConfiguration.cs
@@ -101,7 +101,7 @@ namespace AutoMapper
 
         public LambdaExpression BuildExecutionPlan(MapRequest mapRequest)
         {
-            var typeMap = ResolveTypeMap(mapRequest.RuntimeTypes);
+            var typeMap = ResolveTypeMap(mapRequest.RuntimeTypes) ?? ResolveTypeMap(mapRequest.RequestedTypes);
             if(typeMap != null)
             {
                 return GenerateTypeMapExpression(mapRequest, typeMap);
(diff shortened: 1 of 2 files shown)


jbogard:
Well that was easier than I thought.

jbogard:
Will this also help that one issue that the person wanted to use the explicitly used generic arguments instead of the runtime types?

lbargaoanu:
Yes, I'd say so. We'll see :)

PR #1876, opened 2017-01-02
Add a per member option DoNotInline

lbargaoanu:
From https://github.com/AutoMapper/AutoMapper/issues/1794#issuecomment-269703133.
It does happen without collections too. It's about how big the execution plan becomes. I guess you should pick and choose where you don't inline, but I didn't bother with that here.


TylerCarlson1:
Is there any metric of how much faster it is on first map?  I would think it wouldn't be all that much faster if you have an item with all its collections filled as the first map.  Because you will have to make all the maps anyway it's just in pieces instead of all at once.

Also there's going to be a performance hit for every item in the list cause you will have to find the right map func in the typemapplan cache.  It's a give and take on performance I know.  I'm just curious of how much benefit it actually gives.

lbargaoanu:
Huge. Try the tests for yourself. It's not linear, after a certain code size, stuff hits the fan :) The depth of it might make things worse.

TylerCarlson1:
ok thanks that's all I really wanted to know.  I'll take your word on it :)

jbogard:
This one good to go?

lbargaoanu:
Yes.

src/AutoMapper/PropertyMap.cs
```diff
@@ -32,6 +31,7 @@ public PropertyMap(PropertyMap inheritedMappedProperty, TypeMap typeMap)
 
         public IEnumerable<MemberInfo> SourceMembers => _memberChain;
 
+        public bool Inline { get; internal set; } = true;
```
jbogard: I don't think there's a need to make this internal - none of the other properties are?

lbargaoanu: Yes, I meant to do it and forgot.

lbargaoanu: Done.

lbargaoanu:
For future reference. Another approach here, less ad-hoc, would be to use smth like max depth, with a reasonable default, maybe 10. The problem would then be that the context map thing would need to behave _exactly_ as the inlined version. This doesn't happen today because of declared types/runtime types and because the configuration is not passed from the initial call to the context map call. Certainly more involved, but it would work by default. What we have now seems a reasonable compromise, but I was thinking about this and wanted to leave a record.

Issue #1866, opened 2016-12-23
Need to get ProfileConfiguration before sealed


TylerCarlson1:
AutoMapper.Collection.EF needs the configuration to get property maps for types so that it can auto generate comparisons.  If you use a convention to do mappings there's no problem, because the type maps are generated after the seal where AM.Collection.EF can resolve the type maps and get generated.

Problem is if you explicitly do the mappings they will get sealed and not be able to resolve equivalency using AutoMapper.Collection.EF.  It can't be used because by the time it can get set all the type maps have been sealed and can't be changed.

What I'm suggesting is have a pre-sealed eventhandler that passes IProfileConfiguration, so I can hook into to set up AutoMapper.Collection.EF to use it before types get sealed.  Or possibly an action that can be set in IMapperConfigurationExpression that gets called before sealed.  That way before the expressions are generated, AutoMapper.Collection.EF can be set up to use said maps to update collections.

Right now the work around is to create/initialize IMapper twice.  Once to make the maps in which AutoMapper.Collection.EF can use the configuration.  And a second time to re-do the mappings, but allowing AutoMapper.Collection.EF to do its thing.

In 4.2.x this wasn't a problem because of how the nature of AM worked before.  Type maps weren't sealed until used and that way you could put them after the configuration without issue.

I'm willing to make a PR for this,  I just wanted to explain what's going on first.

jbogard:
What exactly are you doing in the comparison generation? 

I used to have an event a while back, but I thought there could be a better way of doing this, would the ForAllTypeMaps stuff not work?

TylerCarlson1:
When generating the mapping expression of Parent to ParentDto and you get to a property that maps `IList<Child> -> IList<ChildDto>` I need to resolve the TypeMap Child->ChildDto so that I can identify primary keys in ChildDto and generate an equivalency expression based on that property maps.  Currently I resolve the type map by getting the IConfigurationProvider and calling ResolveMap and then seeing if it matches by keys.

Currently there's no way to make it work with the ForAllTypeMaps function.  I guess I could make it work to some extent though, given it gets executed before generating map expressions.

That's not how I intended it to work at all, but I might be able to do that.  Although right now it seems kind of a hack to get it to work with the public API as it stands now.

jbogard:
Is there something we can do BESIDES an event? Those are just...icky which is why I took them out. Is there something in addition to ForAllTypeMaps, like a BeforeSeal or similar?

[...]

PR #1883, opened 2017-01-04
Add Advanced property and put BeforeSeal function in it

TylerCarlson1:
As discussed in #1866


lbargaoanu:
A test would help.

lbargaoanu:
Maybe it's just me, but I don't understand the test :) Maybe it's even correct, but I cannot tell by reading the code.

TylerCarlson1:
Oh yea you are right I didn't even notice that.  By default _sealed is false and doesn't really prove the method was called.
I can make _sealed nullable, so you know it got set.

lbargaoanu:
It makes sense.

src/AutoMapper/AdvancedConfiguration.cs
```diff
@@ -0,0 +1,12 @@
+using System;
+
+namespace AutoMapper
+{
+    public class AdvancedConfiguration
+    {
+        /// <summary>
+        /// Action called against the IConfigurationProvider before it gets sealed
+        /// </summary>
+        public Action<IConfigurationProvider> BeforeSeal { get; set; }
```
TylerCarlson1: Yea I thought about that.  But I got the thing to work with AM.Collections with a single action, so haven't changed it yet.

src/AutoMapper/MapperConfiguration.cs
```diff
@@ -37,7 +37,8 @@ public MapperConfiguration(MapperConfigurationExpression configurationExpression
             Configuration = new ProfileMap(configurationExpression);
             Profiles = new[] { Configuration }.Concat(configurationExpression.Profiles.Select(p => new ProfileMap(p, configurationExpression))).ToArray();
 
-            configurationExpression.Advanced.BeforeSeal?.Invoke(this);
+            foreach (var beforeSealAction in configurationExpression.Advanced.BeforeSealActions)
+                beforeSealAction?.Invoke(this);
```
lbargaoanu: I think this code should be inside Advanced and then you don't need to expose the list.

TylerCarlson1: Everything else goes through a foreach loop of items in this class.  Like profiles you can see the property but to add you call add function, so what's the difference?  It's exposed as an enumerable as to not add another way.  I don't see changing it as "that" big of a deal in exposing it or not.

lbargaoanu: It's encapsulated, that's the difference. An event does this for example.

TylerCarlson1:
Is this good @jbogard or is the before seal action implementation blocking it?

jbogard:
I might do a wee bit of refactoring here but I don't want to block this one.

PR #1889, opened 2017-01-05
Multiple value resolver interfaces

### LINKED ISSUE #1887: To support resolver that implements multiple IValueResolver interfaces
Not sure if this is a bug or simply an unsupported feature, if a resolver implements multiple IValueResovler interfaces, AutoMapper throws an "No coercion operator is defined between types" exception. The following code demonstrates the problem.
`

    public class A { }
    
    public class B { }

    public class C
    {
        public string property1 { get; set; }
    }
    
    public class MyTestResolver :
        IValueResolver<A, C, string>,
        IValueResolver<B, C, string>
    {
        public string Resolve(A source, C destination, string destMember, ResolutionContext context)
        {
            return "from A->C";
        }

        public string Resolve(B source, C destination, string destMember, ResolutionContext context)
        {
            return "from B->C";
        }
    }

    class Program
    {
        {
            AutoMapper.Mapper.Initialize(x =>
            {
                x.CreateMap<B, C>()
                    .ForMember(dest => dest.property1, opt => opt.ResolveUsing<MyTestResolver>());
            });
        }
    }
`

lbargaoanu:
Fixes #1887.
Of course we _don't have to_ support that scenario :)

### DIFF
--- a/src/AutoMapper/Configuration/MappingExpression.cs
+++ b/src/AutoMapper/Configuration/MappingExpression.cs
@@ -101,14 +101,14 @@ namespace AutoMapper.Configuration
 
             public void ResolveUsing(Type valueResolverType)
             {
-                var config = new ValueResolverConfiguration(valueResolverType);
+                var config = new ValueResolverConfiguration(valueResolverType, valueResolverType.GetGenericInterface(typeof(IValueResolver<,,>)));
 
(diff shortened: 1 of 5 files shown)


lbargaoanu:
So the feature is out? Because either way I think the refactorings are worth doing.

PR #1895, opened 2017-01-07
Use ReverseTypeMaps when configuring closed generic TypeMaps

duncanbrown:
Currently, `ReverseMap` and open generics don't work together - the new test in this commit would fail with a 'Missing type map configuration' exception. This fixes that by including reverse maps when configuring closed generic maps.


jbogard:
Is there not an open GH issue for this? Not a big deal if not, just thought I saw one.

duncanbrown:
I don't think so; couldn't see any existing issues, and I didn't create one.

PR #1901, opened 2017-01-11
Preserve references

### LINKED ISSUE #1892: Best way to preserve compatibility for v4 to v5 upgrade
I'm in the process of upgrading to v5 and discovering that we have hundreds of classes with cyclical references so I would like to globally set the PreserveReferences and UseDestinationValue options so I don't have to spend time tracking down which classes need this and which don't (or manually adding this call to hundreds of mappings)

Is this possible or should i just stick with v4 until the end of time?

lbargaoanu:
Fixes #1892. Closes #1894.
I guess it's a breaking change, but nobody should be using InstanceCache anyway. We could make it internal, or even private (private seems best). Funny enough, I wasn't able to make out parameters work, and not for lack of trying. It worked in a simple test, but not in AM.
I took GetHashCode from System.Tuple. I guess it's close to the idea you used in TypePair. We should have a helper to be used everywhere. It's not in the framework yet I believe. I don't think we need to cache the hash code when the fields are read-only. Tuple doesn't. That saves the field and avoids computing it when it's not needed (granted, that's just a few instructions).

### DIFF
--- a/src/AutoMapper/Execution/TypeMapPlanBuilder.cs
+++ b/src/AutoMapper/Execution/TypeMapPlanBuilder.cs
@@ -201,18 +201,15 @@ namespace AutoMapper.Execution
             if(_typeMap.PreserveReferences)
             {
                 var cache = Variable(_typeMap.DestinationTypeToUse, "cachedDestination");
-                var hasDestination = _context.Type.GetDeclaredMethod("HasDestination");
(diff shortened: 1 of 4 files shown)


lbargaoanu:
I took GetHashCode from System.Tuple. I guess it's the same idea as the one you used in TypePair. Maybe we should have a helper to be used by both. It's not in the framework yet I believe.

Issue #1821, opened 2016-11-25
Exception when mapping expression with a property that maps to a string type


arknu:
Say I have two types like this:

```
public class ItemDto
{
	public int Id { get; set; }

	public string Name { get; set; }

	public DateTime CreateDate { get; set; }
}
```

and

```
public class Item
{
	public int Id { get; set; }

	public string Name { get; set; }

	public DateTime Date { get; set; }
}

```

If I then do this, I would expect the expression to be mapped correctly:

```
Mapper.Initialize(cfg =>
{
	cfg.CreateMap<ItemDto, Item>()
	.ForMember(dest => dest.Date, opts => opts.MapFrom(x => x.CreateDate));
});

Expression<Func<Item, object>> exp = x => x.Name;

var mapped = Mapper.Map<Expression<Func<ItemDto, object>>>(exp);
```

Instead, I get an exception:

> System.InvalidCastException: 'An object of the type 'System.Linq.Expressions.Expression`1[System.Func`2[AutomapperTest.Models.ItemDto,System.String]]' cannot be converted to the type 'System.Linq.Expressions.Expression`1[System.Func`2[AutomapperTest.Models.ItemDto,System.Object]]'.'

However, when mapping an expression for the Date/CreateDate property, there is not problem.

The difference seems to be that the expression generated for the string property has a body of type `MemberAccess`, whereas the expression for the Date property has a body of type `Convert`

TylerCarlson1:
I would suggest explicitly casting the string to object and see if that works. Also if you made return type string instead of object that would also avoid an exception.

With expressions you need to be very explicit on types. Can't assume the types will figure themselves out

Sent from my iPhone

On Nov 25, 2016, at 11:47 AM, Asbj?rn Riis-Knudsen <notifications@github.com<mailto:notifications@github.com>> wrote:


Say I have two types like this:

public class ItemDto
{
        public int Id { get; set; }

        public string Name { get; set; }

        public DateTime CreateDate { get; set; }
}


and

public class Item
{
        public int Id { get; set; }

        public string Name { get; set; }

        public DateTime Date { get; set; }
}



If I then do this, I would expect the expression to be mapped correctly:

Mapper.Initialize(cfg =>
{
        cfg.CreateMap<ItemDto, Item>()
        .ForMember(dest => dest.Date, opts => opts.MapFrom(x => x.CreateDate));
});

Expression<Func<Item, object>> exp = x => x.Name;

[...]

Issue #1893, opened 2017-01-07
System.NullReferenceException Mapping Expression<Func<T,bool>> with flattening


drjameshouse:
I'm trying to map an Expression<Func<T,bool>> where the source is a complex object that gets flattened in the destination.

Models:
```
public class OrderLine
{
	public int Id { get; set; }
	public string ItemName { get; set; }
	public int Quantity { get; set; }
}

public class OrderLineDTO
{
	public int Id { get; set; }
	public ItemDTO Item { get; set; }
	public int Quantity { get; set; }
}

public class ItemDTO
{
	public int Id { get; set; }
	public string Name { get; set; }
}
```

Configuration:
```
Mapper.Initialize(cfg =>
{
	cfg.CreateMap<OrderLine, ItemDTO>()
		.ForMember(dto => dto.Name, conf => conf.MapFrom(src => src.ItemName));
	cfg.CreateMap<OrderLine, OrderLineDTO>()
		.ForMember(dto => dto.Item, conf => conf.MapFrom(ol => ol));
	cfg.CreateMap<OrderLineDTO, OrderLine>()
		.ForMember(ol => ol.ItemName, conf => conf.MapFrom(dto => dto.Item.Name));
});
```

And finally the Expression:
```
Expression<Func<OrderLineDTO, bool>> dtoExpression = dto => dto.Item.Name == "Item #1";
var expression = Mapper.Map<Expression<Func<OrderLine, bool>>>(dtoExpression);
```

Exception:

> AutoMapper.AutoMapperMappingException was unhandled
>   HResult=-2146233088
>   Message=Error mapping types.
> 
> Mapping types:
> Expression`1 -> Expression`1
[...]

Issue #1886, opened 2017-01-04
Mapping of Expression<Func<T,bool>> causes exception starting from AutoMapper 5.2.0


vukispasic:
Starting from AutoMapper version 5.2.0 we can't execute mapped queries anymore. Exception is "Additional information: variable 'r' of type 'OmmitedDTOModel3.EntityDTO1' referenced from scope '', but it is not defined". In production code we are using EF so execution of EF query results in exception "The parameter 'r' was not bound in the specified LINQ to Entities query expression". AutoMapper 5.1.1 works fine, as well as versions 4

I made a repro:
https://github.com/vukispasic/AutoMapperExpressionMappingIssue.

TylerCarlson1:
Can you try the [MyGet Build](https://github.com/AutoMapper/AutoMapper/wiki/The-MyGet-build)?

vukasinspasic:
Already tried, same issue

TylerCarlson1:
https://github.com/AutoMapper/AutoMapper/commit/b22ce56ac64a7a2d2ccd2ecb8ef69302c6cee63e#diff-51fc29521ce7d80eddd4a9ef366b9d09R300 Here's the problem most likely.  I'm guessing it needs to get replaced with IsAssignableFrom, because it only fails on Id in your repo, which is a property inherited from the base entity class.

vukasinspasic:
Thank you

vukispasic:
Hello. I fixed the issue as TylerCarlson1 proposed. I also created unit test. How can I publish from my local branch and make pull request? I get 403 (Forbidden). Thanks

TylerCarlson1:
You have to fork the repository in github and push the branch there. Then from that remote branch you can request a PR on github

Sent from my iPhone

On Jan 14, 2017, at 1:32 PM, vukispasic <notifications@github.com<mailto:notifications@github.com>> wrote:


Hello. I fixed the issue as TylerCarlson1 proposed. I also created unit test. How can I publish from my local branch and make pull request? I get 403 (Forbidden). Thanks

-

vukasinspasic:
I made pull request https://github.com/AutoMapper/AutoMapper/pull/1911 as instructed

yasin-uohyd:
Hi @jbogard, when are we going to release in nuget.org

cjrogala:
[...]

PR #704, opened 2015-04-12
ExpressionMapping assemblies with sample AdventureWorks tiered solution.

BlaiseD:
I've created a pull request with actual files you can merge this time.  The changes allow:
1) Mapping of IQueryable expressions Expression<Func<IQueryable<UserModel>, IQueryable<UserModel>>> exp; to Expression<Func<IQueryable<User>, IQueryable<User>>> exp;.
2) Because of 1), provides an alternative to using Project().To().

It includes sample projects for AdventureWorks 2008R2.  Sorry for the confusion.

Thanks.

PR #1916, opened 2017-01-17
Provide an alternative to using ProjectTo() and static initialization…  

BlaiseD:
### Goal: Provide an alternative to using ProjectTo() and static initialization for ASP.NET Core - instead using instance initialization with expression mapping.

- Maps expressions involving flattened (TModel) properties (when the navigation property has been removed from the view model) to their corresponding (TData) navigation properties. e.g. CourseModel.DepartmentName from the model expression becomes Course.Department in the data expression.
```
public class CourseModel
    {
        public int CourseID { get; set; }

        public int DepartmentID { get; set; }
        public string DepartmentName { get; set; }
    }
public class Course
    {
        public int CourseID { get; set; }

        public int DepartmentID { get; set; }
        public Department Department { get; set; }
    }

    public class Department
    {
        public int DepartmentID { get; set; }
    }
```
- Fixes issues #1821, #1886 and #1893. I agree that the use case for deflattening (#1893) is not clear but it works.
- Fixes miscellaneous expression mapping issues.  Please see the tests under XpressionMapperTests.  Many do not succeed with 5.2 code - usually when there is an extension method in the expression.

It maps all the followimg expression from TModel to TData - see the StoreHelpers class in the ContosoUniversity.Repositories project (Samples Solution folder). All calls to map expressions and objects are in this file.

- Expression<Func<TModel, bool>> filter
- Expression<Func<IQueryable<TModel>, IQueryable<TModel>>> queryFunc
- ICollection<Expression<Func<IQueryable<TModel>, IIncludableQueryable<TModel, object>>>> includeProperties

The Contoso example [here](https://github.com/BlaiseD/XpressionMapper.NetCore) shows the solution in action. It is all generics, AutoMapper and EF - hardly any reference to concrete types.

I requested PR #704 canceled a year and a half ago because I didn't think the value was clear but I believe this helps now. I don't want to replace the current one. Maybe they can run side my side with different helpers? Mapper.MapExpression vs Mapper.Map?

To test my implementation please uncomment "//#define XPRESSIONMAPPER" at the top of ExpressionMapper.cs.

My implementation does fail one test as of this writing.  **Can_map_flattened_properties** in ExpressionConversion.cs.

`var config = new MapperConfiguration(cfg => cfg.CreateMap<Source, Dest>().ForMember(dest => dest.ChildValue, opt => opt.MapFrom(src => src.Child.Value)));` works but
`var config = new MapperConfiguration(cfg => cfg.CreateMap<Source, Dest>());` fails.

ChildValue is being mapped somehow by concatenating Child.Value - I haven't figured out the source yet - the property maps only show the relationship for the first configuration.
XpressionMapperTests and ExpressionMapping tests all pass.

Please let me know.  It will be great if it can be part of AutoMapper.

Thanks,
Blaise


TylerCarlson1:
That's insanely hard to figure out at first glance.

As far as the unit test failing, the property map has all the member types.  There's SourceMember or something like that, which is the last in the list.  And then you can get the list itself to get each of the source members in order of execution from source.  I believe it aggregates all these properties into an expression.  Don't know exact lines but I'm pretty sure it's something along these lines.

Still need to look at the code more to fully understand all what's going on.

BlaiseD:
I suspect the build fails because the AutoMapper versions have been set as follows in the project.json files: "AutoMapper": { "target": "project" }.

Error messages: Unable to resolve 'AutoMapper' for '.NETCoreApp,Version=v1.1'.

BlaiseD:
Ok - cool - thanks

Sent from my iPhone

On Jan 17, 2017, at 6:10 PM, Tyler Carlson <notifications@github.com<mailto:notifications@github.com>> wrote:


That's insanely hard to figure out at first glance.

As far as the unit test failing, the property map has all the member types. There's SourceMember or something like that, which is the last in the list. And then you can get the list itself to get each of the source members in order of execution from source. I believe it aggregates all these properties into an expression. Don't know exact lines but I'm pretty sure it's something along these lines.

Still need to look at the code more to fully understand all what's going on.

BlaiseD:
@TylerCarlson1  You were right.  There's a list of source members I should have been using instead of just the "leaf" source member - thought I would have noticed before now.  It is fixed now.

BlaiseD:
Seems the build failures are not related to my PR code.  The new code was not referenced in the latest check-in.  Correct me if I'm wrong.

TylerCarlson1:
https://ci.appveyor.com/project/jbogard/automapper/build/1282-pcjnqkkv/tests you have multiple tests failing

TylerCarlson1:
Also yes probably.  You might have to merge master into your PR and then fix the failed tests

BlaiseD:
Those code was checked in to use the existing mapper i.e. those tests [(](https://ci.appveyor.com/project/jbogard/automapper/build/1282-pcjnqkkv/tests) ) were expected to fail.

All good now (with one additional tweak).  Thanks for your help.

BlaiseD:
Removed the Contoso Sample from the PR.  I think it exaggerated the size of the changes.  A sample solution can be found [here](https://github.com/BlaiseD/XpressionMapper.NetCore).

jbogard:
@BlaiseD whew! with that sample I couldn't get my head around. You could also create another branch off of this branch for a sample PR with a sample if you like.

BlaiseD:
@jbogard  I'm not sure that will help but there are only two relevant methods in the ContosoUniverity solution:

1. In the ContosoUniversity.Repositories project, **StoreHelpers.ModelQueryAsync** maps expressions (TModel to TData) on the way down and maps objects (TData to TModel) on the way up.
2. StoreHelpers.ModelQueryAsync ultimately calls **GenericRepository<T>.GetAsync** in ContosoUniversity.Crud (all TData at this point).

Finally there are unit tests in MapperTests in the XpressionMapper.Tests project.  (Same tests as in XpressionMapperTests in the PR branch)

This PR essentially takes the XpressionMapper project from the ContosoUniverity solution and moves into the AutoMapper.XpressionMapper namespace.  To test the proposed functionality the existing ExpressionMapper has been updated to call a MapExpression extention method from AutoMapper.XpressionMapper.Extensions.

The rest of the code in the ContosoUniverity solution is just everything else outside of AutoMapper and expression mapping that makes the application work.

Clear as mud?

TylerCarlson1:
@BlaiseD if you wanted to extend it as optional you could.  You could make separate dll and just make an extension method off of configuration to replace the Expression Mapper with your XpressionMapper int the Mappers list.

As far as adding the functionality I would suggest either updating existing code to support the tests that fail in this PR, or replacing the expression mapper completely with this implementation (after careful review of course).

I still haven't gotten around to looking real deep into the implementation yet.  Mostly because of the Sample solution.

I don't know about the debate with Expression translation having 1 Visitor, like it is now, or many visitors all doing one specific thing, like this PR.  They each have there problems when it comes to debugging which is a HUGE pain no matter what you do.

BlaiseD:
@TylerCarlson1 Extending it as optional has some of the same problems with the existing dynamic i.e. there are likely to be breaking changes with future versions of AutoMapper.

As far as adding the functionality I'd vote for replacing because aside from the failing tests already working in the PR there is added functionality (mapping Include expressions from flattened value types to the corresponding navigation property) which depends on the approach taken in the PR - but it all depends on what you guys think.

Please feel free to make/recommend changes.

Thanks.

BlaiseD:
@TylerCarlson1 With regard to sample code are you looking for something that falls somewhere between the test cases (XpressionMapperTests.cs in the PR) and the [sample solution](https://github.com/BlaiseD/XpressionMapper.NetCore)?

TylerCarlson1:
I haven't looked too much into the Include expressions (if that's not de-flattening), but I think 2 known issues this solves that I would want to add to the existing implementation.

1 is the casting of result type to object.  Expression of object returns a string type but doesn't cast to object in the end so can't translate. (Might be more to this but should be easy enough)
2 is De-Flattening the mapping.  This will probably be much harder to add on.

And for adding onto the existing code, I would suggest it just so that the changes, that might break thing, will be easier to identify what went wrong.  If it's a complete re-write it's hard to figure out what could of changed.

I'll try to get around to looking over it.  Maybe try to put the test and the solution in one commit so can see what changed what.

As for optional possibility @jbogard AdvancedSettings?

TylerCarlson1:
@BlaiseD I would want the test you have to all pass,  I would think the tests would sufficiently cover all the scenarios in your sample solution.

BlaiseD:
I suggested replacing because I thought you were offering it as an option (3 comments ago). The initial description above calls for using the MapExpression extension method for the new functionality.

The existing version is good work.

At the same time in addition to #1821 (expressions involving string selections and comparisons fail) and #1893 (de-flattening fails), mappings like the following fail in 5.2:

(Necessary for EF Core)
`Expression<Func<IQueryable<AccountModel>, IIncludableQueryable<AccountModel, object>>> selection = a => a.Include(x => x.ThingModels).ThenInclude(y => y.Car);
Expression<Func<IQueryable<Account>, IIncludableQueryable<Account, object>>> selectionMapped =
                mapper.Map<Expression<Func<IQueryable<Account>, IIncludableQueryable<Account, object>>>>(selection);`

(Necessary for keeeping the IQueryable instance in the data access component)
`Expression<Func<IQueryable<UserModel>, IQueryable<UserModel>>> exp = q => q.OrderByDescending(s => s.Id).ThenBy(s => s.FullName);
Expression<Func<IQueryable<User>, IQueryable<User>>> expMapped = mapper.Map<Expression<Func<IQueryable<User>, IQueryable<User>>>>(exp);`

Finally there is the additional functionality allowing us to map includes for flattened properties to navigation properties.

From what I understood, your argument against replacing the current mapping visitor is that we may break existing functionality but it seems to me there's a lot that is already broken that this PR fixes relative to what it might possibly break.

This is how software works.  We replace methods, classes and components all the time.  I wouldn't take it personally.

This PR proves that the issues can eventually be fixed in the existing code (it's clear now that there are solutions) but it also proves that the approach used in this PR works better - proofs in the pudding.

The code in this PR could also be replaced at some point.  You accept both small and larger fixes right?

Thanks.

TylerCarlson1:
I was saying it's possible.  I don't know what @jbogard wants to do (Probably still trying to wrap his head around it)
I'm not saying it has more cause to break I'm saying if something does break from like 5.2 to 5.3, then we would have a smaller amount of changes that could possibly be culprit, where as if it's a complete re-write then you have to go through the entire thing from scratch to find the issue.  Either way it's a complete pain to go through and depending on the issue it might not guarantee to be helpful.

As far as your tests I did mess around with it and in https://github.com/TylerCarlson1/AutoMapper/tree/Xpression_Test I got all your tests except `Map_includes_trim_string_nested_in_select_using_explicit_types()` which boggles my mind what you are trying to achieve.  Need to look at your code's expression I guess to know.

Some other existing tests fail, but I was focused on the new ones, but those look like simple fixes.

From what I saw it was some relatively simple issues to fix.  Use multiple genericsinstead of assuming 1.  Convert result of expression to expected type #1821.  Recursion not applying in all cases.  Assuming expressions are set up in certain order instead of trying to figure out the order and do it that way.

I'll make a PR once I got all the existing tests fixed minus the one mentioned above, so can compare.  I'm sure I'm missing something

BlaiseD:
I really wouldn't expect any of the "Map_includes_" to work.  Those are related to the new functionality,  In case of `Map_includes_trim_string_nested_in_select_using_explicit_types` you're returning a list of enumerable strings (Car.Color) wrapped in a Convert.  It should really return a list of enumerable Car.  The goal is at the top of the PR which is to allow the user (developer) to write all queries against the view model without being concerned about anything else.  Then in the case of a flattened object that does not include a navigation property (DepartmentModel in the example at the top of the PR), then how would tell the context to include Department in the TData query.  The design is to use the flattened property. MapExpressionAsInclude will map ThingModel.Color to Thing.Car rather than Thing.Car.Color (because we need the navigation property in the include statement).

Your updates seem to have worked minus without the new MapExpressiomnAsInclude functionality.

Make sense?

One thing though - I would try to create the method call or the expression using the correct type.  I think I notice conversions to object in cases where I haven't had to - just FYI.

TylerCarlson1:
Yea the include thing I didn't get at all.  I guess I'll need to look into what's all going on there.  Yea I just did convert instead of Map, which you did if I understand correctly.

I personalty don't see the point in doing that right now, because if you call Map you won't be able to translate the Expression in SQL.  I would prefer exceptions and hard failures telling you you can't do that vs letting them do whatever they want and have them fail in EF cause they aren't supported.

The only place I can see is if you compile the expressions then use them.  Maybe your sample has that, I haven't checked but I guess that's a valid concern that it's supported with maps.  Still feels weird right now (Will look into it more and see how to replicate it).

With those conversions I used ToType which checks type and if types match return it and if not do convert.  Maybe that's what you are noticing.

BlaiseD:
I haven't figured out the why but here are two examples:

**Map__select_method_projecting_to_anonymous_type**
`{s => s.Account.Things.Select(x => new <>f__AnonymousType0`1(MM = x.Bar)).Where(b => b.MM.EndsWith("3"))} `- PR 1916

`{s => Convert(s.Account.Things.Select(x => new <>f__AnonymousType0`1(MM = x.Bar)).Where(b => b.MM.EndsWith("3")))}` - AutoMapper-Xpression_Test

**Map_orderBy_thenBy_expression**
`{q => q.OrderByDescending(s => s.UserId).ThenBy(s => Concat(s.FirstName, " ", s.LastName))}` - PR 1916

`{q => Convert(q.OrderByDescending(s => s.UserId).ThenBy(s => Concat(s.FirstName, " ", s.LastName)))}` - AutoMapper-Xpression_Test

TylerCarlson1:
Maybe it's because of the type is IEnumerable and result is WhereArrayIterator and ThenByIterator and I just convert them.

jbogard:
Can someone explain the 10K ft view of this? I'm unclear about what problem this is solving?

TylerCarlson1:
10k feet view.  Current version only supports common expression mapping scenarios, and this will make it work in **EVERY** expression mapping scenario.  (At least most people could ever imagine)

Slightly lower View
Fixes
#1821 Old system does `Expression<Func<ItemDto, string>>` and this will translate to `Expression<Func<ItemDto, object>>` so that it won't throw exception.
#1893 Will De-Flatten properties to map expressions.  AM doesn't support this out and you would suggest not flattening as a workaround.  Ex: Translate `Parent.Child.GrandChild` to `ParentChildGrandChild` which isn't supported.  (`ParentChildGrandChild` to `Parent.Child.GrandChild` works just fine)

Other non-issue fixes
Multi Parameter Func support.  Things like `(m, i) => m.Value + i.Value` will work instead of throwing exception
Custom Expressions cause problems that will be fixed

Other things
IncludeMapExpression will call Mapper.Map if translated expression doesn't have correct types from expected results.  I'm iffy on this personally.

In other words this PR is like an fix all solution replacement to expression mapping that's more powerful.  Also possibly adding IncludeExpressionMapping that will call Mapper.Map if needed to make expression not throw exception.

Hopefully that helps somewhat.

BlaiseD:
Thanks for that description @TylerCarlson1 .  One correction: the MapExpressionAsInclude helper behaves just like MapExpression except that it returns the navigation property of the mapped value types (useful for getting the Include navigation property when it does not exist in the view model).

I'll add to @TylerCarlson1 overview the following:

1. Allows for a generic repository which returns collections rather than an IQueryable.  The IQueryable instance never leaves the repository.
2. All selects are AsNoTracking by default (Enforced in the Generic Repository).
3. No need for static mapper initialization with ASP.NET Core (no need for ProjectTo).
4. Makes querying the view model collections possible with basically just generics and AutoMapper in the data stack.

ProjectTo solves some of the same problems overall but with ProjectTo you need the TModel, TData and IQueryable in the same method.

This design allows all queries to be executed against the view models - see Departments,  Courses, Students and Instructors Controllers in ContosoUniverity.WebApi.

So in addition to the 4 points I've listed and more complete expression mapping I think it makes the design cleaner - all mapping of expressions and objects far away from the actual query.

I'm curious about what your opinions are of

- The overall design of the web solution from the Web API controller down
- The expression mapper itself.

Thanks much.

jbogard:
Is it possible to get this PR without the #if stuff? It's hard to follow the diffs when the old stuff is still in there.

BlaiseD:
@jbogard - done.  I'm ok with leaving ExpressionMapper.cs untouched - and just call the extension methods directly if that is an option.

jbogard:
@TylerCarlson1 anything preventing an approval and merge on this one?

@BlaiseD does any of the wiki docs need to change for this?

BlaiseD:
@jbogard  Only [Expression Translation](https://github.com/AutoMapper/AutoMapper/wiki/Expression-Translation-(UseAsDataSource)) can be affected correct?  Non of the changes make the existing documentation obsolete.  I'll add a section on mapping "includes" for flattened objects - sounds like you're planning to approve the changes?

jbogard:
Yeah, there are some style things that are different but that's no reason to hold up the PR. If there are things that this PR enables, that's mainly what I'm looking at too.

BlaiseD:
@jbogard That's more than cool. I'll update the Wiki for new functionality - seems like I have the permissions.  Interested to know what the coding style issues are.

Thanks.

jbogard:
Regions and field names. Like I said, no big deal.

TylerCarlson1:
@jbogard No not really.  I still think include will cause problems, but that doesn't effect existing functionality.  You have to know about it and search it out, but mapping in general won't suffer for it.

And maybe formatting needs changing around, but that's not that big of a deal

lbargaoanu:
@BlaiseD I only took a quick look at the code, but I think it could benefit from all the utility methods we have, usually extension methods (on Type for example). I guess it's not easy to tell what we have without knowing the AM codebase, but keep it in mind.

BlaiseD:
@lbargaoanu Thanks for the heads-up.  Yes this code has lived outside the AM code base for a while.
@jbogard I've added a section called "Mapping Flattened Properties to Navigation Properties" to [Expression Translation](https://github.com/AutoMapper/AutoMapper/wiki/Expression-Translation-(UseAsDataSource)) in the Wiki.  If it's too early we can remove the section - it is out out of sync with the latest release.

BlaiseD:
Nice!

Sent from my iPhone

On Feb 28, 2017, at 10:20 AM, Jimmy Bogard <notifications@github.com<mailto:notifications@github.com>> wrote:


Merged #1916<https://github.com/AutoMapper/AutoMapper/pull/1916>.

PR #1917, opened 2017-01-18
Reuse GetHashCode

lbargaoanu:


src/AutoMapper/TypePair.cs
```diff
@@ -32,7 +30,7 @@ public override bool Equals(object obj)
             return obj is MapRequest && Equals((MapRequest) obj);
         }
 
-        public override int GetHashCode() => _hashcode;
+        public override int GetHashCode() => HashCodeCombiner.Combine(RequestedTypes, RuntimeTypes);
```
TylerCarlson1:
```
private int? _hashcode;
public override int GetHashCode() => _hashcode ?? (_hashcode = HashCodeCombiner.Combine(RequestedTypes, RuntimeTypes)).Value;
```
Every so slightly optimized by caching like original code.
Same with other implementations and each map does a couple of comparisons.
And inline collections will have to compare again as well.
Might be significant.

src/AutoMapper/TypePair.cs
```diff
@@ -185,6 +180,13 @@ public int Compare(Type x, Type y)
...
             }
         }
+    }
 
+    public static class HashCodeCombiner
+    {
+        public static int Combine<T1, T2>(T1 obj1, T2 obj2) =>
+            CombineCodes(obj1.GetHashCode(), obj2.GetHashCode());
+
+        public static int CombineCodes(int h1, int h2) => ((h1 << 5) + h1) ^ h2;
```
TylerCarlson1:
```
public static int Combine(object obj1, object obj2, params object[] addObjs) =>
            addObjs.Aggregate(obj1.GetHashCode().CombineCodes(obj2.GetHashCode()), (code, obj) => code.CombineCodes(obj.GetHashCode()));

public static int CombineCodes(this int h1, int h2) => ((h1 << 5) + h1) ^ h2;
```

Maybe make Combine extension as well and rename CombineHashCode.
Aggregate maybe too hard to read, just wanted to keep it one line.

src/AutoMapper/QueryableExtensions/ExpressionRequest.cs
```diff
@@ -54,12 +54,12 @@ public override bool Equals(object obj)
...
-            {
-                var hashCode = SourceType.GetHashCode();
-                hashCode = (hashCode*397) ^ DestinationType.GetHashCode();
-                return MembersToExpand.Aggregate(hashCode, (currentHash, p) => (currentHash * 397) ^ p.GetHashCode());
+            var hashCode = HashCodeCombiner.Combine(SourceType, DestinationType);
+            foreach(var member in MembersToExpand)
+            {
+                hashCode = HashCodeCombiner.CombineCodes(hashCode, member.GetHashCode());
             }
+            return hashCode;
```
TylerCarlson1:
```
public override int GetHashCode() => HashCodeCombiner.Combine(SourceType, DestinationType, MembersToExpand);
```
as per refactoring note.

Also can cache, but I don't think you get much performance gain out of it.

src/AutoMapper/TypePair.cs
lbargaoanu: That's just wasteful. See https://github.com/AutoMapper/AutoMapper/pull/1901.

src/AutoMapper/TypePair.cs
lbargaoanu: That's not good for performance :) GetHashCode should be fast and not allocate.

src/AutoMapper/TypePair.cs
TylerCarlson1: Hmm I wouldn't say that Tuple did it guarantees that not caching it give the best performance.
Either way I don't think it's THAT critical to performance.  The only way to tell is make couple hundred type maps and map each one and then you can see if performance degrades, as it checks each has code for every type map and map request to find the execution path.

Also I don't see a reason not to cache it, because I really think if AM creates any of these mutable classes TypePair, MemberRequest, ect they WILL have to call GetHashCode() because you will be checking/adding to the _typeMapCache and every time you go through it the hashcodes will have to be called, on both sides.

src/AutoMapper/TypePair.cs
TylerCarlson1: If you cache the result it's even faster than re-calculating it :)

src/AutoMapper/TypePair.cs
lbargaoanu: That comment explains why and measuring confirms it. The main reason is reusing the code, but it's also slightly faster and uses less space.

src/AutoMapper/TypePair.cs
lbargaoanu: Actually no, because sometimes you'll compute it when it's not needed. Try the simple benchmark with your changes and compare with my version.

Issue #1907, opened 2017-01-15
AssertConfigurationIsValid should (or at least be able to) bring attention to enum type mismatches


JonathanLydall:
```csharp
public class Tests
{
    [Fact]
    public void TestCase()
    {
        Mapper.Initialize(cfg =>
        {
            cfg.CreateMap<SourceType, DestinationType>();
        });

        // Would ideally throw an exception due to PropertyWithMatchingName having different
        // enum types on the source and destination types.
        Mapper.AssertConfigurationIsValid();

        // Works:
        var result0 = Mapper.Map<DestinationType>(new SourceType
        {
            PropertyWithMatchingName = SourceEnum.Common
        });

        // Results in an invalid enum value being assigned:
        var result1 = Mapper.Map<DestinationType>(new SourceType
        {
            PropertyWithMatchingName = SourceEnum.UniqueToSourceEnum
        });
    }
}

public class SourceType
{
    public SourceEnum PropertyWithMatchingName { get; set; }
}

public class DestinationType
{
    public DestinationEnum PropertyWithMatchingName { get; set; }
}

public enum SourceEnum
{
    Common = 0,
    UniqueToSourceEnum = 2,
}

public enum DestinationEnum
{
    Common = 0,
    UniqueToDestinationEnum = 3,
}
```

Possibly related to #1609.

It would be useful if `Mapper.AssertConfigurationIsValid()` could optionally throw an exception to bring attention to a mapping issue in cases like this.

There are two kinds of checks that could be performed:
1. Do the source and destination types match or otherwise have a mapping configured?
2. Are all enum values in the source type catered for on the destination enum?

Personally I would happy with just check `1`, even if it's an additional check which I have to explicitly opt into.

While I don't personally see a need for check `2`, it seems like issue #1609 wants something like it. What defines as "catered for" in check `2` might be something like checking if the field values and field names match, like the `SourceEnum.UniqueToSourceEnum` field in my example above has nothing corresponding on the `DestinationEnum` enum.

lbargaoanu:
Maybe make validation extensible. AssertConfigurationIsValid could receive a delegate that would be called for each property map.

JonathanLydall:
It also occurred to me that it would be useful to optionally be able to prevent automatic enum value to string conversions.

[...]

Issue #1609, opened 2016-08-10
EnumToEnumMapper should throw exception when enums are not compatible


att741:
When there's no strict 1-1 mapping between enums an exception should be thrown or it should be configurable.

Strict 1-1 mapping:
- Value count match
- Value name match

jbogard:
What should throw an exception?

jbogard:
Fixed by #1919

stevenxi:
Not sure why, but not getting exception when enums are not matching.
Using version 7.0.1
I think this issue should be reopen?

jbogard:
Generally we don't re-open years-old issues. New issue with a repro is best.

PR #1919, opened 2017-01-19
Custom config validation

lbargaoanu:
For things like #1907 and #1609.

### DIFF
--- a/src/AutoMapper/ConfigurationValidator.cs
+++ b/src/AutoMapper/ConfigurationValidator.cs
@@ -11,10 +11,12 @@ namespace AutoMapper
     public class ConfigurationValidator
     {
         private readonly IConfigurationProvider _config;
+        private readonly Action<ValidationContext> _extraValidator;
 
-        public ConfigurationValidator(IConfigurationProvider config)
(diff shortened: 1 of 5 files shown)


src/AutoMapper/ConfigurationValidator.cs
```diff
@@ -11,10 +11,12 @@ namespace AutoMapper
     public class ConfigurationValidator
     {
         private readonly IConfigurationProvider _config;
+        private readonly Action<ValidationContext> _validator;
 
-        public ConfigurationValidator(IConfigurationProvider config)
+        public ConfigurationValidator(IConfigurationProvider config, Action<ValidationContext> validator)
```
jbogard: I like the idea here. But can we do this as part of IConfigurationProvider? I don't want to add more stuff to the constructor on this or the MapperConfiguration object

lbargaoanu: OK, so you want a list of actions? I don't know if that's useful, but for consistency.

lbargaoanu:
Updated.

PR #1923, opened 2017-01-26
Expression improvements w/ Explanation

TylerCarlson1:
From discussion on #1916 to try to re-create the functionality of the PR, sans Include members.

First add unit tests from #1916 and then fix issues one by one.
In fixing the tests and explaining what happens in the Commit description you can get a better idea of what's going on.
Multiple tests pass when fixing a single GH Issue.
One test which involves the include doesn't pass so skip it.

### DIFF
--- a/src/AutoMapper/Mappers/ExpressionMapper.cs
+++ b/src/AutoMapper/Mappers/ExpressionMapper.cs
@@ -297,7 +297,7 @@ namespace AutoMapper.Mappers
                 {
                     if (node.Expression is MemberExpression)
                         return GetConvertedSubMemberCall(node);
-                    return node;
+                    return _parentMappingVisitor != null ? _parentMappingVisitor.Visit(node) : node;
                 }
 
                 var constantVisitor = new IsConstantExpressionVisitor();
@@ -309,7 +309,7 @@ namespace AutoMapper.Mappers
 
                 var replacedExpression = Visit(node.Expression);
                 if (replacedExpression == node.Expression)
-                    replacedExpression = _parentMappingVisitor.Visit(node.Expression);
+                    return _parentMappingVisitor != null ? _parentMappingVisitor.Visit(node) : node;
 
                 if (propertyMap.CustomExpression != null)
                     return propertyMap.CustomExpression.ReplaceParameters(replacedExpression);
--- a/src/AutoMapper/Mappers/ExpressionMapper.cs
+++ b/src/AutoMapper/Mappers/ExpressionMapper.cs
@@ -1,4 +1,5 @@
 ﻿using System.Collections;
+using Expression = System.Linq.Expressions.Expression;
 
 namespace AutoMapper.Mappers
 {
@@ -10,6 +11,7 @@ namespace AutoMapper.Mappers
     using System.Reflection;
     using Configuration;
     using Execution;
+    using static Expression;
 
     public class ExpressionMapper : IObjectMapper
     {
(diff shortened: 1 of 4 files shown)


jbogard:
Do the Include members make sense to keep, or should we try to put those in the EF6 extension?

jbogard:
@BlaiseD does this PR do what you're looking for, or should we look at yours more?

jbogard:
@BlaiseD also, if parts of your PR are very EF-specific, we can also look at putting some pieces in the AutoMapper.EF6 package.

TylerCarlson1:
@jbogard honestly I forgot to look into those in more detail.

My initial guess is they are and shouldn't be added by default.  At least he didn't have them be the default behavior.

I'll look into it more this time :)

TylerCarlson1:
OK I looked at it more and made my assessment of Include members.

From what I see all it does is remove strings, structs, and nullables properties from being converted in the lambda statement.  It just strait up removes them if they are the last properties in a MemberExpression chain and throws an ugly exception. [Here's](https://github.com/AutoMapper/AutoMapper/pull/1916/files#diff-e8ba623ed86da2b4f79378c1fe5b1f22R29) the heart of the change and I have no clue why you would remove parameters from being translated.  De-flatting works fine without Include members.

It's the only reason that one test I think makes no sense even passes.  If you apply this include to any other test they all break.  If you remove this and use the Xpression mapping by itself everything but that test will pass.  (Same as my PR right now)

So my assessment is that Include should be something completely separate, because it can easily break expected functionality.  Also I'm not sure how that can be used, even @BlaiseD has some concerns about the test in the comments.

@BlaiseD also added some new extension methods that take a list of expressions and translate them all, which I think it out of scope for what AM should do.

As far as I see right now, this should do everything @BlaiseD's PR does except Include, which I would argue shouldn't be added to AM in its current state.

BlaiseD:
@jbogard and @TylerCarlson1 . The EF.6 package may be an option - that package depends on EF6 today though. XPressionMapper in the PR does not depend on EF 6 and can be used with either EF6 or EF Core.

Also the MapAsInclude functionality still depends on the rest of the XpressionMapper logic.  Would you want to move all of that to a separate package?  That's really the status quo as in the package [here](https://github.com/BlaiseD/XpressionMapper.NetCore).

The need for mapping as include again is because the navigation property may not exist in the DTO/ViewModel.

Finally I may be biased but Xpression works better.  I mentioned the following differences before as examples of where the current #1923 requires a conversion.

**Map__select_method_projecting_to_anonymous_type**
`{s => s.Account.Things.Select(x => new <>f__AnonymousType0`1(MM = x.Bar)).Where(b => b.MM.EndsWith("3"))} `- PR 1916

`{s => Convert(s.Account.Things.Select(x => new <>f__AnonymousType0`1(MM = x.Bar)).Where(b => b.MM.EndsWith("3")))}` - AutoMapper-Xpression_Test

**Map_orderBy_thenBy_expression**
`{q => q.OrderByDescending(s => s.UserId).ThenBy(s => Concat(s.FirstName, " ", s.LastName))}` - PR 1916

`{q => Convert(q.OrderByDescending(s => s.UserId).ThenBy(s => Concat(s.FirstName, " ", s.LastName)))}` - AutoMapper-Xpression_Test

Ultimately it is your call.  The new extension methods cannot break existing functionality since they do not call existing code. I'm assuming there are no references to XpressionMapper in ExpressionMapper - just call mapper.MapExpression to use XpressionMapper.

Currently with #1923 I'd have to use a separate XpressionMapper library in conjunction with AutoMapper.

Thanks so much.

TylerCarlson1:
>The need for mapping as include again is because the navigation property may not exist in the DTO/ViewModel.

If that's the case I think it should be remade to account for the class's Navigation Properties, and not just all strings, structs, and nullables.  And at that point it's referencing EF to do so and should be in AM.EF6

PR #1947, opened 2017-02-09
Fix to static method TypePair.Create() pulling type from instance of source (always) instead

BlackjacketMack:
of trusint inbound types.  Specifically, if someone wanted to map a source in different forms (e.g. via interfaces)
to a destination they would be unable to do so.

### DIFF
--- a/src/AutoMapper/TypePair.cs
+++ b/src/AutoMapper/TypePair.cs
@@ -57,23 +57,11 @@ namespace AutoMapper
 
         public static TypePair Create<TSource>(TSource source, Type sourceType, Type destinationType)
         {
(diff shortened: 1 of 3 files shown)


BlackjacketMack:
Several tests were failing on master prior to branching this pull-request fyi.

BlackjacketMack:
I'll go in and take a look at those tests failing...locally the test suite was taking a huge amount of time (+30min.) to run the OATH test library.

BlackjacketMack:
Hmmm...It breaks too many things.  I'll drop it as a pull request and list it as an issue.  I'm still curious why it doesn't behave as expected.

PR #1948, opened 2017-02-10
We usually want ResolveTypeMap, not TypeMapRegistry.GetTypeMap

lbargaoanu:
I think we fixed a few issues around TypeMapRegistry.GetTypeMap and now it's not used much.


jbogard:
Looks good, not sure why we had those on TypeMapRegistry to begin with.

Issue #1956, opened 2017-02-15
ProjectTo<>(...,Expession<<>>[]membersToExpand): lambda for a Value Type causes NullReferenceException


moudrick:
ProjectTo method, 
membersToExpand parameter, 
Expession[] signature, 

lambda for a Value Type causes NullReferenceException

Details are at SO http://stackoverflow.com/questions/41945809/automapper-5-2-ignores-explicitexpansion-if-it-is-configured-in-base-dto-mapping

See section UPD2 of the Question, code snippet, 
`[TestMethod] public void ProjectAllWithMemberExpression_Exception() => _iq.ProjectTo<DtoDerived>(_ => _.Name, _ => _.Desc);`

lbargaoanu:
What's the scenario here? It seems strange to use ExplicitExpansion with a value type.

moudrick:
I agree it is strange for non-nullable types. In case of nullable, one may need skipping its expansion. As a matter of fact, int? and any nullable type is actually a value type, despite the fact that it can be null as reference type do.

lbargaoanu:
ExplicitExpansion is used to avoid fetching a child object, but is it worth it for a simple value? :)

moudrick:
Yes, if you want your projection to avoid generating additional joins on flattening db model. E.g. if you MapFrom a field like `_ => _.InnerObject.AnotherInnerObject.ScalarField`

lbargaoanu:
Aha, that makes sense.

moudrick:
Well, I updated the PR with more examples that illustrate optional mapping from int to int? with deep inner member flattening.

lbargaoanu:
Another way to solve this would be to make the expression type-safe, but that is probably overkill.

PR #1957, opened 2017-02-15
5.2.0 issue value type projection expression

moudrick:
#1956

### DIFF
--- a/src/AutoMapper/QueryableExtensions/ProjectionExpression.cs
+++ b/src/AutoMapper/QueryableExtensions/ProjectionExpression.cs
@@ -102,21 +102,10 @@ namespace AutoMapper.QueryableExtensions
 
         internal class MemberVisitor : ExpressionVisitor
         {
-            protected override Expression VisitLambda<T>(Expression<T> node)
+            protected override Expression VisitMember(MemberExpression node)
             {
-                MemberExpression memberExpression = 
-                    node.Body.NodeType == ExpressionType.Convert 
-                        ? (node.Body as UnaryExpression)?.Operand  as MemberExpression
-                        :  node.Body as MemberExpression;
-                if(memberExpression != null)
-                {
-                    if(MemberPath != null)
-                    {
-                        throw new InvalidOperationException("There are more than one lambda member expressions.");
-                    }
-                    MemberPath = GetMemberPath(memberExpression);
-                }
-                return base.VisitLambda<T>(node);
+                MemberPath = GetMemberPath(node);
+                return node;
             }
 
             private IEnumerable<MemberInfo> GetMemberPath(MemberExpression memberExpression)
--- a/src/AutoMapper/QueryableExtensions/ProjectionExpression.cs
+++ b/src/AutoMapper/QueryableExtensions/ProjectionExpression.cs
@@ -87,7 +87,8 @@ namespace AutoMapper.QueryableExtensions
 
         internal IQueryable<TResult> To<TResult>(IObjectDictionary parameters, MemberPaths memberPathsToExpand)
         {
-            var membersToExpand = memberPathsToExpand.SelectMany(m => m).Distinct().ToArray();
+            IEnumerable<MemberInfo> selectMany = memberPathsToExpand.SelectMany(m => (m ?? new List<MemberInfo>()) as IList<MemberInfo> ?? m /*?.ToList() */);
+            MemberInfo[] membersToExpand = selectMany.Distinct().ToArray();
 
             var mapExpression = _builder.CreateMapExpression(_source.ElementType, typeof(TResult), parameters, membersToExpand);
 
@@ -104,7 +105,10 @@ namespace AutoMapper.QueryableExtensions
         {
             protected override Expression VisitLambda<T>(Expression<T> node)
             {
-                var memberExpression = node.Body as MemberExpression;
+                MemberExpression memberExpression = 
+                    node.Body.NodeType == ExpressionType.Convert 
+                        ? (node.Body as UnaryExpression)?.Operand  as MemberExpression
+                        :  node.Body as MemberExpression;
                 if(memberExpression != null)
                 {
                     if(MemberPath != null)
--- a/src/AutoMapper/QueryableExtensions/ProjectionExpression.cs
+++ b/src/AutoMapper/QueryableExtensions/ProjectionExpression.cs
@@ -87,8 +87,7 @@ namespace AutoMapper.QueryableExtensions
 
         internal IQueryable<TResult> To<TResult>(IObjectDictionary parameters, MemberPaths memberPathsToExpand)
         {
-            IEnumerable<MemberInfo> selectMany = memberPathsToExpand.SelectMany(m => (m ?? new List<MemberInfo>()) as IList<MemberInfo> ?? m /*?.ToList() */);
-            MemberInfo[] membersToExpand = selectMany.Distinct().ToArray();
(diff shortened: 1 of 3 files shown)


src/AutoMapper/QueryableExtensions/ProjectionExpression.cs
```diff
@@ -87,7 +87,8 @@ public IQueryable<TResult> To<TResult>(IObjectDictionary parameters, params Expr
 
         internal IQueryable<TResult> To<TResult>(IObjectDictionary parameters, MemberPaths memberPathsToExpand)
         {
-            var membersToExpand = memberPathsToExpand.SelectMany(m => m).Distinct().ToArray();
+            IEnumerable<MemberInfo> selectMany = memberPathsToExpand.SelectMany(m => (m ?? new List<MemberInfo>()) as IList<MemberInfo> ?? m /*?.ToList() */);
+            MemberInfo[] membersToExpand = selectMany.Distinct().ToArray();
```
lbargaoanu: m should never be null. So the right behavior here is to crash, not check for null. That only makes it more difficult to debug.

src/AutoMapper/QueryableExtensions/ProjectionExpression.cs
```diff
@@ -104,7 +105,10 @@ internal class MemberVisitor : ExpressionVisitor
         {
             protected override Expression VisitLambda<T>(Expression<T> node)
             {
-                var memberExpression = node.Body as MemberExpression;
+                MemberExpression memberExpression = 
+                    node.Body.NodeType == ExpressionType.Convert 
+                        ? (node.Body as UnaryExpression)?.Operand  as MemberExpression
+                        :  node.Body as MemberExpression;
```
lbargaoanu: That's replicated code, you need the cast to MemberExpression only once. Also, I'm not sure how we would know any convert expression is about this case. I think we would need more here.

src/UnitTests/Projection/MembersToExpand_ExpressionArray.cs
```diff
@@ -0,0 +1,40 @@
...
+using AutoMapper.QueryableExtensions;
+using Should;
+using Xunit;
+
+namespace AutoMapper.UnitTests.Projection
+{
+    using NameSourceType = String        ; using NameDtoType = String       ; // Example of Reference Type
+    using DescSourceType = Nullable<int> ; using DescDtoType = Nullable<int>; // Example of Value Type
+
+    public class MembersToExpand_ExpressionArray  : AutoMapperSpecBase
```
lbargaoanu: Tests in memory can be misleading. We need a test in IntegrationTests.Net4.

src/AutoMapper/QueryableExtensions/ProjectionExpression.cs
lbargaoanu: I guess more user-friendly would be to throw a nice error, but that wouldn't help much, it's still our bug.

src/AutoMapper/QueryableExtensions/ProjectionExpression.cs
lbargaoanu: The cast should always be from a value type to object I believe.

src/AutoMapper/QueryableExtensions/ProjectionExpression.cs
moudrick: Agree. Actually this was my getting-throw hacking piece that I forgot to remove after debug. My fault, I'll reverse this hunk in next update to this PR.

src/AutoMapper/QueryableExtensions/ProjectionExpression.cs
moudrick: No. If to remove this fix, you'll get NullReferenceException on passing lambda of accessing a member of value type, after the line `var memberExpression = node.Body as MemberExpression;`. as operator returns null due to top level body is other than `MemberExpression`, namely, Convert operation. You can make sure that it is true by reversing this hunk and running/debugging the tests from this PR (kind of TDD approach professing). This is a bug of crash level that forces using ProjectTo signature with **string[]** notation of **membersToExpand** that makes compile-time control on field names more difficult and less reliable.

This is because value types need **boxing** to be assigned to **object** target. In ExpressionTree, it is expressed in wrapping it to Expression.Convert which is actually UnaryExpression with one operand that actually is MemberExpression with return type of value type. If we not convert, it is not assignable to `Expression<Func<Source, object>>`.

`Expression<Func<Dto, int?  >>` is not assignable to `Expression<Func<Dto, object>>`. You cannot pass both . In case of code in tests, the compiler infers them all as E`xpression<Func<Dto, object>>` and adds Expression.Convert where necessary. The code below demonstrates it.

```
Expression<Func<Dto, int?  >> nullableMemberExpression1 = _ => _.Desc; Console.WriteLine(nullableMemberExpression1); // _ => _.Desc
Expression<Func<Dto, object>> nullableMemberExpression2 = _ => _.Desc; Console.WriteLine(nullableMemberExpression2); // _ => Convert(_.Desc)
Expression<Func<Dto, object>> referenceMemberExpression = _ => _.Name; Console.WriteLine(referenceMemberExpression); // _ => _.Name

// Expression<Func<Dto, object>> projectExpression = nullableMemberExpression1;
// error CS0029: Cannot implicitly convert type 'System.Linq.Expressions.Expression<System.Func<AutoMapper.UnitTests.Projection.MembersToExpand_ExpressionArray.Dto, int?>>' to 'System.Linq.Expressions.Expression<System.Func<AutoMapper.UnitTests.Projection.MembersToExpand_ExpressionArray.Dto, object>>'

// _iq.ProjectTo<Dto>(Configuration, referenceMemberExpression, nullableMemberExpression1);
// error CS1503: Argument 4: cannot convert from 'System.Linq.Expressions.Expression<System.Func<AutoMapper.UnitTests.Projection.MembersToExpand_ExpressionArray.Dto, int?>>' to 'System.Linq.Expressions.Expression<System.Func<AutoMapper.UnitTests.Projection.MembersToExpand_ExpressionArray.Dto, object>>'

_iq.ProjectTo<Dto>(Configuration, nullableMemberExpression1).First().Desc.ShouldEqual(_iqf.Desc);
_iq.ProjectTo<Dto>(Configuration, referenceMemberExpression, nullableMemberExpression2);

```

src/UnitTests/Projection/MembersToExpand_ExpressionArray.cs
moudrick: Ok, I'll do. I just need some time to find out a better way to introduce them.

src/AutoMapper/QueryableExtensions/ProjectionExpression.cs
lbargaoanu: That's not what I meant, but I guess it doesn't matter. There is a simpler approach. 
The test should look like [this](https://github.com/AutoMapper/AutoMapper/blob/master/src/IntegrationTests.Net4/ByteArray.cs). An in memory test is not relevant because EF supports only certain expressions and that's what matters.

jbogard:
Fixes #1956

src/AutoMapper/QueryableExtensions/ProjectionExpression.cs
moudrick: @lbargaoanu, 
thanks for more universal fix for this issue, and for merging it in!
 just created a new PR #1965 with promised integration tests, sorry for delay, it took time to figure out how to execute them first time.

Issue #1958, opened 2017-02-15
Automapper 5.2 ignores ExplicitExpansion if it is configured in Base DTO mapping


moudrick:
comprehensive details and discussion in Q&A@SO: Automapper 5.2 ignores ExplicitExpansion if it is configured in Base DTO mapping

lbargaoanu:
Include doesn't really work with EF, so why bother?

moudrick:
Is it planned to make Include working with EF? If it is not a rocket science and if any hint where to do this, I could contribute it.

lbargaoanu:
No. Search the older issues, it has been discussed a few times.

jbogard:
EF can't work AFAIK. Unless you can produce a Select projection without AutoMapper against EF, we can't help.

PR #1959, opened 2017-02-15
5.2.0 issue explicit expansion inheritance

moudrick:
#1958

### DIFF
--- a/src/AutoMapper/PropertyMap.cs
+++ b/src/AutoMapper/PropertyMap.cs
@@ -41,7 +41,7 @@ namespace AutoMapper
         public LambdaExpression CustomExpression { get; private set; }
         public MemberInfo CustomSourceMember { get; set; }
         public bool UseDestinationValue { get; set; }
-        public bool ExplicitExpansion { get; set; }
+        public bool ExplicitExpansion { get { return _explicitExpansion.GetValueOrDefault(); } set { _explicitExpansion = value; } } private bool? _explicitExpansion; 
         public object NullSubstitute { get; set; }
         public ValueResolverConfiguration ValueResolverConfig { get; set; }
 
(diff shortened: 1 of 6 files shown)


src/AutoMapper/PropertyMap.cs
```diff
@@ -41,7 +41,7 @@ public PropertyMap(PropertyMap inheritedMappedProperty, TypeMap typeMap)
         public LambdaExpression CustomExpression { get; private set; }
         public MemberInfo CustomSourceMember { get; set; }
         public bool UseDestinationValue { get; set; }
-        public bool ExplicitExpansion { get; set; }
+        public bool ExplicitExpansion { get { return _explicitExpansion.GetValueOrDefault(); } set { _explicitExpansion = value; } } private bool? _explicitExpansion;
```
jbogard: Should ExplicitExpansion just be nullable?

jbogard: I say that because I converted other inheritable members to nullable

lbargaoanu: But if Include doesn't really work with EF, how is this useful?

jbogard: Just purely for inheriting configuration. Include works two ways - applying base config to children, and at runtime mapping doing the polymorphic stuff. I can see the configuration inheritance being valuable.

moudrick: Anyway, it is confusing pretty much with or without working with EF. 

If inheritance can be done within some other scope, my issue could be added to that scope as a related issue. If this is going to be fixed in that scope and the code in my PR is not much helpful, the PR can be just declined, or I'll revoke it, if GitHub allows something like this, If the issue is found irrelevant, we can just close it, leaving this confuse intact.

Thanks, Oleksii.

jbogard: Well I really just wanted my original question answered - should ExplicitExpansion just be nullable, and remove that backing field.

moudrick: @jbogard, I think that ExplicitExpansion and other bool properties listed in the link below need to be non-nullable booeans, but backing fields have to be nullable for including base configuration sake.

The list of bool mapping config properties is @ SO answer by Ivan Stoev: http://stackoverflow.com/questions/41945809/automapper-5-2-ignores-explicitexpansion-if-it-is-configured-in-base-dto-mapping/42012290#42012290

The coding offer to introduce nullable backing field in below in the Ivan Stoev's comment to this answer: http://stackoverflow.com/questions/41945809/automapper-5-2-ignores-explicitexpansion-if-it-is-configured-in-base-dto-mapping/42012290#comment71390369_42012290

jbogard: For consistency's sake, I'd make it a nullable property like all the others.

PR #1972, opened 2017-02-21
[Experiment] Mappers per profile

lbargaoanu:
The simplest thing that makes sense. At least to me :)
Remove needs more work. Whenever we create a profile we would have to pass the global config. If the user creates it, we have to revert to the default mappers.


lbargaoanu:
With remove, the tests look like [this](https://github.com/lbargaoanu/AutoMapper/blob/ProfileCtor/src/UnitTests/Profiles.cs#L13). I don't see a better way if a profile does things in the constructor.

Issue #1973, opened 2017-02-22
Collection is read-only exception when try to map to ICollection with fixed size


Tasteful:
Try to set values on auto-initialized ICollection throws the following exception

Is this a correct behavior to not been able to update the properties that is ICollection but have a fixed size?

```
Error mapping types.

Mapping types:
B -> A
CollectionMappingTest+B -> CollectionMappingTest+A

Type Map configuration:
B -> A
CollectionMappingTest+B -> CollectionMappingTest+A

Property:
Items
Collection is read-only.
```

Following test code is used
```
using System;
using System.Collections.Generic;
using System.Linq;
using AutoMapper;
using Xunit;

    public class CollectionMappingTest
    {
        [Fact]
        public void ReadOnlyCollectionTest()
        {
            Mapper.Initialize(cfg =>
            {
                cfg.CreateMap<A, B>()
                   .ReverseMap();
                cfg.CreateMap<AA, BB>()
                   .ReverseMap();
            });

            var itemA = new A
            {
                Items = new []{ new AA { Value = "1" }, new AA { Value = "2" }, new AA { Value = "3" } }
            };
            Assert.Equal(3, itemA.Items.Count);
            Assert.True(itemA.Items is Array);

            var itemB = Mapper.Instance.Map<B>(itemA);
            Assert.Equal(3, itemB.Items.Count);

            Assert.False(itemB.Items is Array);
            itemB.Items.Remove(itemB.Items.Last());

            Assert.Equal(2, itemB.Items.Count);

            Mapper.Instance.Map(itemB, itemA);
            Assert.Equal(2, itemA.Items.Count);
        }
        public class A
        {
            public ICollection<AA> Items { get; set; }
        }
        public class B
        {
            public ICollection<BB> Items { get; set; }
        }
        public class AA
        {
            public string Value { get; set; }
        }
        public class BB
        {
            public string Value { get; set; }
        }
    }
```
AutoMapper version: 5.2.0

-----

In AutoMapper 4 I was using an own https://github.com/AutoMapper/AutoMapper/blob/v4.2.1/src/AutoMapper/Mappers/CollectionMapper.cs that make an override of the inner `EnumerableMapper<TCollection, TElement>` implementation with the following snippet to solve this.
```
            protected override object GetOrCreateDestinationObject(ResolutionContext context, Type destElementType, int sourceLength)
            {
                // If the source is an array, assume we can add to it...
                if (context.DestinationValue is Array)
[...]

PR #1974, opened 2017-02-22
Collection is read-only exception when try to map to ICollection with fixed size

Tasteful:
Unit tests and fix in the expression builder to do additional check if the destination collection is an read only array.

#1973

### DIFF
--- a/src/AutoMapper/Mappers/CollectionMapper.cs
+++ b/src/AutoMapper/Mappers/CollectionMapper.cs
@@ -41,12 +41,14 @@ namespace AutoMapper.Mappers
                                        configurationProvider.Configuration.AllowNullCollections;
             var ifNullExpr = allowNullCollections ? Constant(null, passedDestination.Type) : (Expression) newExpression;
             var clearMethod = destinationCollectionType.GetDeclaredMethod("Clear");
-            var checkNull =  
+            var createNewAndAssignExpression = Assign(newExpression, passedDestination.Type.NewExpr(ifInterfaceType));
+
+            var checkNull =
                 Block(new[] { newExpression, passedDestination },
                     Assign(passedDestination, destExpression),
                     IfThenElse(condition ?? Constant(false),
-                                    Block(Assign(newExpression, passedDestination), Call(newExpression, clearMethod)),
-                                    Assign(newExpression, passedDestination.Type.NewExpr(ifInterfaceType))),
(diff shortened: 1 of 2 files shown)


lbargaoanu:
In my view, this is a hack. Allow passing an invalid destination for this mapper and then ignoring it. Something explicit like DoNotUseDestinationValue makes more sense to me.

Tasteful:
The defined type is `ICollection<AA>` but is created of an instance of `AA[]`. The Array `AA[]` implements `ICollection<AA>` interface. So both the declaration with `ICollection<AA>` and assigning with `AA[]` is valid of the coding perspective.

`CollectionMapper` is using the `Add` method on the array and regarding to documentation should throw exception, see https://msdn.microsoft.com/en-us/library/bb342410%28v=vs.110%29?f=255&MSPPError=-2147217396 and this known situation is not handled in the `ICollectionMapper`.

lbargaoanu:
Yes, because the array mapper handles this case. Maybe @jbogard will see it your way.

Tasteful:
How can I setup in the configuration that the `ArrayMapper` should be used instead of the `ICollectionMapper` if the destination definition type is `ICollection` but have a run time value type of `Array`?

Tasteful:
Hack to do this check outside without modification in `CollectionMapper` but I think that will take extra performance for the operation.
```
                var propertiesTypeCache = new ConcurrentDictionary<Type, Lazy<ICollection<PropertyInfo>>>();
                var typeActivator = new ConcurrentDictionary<Type, Lazy<Func<IList>>>();
                cfg.ForAllMaps((typeMap, mapping) => mapping.BeforeMap((src, dest) =>
                {
                    if (dest != null)
                    {
                        foreach (var prop in propertiesTypeCache.GetOrAdd(dest.GetType(), t => new Lazy<ICollection<PropertyInfo>>(()=> t.GetProperties().Where(x => x.PropertyType.IsGenericType && x.PropertyType.GetGenericTypeDefinition() == typeof(ICollection<>)).ToList())).Value)
                        {
                            var items = prop.GetValue(dest) as Array;
                            if (items != null)
                            {
                                var collection = typeActivator.GetOrAdd(prop.PropertyType, t => new Lazy<Func<IList>>(() =>
                                {
                                    var a = typeof(List<>).MakeGenericType(t.GetGenericArguments());
                                    return () => (IList)Activator.CreateInstance(a);
                                })).Value();

                                foreach (var item in items)
                                {
                                    collection.Add(item);
                                }

                                prop.SetValue(dest, collection);
                            }
                        }
                    }
                }));
```

lbargaoanu:
DoNotInline would force using the runtime types, but that crashes in the array mapper. That could be easily fixed. But does it make sense? :)

PR #1976, opened 2017-02-23
Allow arrays to be typed as ICollection with DontInline

lbargaoanu:
Technically this is a fix for #1973, but I'm not sure it would make sense to the regular user. It's a bug regardless.


jbogard:
Hmmmm, should we have a better name for that method? It's really talking about the implementation vs. what it's doing. Maybe something with "Runtime" in the name, since it's switching from config-time type checking to runtime type checking?

lbargaoanu:
Yes, I was thinking MapAtRuntime, but I'm not sure if it fits the original use case.

jbogard:
Ah right the original use case was "the expression tree is large enough to crash the process".

lbargaoanu:
Rebased.

src/AutoMapper/Configuration/MemberConfigurationExpression.cs
```diff
@@ -20,7 +20,7 @@ public MemberConfigurationExpression(MemberInfo destinationMember, Type sourceTy
 
         public MemberInfo DestinationMember => _destinationMember;
 
-        public void DoNotInline()
+        public void MapAtRuntime()
         {
             PropertyMapActions.Add(pm => pm.Inline = false);
```
jbogard: Should we rename this property too?

lbargaoanu: I'd say no. I think for us inline makes more sense. MapAtRuntime is a little vague :)

PR #1984, opened 2017-03-02
Break out object mappers

jbogard:
We had a few mappers in the same file, these are broken out. Also, the enum one was combining enum -> primitive and primitive -> enum, so I split that out into two simpler ones.


src/AutoMapper/Mappers/MultidimensionalArrayMapper.cs
```diff
@@ -13,7 +13,7 @@ public class MultidimensionalArrayMapper : IObjectMapper
     {
         static MultidimensionalArrayFiller filler;
```
lbargaoanu: This is certainly broken. But I guess nobody is using it.

Issue #1985, opened 2017-03-02
Get rid of ObjectCreator


jbogard:
It doesn't really have a reason to live, most things should either go to the delegate factory to create an expression or just inline the logic.

lbargaoanu:
I think it could be useful if we want to handle null in a generic way.

PR #1990, opened 2017-03-06
Consolidate ctors

jbogard:


src/AutoMapper/MapperConfiguration.cs
```diff
@@ -201,40 +193,41 @@ public TypeMap ResolveTypeMap(Type sourceType, Type destinationType)
...
-                {
-                    typeMap.Seal(this);
-                }
-            }
+            var typeMap = _typeMapPlanCache.GetOrAdd(typePair);
+
+            if (typeMap == null)
+                return null;
+
+            lock (typeMap)
```
lbargaoanu: In most cases there is no need for a lock here. That's why those checks were in place. This is wasteful.

jbogard:
Fixes #1985

src/AutoMapper/MapperConfiguration.cs
jbogard: Which part is wasteful, the lock or the null check?

lbargaoanu: The lock, It should be taken only when needed.

lbargaoanu: Using Lazy for the map expression is an alternative, but that's needed only when creating maps on the fly so I don't think it's worth it.

jbogard: Oh that lock was already there, it just started blowing up because the type map created could be null.

lbargaoanu: Yes, but the lock was taken only in certain conditions. You deleted those checks. That's my point :)

PR #1991, opened 2017-03-07
We need the lock

lbargaoanu:
This is what I meant. But a test is the best :)


jbogard:
Ah, that helps a lot! I had removed it and everything passed.

lbargaoanu:
The complexity comes from creating maps on the fly.

PR #2023, opened 2017-03-23
Introducing Internal namespaces and classes to support extensibility

### LINKED ISSUE #2015: Building custom IObjectMappers based on existing mappers requires a lot of internal class duplication
Some general feedback on upgrading some custom IObjectMappers to the latest signatures. I've had to take copies of:-

1. CollectionMapperExtensions
2. ExpressionExtensions
3. TypeHelper
4. TypeExtensions

As these are all internal classes, the time to implement a custom object mapper which is closely based on an out of the box mapper is much longer. Is there a way this functionality can be exposed without causing method name ambiguity due to duplicate extension method names?

SaltyDH:
Fixes #2015.


lbargaoanu:
@jbogard I think we should merge this. It touches a lot of files.

PR #2065, opened 2017-04-10
Use the existing destination

### LINKED ISSUE #2064: Mapping to readonly IEnumerable property
When trying to map to a readonly IEnumerable property backed by an instance of an IList field AutoMapper throws a NullReferenceException.

### Source/destination types

```csharp
public class Source
{
    private readonly List<string> _myCollection = new List<string>
    {
        "one", "two"
    };

    public string[] MyCollection => _myCollection.ToArray();
}

public class Destination
{
    private IList<string> _myCollection = new List<string>();
    public IEnumerable<string> MyCollection => _myCollection;
}
```

### Mapping configuration

```csharp
CreateMap<Source, Destination>()
                .ForMember(m => m.MyCollection, opt =>
                {
                    opt.MapFrom(src => src.MyCollection);
                    opt.UseDestinationValue();
                });
```

### Expected behavior
The `UseDestinationValue` option should add items to the IList<> backing field. Changing the property to a `IList<string>` works as expected.

### Actual behavior
AutoMapper throws a `NullReferenceException`.

### Steps to reproduce

```csharp
var result = Mapper.Map<Destination>(new Source());
```

lbargaoanu:
Fixes #2064.

### DIFF
--- a/src/AutoMapper/Mappers/EnumerableMapper.cs
+++ b/src/AutoMapper/Mappers/EnumerableMapper.cs
@@ -20,13 +20,13 @@ namespace AutoMapper.Mappers
         {
             if(destExpression.Type.IsInterface())
             {
-                var listType = typeof(List<>).MakeGenericType(ElementTypeHelper.GetElementType(destExpression.Type));
-                destExpression = Default(listType);
+                var listType = typeof(IList<>).MakeGenericType(ElementTypeHelper.GetElementType(destExpression.Type));
+                destExpression = Convert(destExpression, listType);
             }
             return MapCollectionExpression(configurationProvider, profileMap, propertyMap, sourceExpression,
                 destExpression, contextExpression, IfEditableList, typeof(List<>), MapItemExpr);
         }
 
-        private static Expression IfEditableList(Expression dest) => And(TypeIs(dest, typeof(IList)), Not(TypeIs(dest, typeof(Array))));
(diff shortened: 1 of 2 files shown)


src/AutoMapper/Mappers/EnumerableMapper.cs
```diff
@@ -20,13 +20,13 @@ public Expression MapExpression(IConfigurationProvider configurationProvider, Pr
...
-                destExpression = Default(listType);
+                var listType = typeof(IList<>).MakeGenericType(ElementTypeHelper.GetElementType(destExpression.Type));
+                destExpression = Convert(destExpression, listType);
             }
             return MapCollectionExpression(configurationProvider, profileMap, propertyMap, sourceExpression,
                 destExpression, contextExpression, IfEditableList, typeof(List<>), MapItemExpr);
         }
 
-        private static Expression IfEditableList(Expression dest) => And(TypeIs(dest, typeof(IList)), Not(TypeIs(dest, typeof(Array))));
+        private static Expression IfEditableList(Expression dest) => AndAlso(TypeIs(dest, typeof(IList)), Not(TypeIs(dest, typeof(Array))));
```
jbogard: Sigh. AndAlso vs And and Or vs OrElse

lbargaoanu: I don't know what they were thinking.

PR #2074, opened 2017-04-13
Unflatten

### LINKED ISSUE #2068: Support un-flattening
Today, "ForMember" only supports top-level properties. This creates the PropertyMap and all its ilk.

I'm proposing un-flattening, by allowing mapping to a path:

```
CreateMap<Foo, Bar>()
    .ForPath(dest => dest.Foo.Bar.Baz, opt => opt.MapFrom(src => src.Boo.Buzz.Bizz));
```

`ForPath` would be new, it would allow any chain of member access expressions. For simplicity, we would not support anything but a chain of member access. No collections, array access, methods etc. Just property/field chains.

The "opt" part would be a new expression interface. We would add new supporting DSL interfaces, configuration gatherers, and Map config objects (PathMap to go along with PropertyMap etc.)

We can limit the creation of these PathMaps to:

- When you call ForPath explicitly
- When you use MemberList.Source in CreateMap (or ReverseMap).

To support full un-flattening (as our users would see it as done):

- [x] Explicit configuration of paths (ForPath) #2074 
- [x] ReverseMap creates PathMaps based on PropertyMaps #2080 
- [x] PreserveReferences on ReverseMap for when we detect maps used in child properties #2124 
~~- [ ] UseDestinationValue (needs more investigation)~~ defer this one

lbargaoanu:
A first cut, but feedback is welcome.


TylerCarlson1:
Stupid question.  Is this only used when mapping two types together and will not infer onto sub type maps?
So if you map something like `A.B.C -> AVM.BVM.CVM` the user won't infer that `B.C -> BVM.CVM` if you were to map B to BVM.  This kind of mapping only applies with these types and these types explicitly?

I know either way you do it it will be confusing to some people I'm just curious if the path will ever be meant to be extended to other type maps.  I know that's much harder to do.

Also curious if you could merge this logic into ForMember and get all the same MapFrom functionality, but I need to look at the code a lot more to see.

lbargaoanu:
I'll just say it's like ForMember but with a path :)

lbargaoanu:
Ready for review.

jbogard:
@TylerCarlson1 I wasnt thinking to merge it to ForMember, as my assumption is we'd treat paths differently than just 1 property to another. We can also constrain it more so that we can appropriately do `CreateMap().ReverseMap()`

TylerCarlson1:
@jbogard Ok that's fine I wanna make sure, cause people can make some weird assumptions, so we can get all our facts in order on what it does and what it doesn't.

src/AutoMapper/Execution/TypeMapPlanBuilder.cs
```diff
@@ -161,8 +165,37 @@ private Expression CreateAssignmentFunc(Expression destinationFunc, bool constru
...
-
+        }
+
+        private Expression HandlePath(PathMap pathMap)
+        {
+            var destination = ((MemberExpression)pathMap.DestinationExpression.ConvertReplaceParameters(_destination)).Expression;
+            var createInnerObjects = CreateInnerObjects(pathMap, destination);
+            var setFinalValue = CreatePropertyMapFunc(new PropertyMap(pathMap.DestinationMember, pathMap.TypeMap)
+            {
+                CustomExpression = pathMap.SourceExpression
```
jbogard: Can we handle this differently? The access modifier change is a bit odd. Maybe an interface or base class here?

src/AutoMapper/PropertyMap.cs
```diff
@@ -38,7 +39,7 @@ public PropertyMap(PropertyMap inheritedMappedProperty, TypeMap typeMap)
         public LambdaExpression CustomResolver { get; set; }
         public LambdaExpression Condition { get; set; }
         public LambdaExpression PreCondition { get; set; }
-        public LambdaExpression CustomExpression { get; private set; }
+        public LambdaExpression CustomExpression { get; internal set; }
```
jbogard: Let's see if we can make this a private property and maybe an interface/base class so that the plan builder can use both a PropertyMap and a PathMap.

lbargaoanu:
Yes, but to do that we need to either split from TryPropertyMap the common part or to dummy implement the whole PM interface. It doesn't seem better. I guess in time we'll need other things from PM.
________________________________________

jbogard:
@lbargaoanu so at the moment it's just using a property map but with most things missing, right?

lbargaoanu:
Yes. The setter thing we can solve with a new constructor.

jbogard:
Let's do that for now then - an overloaded ctor that takes a PathMap.

lbargaoanu:
The diff is messed up for TypeMapPlanBuilder. Sorry about that, but, as usual, I blame GitHub :)

PR #2075, opened 2017-04-14
Support for EnumMemberAttribute

### LINKED ISSUE #1741: Automapper not respecting EnumMemberAttribute value when mapping
It seems that automapper only looks at the defined enum value but ignores [EnumMember(Value = "..")] attribute. 

```
class Test 
{
}

class TestDto
{
    public TestEnum Value { get; set; }
}

public enum TestEnum 
{
    [EnumMember(Value = "some-value")]
    SomeValue
}

Mapper.CreateMap<Test, TestDto>();

var test = new Test { Value = "some-value };
// fails: ArgumentException: Requested value 'some-value' was not found.
var testDto = Mapper.Map<TestDto>(test);

var test = new Test { Value = "SomeValue };
// works
var testDto = Mapper.Map<TestDto>(test);
```

When converting the testDto to JSON it all works fine and displays the `EnumMember`'s value. 

I'm not huge fan of this, but I'm working on an ASP.NET Core application which uses an legacy database and the values there are stored with dashes, so it'd be nice if AutoMapper respect EnumMember too. Using strings here isn't an option.

Tornhoof:
As suggested by  @lbargaoanu in #1741 this is a Pull Request to add EnumMemberAttribute support.

- Changed existing StringToEnumMapper.cs to support a switch table with enum member attribute values, iff EnumMemberAttribute exists for the parsing of strings to enums, it is done case-insensitive (using UpperCaseInvariant
- Added a EnumToStringMapper.cs to do the same the other way round
- Added References to System.Runtime.Serialization.Primitives
- Added Tests for both sides

Fixes #1741.

### DIFF
--- a/src/AutoMapper/AutoMapper.csproj
+++ b/src/AutoMapper/AutoMapper.csproj
@@ -39,6 +39,7 @@
         <PackageReference Include="System.Resources.ResourceManager" Version="4.3.0" />
         <PackageReference Include="System.ComponentModel.TypeConverter" Version="4.3.0" />
 		<PackageReference Include="System.Collections.Specialized" Version="4.3.0" />
+		<PackageReference Include="System.Runtime.Serialization.Primitives" Version="4.3.0" />		
     </ItemGroup>
 
     <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard1.1' ">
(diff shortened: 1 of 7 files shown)


lbargaoanu:
Looks good, but I have one quibble. I don't think ToUpper is the best way to compare strings ignoring case. Using StringComparison.OrdinalIgnoreCase seems better. I guess Unicode stuff doesn't apply here, but there is also the allocation. Anyway, you can wait for @jbogard's take if you prefer. Maybe he doesn't care :)

Tornhoof:
Changed comparison to StringComparer.OrdinalIgnoreCase as per comment by @lbargaoanu

lbargaoanu:
@jbogard I think this is good to go.

jbogard:
My blocker atm is the new framework reference. For full .NET (`net45`) references, we really would rather have a framework reference instead of the NuGet package reference.

However, `dotnet pack` has an open issue around this https://github.com/NuGet/Home/issues/4853

jbogard:
FYI we're still blocked on that NuGet issue.

Tornhoof:
Looks like this won't change until july

jbogard:
@Tornhoof yeah, that's unfortunate. I could merge it in, but it'd be a ton of work to make that dependency show up correctly :(

PR #2080, opened 2017-04-20
Reverse MapFrom

### LINKED ISSUE #2067: ReverseMap() - Working with complex types
Like mentioned in #359 I think the default behavior of ReverseMap() is not really obvious. Right now it only works bidirectional for mappings that are implicit like so:

```C#
class A {
 string Member { get; set; }
}

class B {
 string Member { get; set; }
}

cfg.CreateMap<A, B>() .ReverseMap();
```

In this scenario where both classes have members that are named equally ReverseMapping works perfect. But if classes contain members that are different and not mapped via the default (per naming) mapping this doesn´t work properly.

```C#
class A {
 string Member { get; set; }
 string OtherMember { get; set; }
}

class B {
 string Member { get; set; }
 string NonMember { get; set; }
}

cfg.CreateMap<A, B>() .ReverseMap();
```
The way to go to make this work in one direction is to ad the Mapping via `ForMember`.

```C#
cfg.CreateMap<A, B>()
    .ForMember(x => x.NonMember, m => m.MapFrom(a => a.OtherMember))
    .ReverseMap();
```

This now works properly in one direction, but adding `ReverseMap()` doesn´t make it work in both directions so you have to make the mapping similar to this:

```C#
cfg.CreateMap<A, B>()
    .ForMember(x => x.NonMember, m => m.MapFrom(a => a.OtherMember))
    .ReverseMap()
            .ForMember(x => x.OtherMember, m => m.MapFrom(a => a.NonMember));
```
So basically this might lead to some issues. First of all this is code duplication and will duplication will always be error prone. Next I think the default behavior of `ReverseMap()` is not really clear for everyone that uses AutoMapper. 

With this issue I wanted to state a few questions:

* Why is this behavior not implemented? Are there any specific reasons that are not obvious for us right now? Would be great to here about those reasons because I can´t imagine an issue with this right now.

* How can this behavior be achieved? I think there are 3 possible solutions for this:
  1. Add this behavior to `ReverseMap()` per Default (or via config, or param)
  2. Add another Method like `ForMember` for example `MapFromWithReverse`
  3. Add another Method like `MapFrom`  for example `MapFromWithReverse`

Maybe we can use this issue to discuss the if we should, or if we should not implement this behavior and also if we should do so how´s the best way to get this to work.

lbargaoanu:
Fixes #2067.

### DIFF
--- a/src/AutoMapper/IMappingExpression.cs
+++ b/src/AutoMapper/IMappingExpression.cs
@@ -26,7 +26,7 @@ namespace AutoMapper
         /// Create a type mapping from the destination to the source type, using the destination members as validation.
         /// </summary>
         /// <returns>Itself</returns>
-        IMappingExpression ReverseMap(ReverseOptions reverseOptions = ReverseOptions.Default);
+        IMappingExpression ReverseMap();
 
         /// <summary>
         /// Replace the original runtime instance with a new source instance. Useful when ORMs return proxy types with no relationships to runtime types.
@@ -397,9 +397,8 @@ namespace AutoMapper
         /// <summary>
         /// Create a type mapping from the destination to the source type, using the <typeparamref name="TDestination"/> members as validation
         /// </summary>
-        /// <param name="reverseOptions">what to reverse</param>
         /// <returns>Itself</returns>
-        IMappingExpression<TDestination, TSource> ReverseMap(ReverseOptions reverseOptions = ReverseOptions.Default);
+        IMappingExpression<TDestination, TSource> ReverseMap();
 
         /// <summary>
         /// Customize configuration for an individual source member
@@ -440,17 +439,5 @@ namespace AutoMapper
         /// </summary>
         /// <returns>Itself</returns>
         IMappingExpression<TSource, TDestination> DisableCtorValidation();
-    }
-
-    /// <summary>
-    /// Specify what to reverse.
-    /// </summary>
-    public enum ReverseOptions
-    {
-        Default,
-        /// <summary>
-        /// Reverse simple member accesses MapFrom-s.
-        /// </summary>
-        MapFroms
     }
 }
--- a/src/AutoMapper/IMappingExpression.cs
+++ b/src/AutoMapper/IMappingExpression.cs
@@ -197,7 +197,7 @@ namespace AutoMapper
         /// <param name="memberOptions">Callback for member options</param>
         /// <returns>Itself</returns>
         IMappingExpression<TSource, TDestination> ForPath<TMember>(Expression<Func<TDestination, TMember>> destinationMember,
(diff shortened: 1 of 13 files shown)


src/AutoMapper/IMappingExpression.cs
```diff
@@ -440,4 +441,16 @@ public interface IMappingExpression<TSource, TDestination>
...
+    /// <summary>
+    /// Specify what to reverse.
+    /// </summary>
+    public enum ReverseOptions
+    {
+        Default,
+        /// <summary>
+        /// Reverse simple member accesses MapFrom-s.
+        /// </summary>
+        MapFroms
```
jbogard: Why do you need this option at all? When would you not want this to happen?

lbargaoanu: It would break existing code.

jbogard: Because today those value would just be skipped?

lbargaoanu: Yes, that's what I meant.

jbogard:
Hmmm so the question is SHOULD we break it if that's the behavior people
actually want.

lbargaoanu:
Well, it's easy for me to say, yes, let's break it, after all the code will be cleaner. And surely no matter what some people will not be happy. I guess the question is how many :)

jbogard:
ah, people are always unhappy!

lbargaoanu:
I guess otherwise they wouldn't get any work done :) So, get rid of the enum and change the default behavior?

lbargaoanu:
I've tried to handle the default flattening. The code is a little silly. It just checks that more than one member is involved and it ignores extension methods. But the fact that the tests pass might mean that it's just good enough :)

jbogard:
And yep, let's get rid of the enum.

lbargaoanu:
Done.

src/AutoMapper/PropertyMap.cs
```diff
@@ -37,7 +39,7 @@ public PropertyMap(PropertyMap inheritedMappedProperty, TypeMap typeMap)
 
         public Type DestinationPropertyType => DestinationProperty.GetMemberType();
 
-        public IEnumerable<MemberInfo> SourceMembers => _memberChain;
+        public IList<MemberInfo> SourceMembers => _memberChain;
```
jbogard: What's the motivation to make it a list instead of enumerable?

jbogard: What's the motivation to make it a list instead of enumerable?

lbargaoanu: I'm checking the count. If it's greater than one, I assume it's the default flattening.

lbargaoanu: A readonly interface might work.

PR #2107, opened 2017-05-15
Make object mappers help with validation

### LINKED ISSUE #2101: Make IObjectMappers self-validating
Instead of the configuration validator checking for IObjectMapper types and deciding on a strategy for each, make IObjectMappers validate themselves. For most mappers, this will be a no-op, but for ones that can potentially contain other child types, they can validate themselves.

lbargaoanu:
Fixes #2101.
@jbogard I don't know if this is what you had in mind, but I think it does the job.


jbogard:
Yep! Wasn't sure if IObjectMapper should just have the method, but it looks like most of the time the mappers don't need to do anything. Should IObjectMapperInfo implement IObjectMapper?

lbargaoanu:
I guess it doesn't really make sense on its own. Done.

PR #2110, opened 2017-05-16
AddProfiles() includes internal types when scanning assemblies

### LINKED ISSUE #2063: AddProfiles() should be able to include internal profiles
Currently, the Code for `AddProfilesCore()` calls `GetExportedTypes()` for all assemblies, which only looks at public types. This means that if my current assembly has profiles declared as `internal`, this won't work:
```csharp
    cfg.AddProfiles(Assembly.GetExecutingAssembly());
});
```
It would be nice to be able to include internal profiles in some way.

danielsvr:
- added tests for scanning internal profiles
 - determine the loading strategy based on "ScanForInternalProfiles" property
 - modified the current implementation so that the scanning is done later in the process, and not immediately (more user friendly)

Fixes #2063.

### DIFF
--- a/src/AutoMapper/AdvancedConfiguration.cs
+++ b/src/AutoMapper/AdvancedConfiguration.cs
@@ -22,6 +22,12 @@ namespace AutoMapper
         /// <param name="validator">the validation callback</param>
         public void Validator(Validator validator) => _validators.Add(validator);
 
+        /// <summary>
+        /// Gets or sets a value indicating that internal profiles should be included when scanning assemblies.
+        /// </summary>
+        public bool ScanForInternalProfiles { get; set; }
+
+
         internal Validator[] GetValidators() => _validators.ToArray();
     }
 }
--- a/src/AutoMapper/AdvancedConfiguration.cs
+++ b/src/AutoMapper/AdvancedConfiguration.cs
@@ -22,12 +22,6 @@ namespace AutoMapper
         /// <param name="validator">the validation callback</param>
         public void Validator(Validator validator) => _validators.Add(validator);
 
-        /// <summary>
-        /// Gets or sets a value indicating that internal profiles should be included when scanning assemblies.
-        /// </summary>
-        public bool ScanForInternalProfiles { get; set; }
-
-
         internal Validator[] GetValidators() => _validators.ToArray();
     }
 }
--- a/src/AutoMapper/Configuration/MapperConfigurationExpression.cs
+++ b/src/AutoMapper/Configuration/MapperConfigurationExpression.cs
@@ -63,12 +63,13 @@ namespace AutoMapper.Configuration
 
         private void AddProfilesCore(IEnumerable<Assembly> assembliesToScan)
         {
-            var allTypes = assembliesToScan.Where(a => !a.IsDynamic).SelectMany(a => a.ExportedTypes).ToArray();
+            var allTypes = assembliesToScan.Where(a => !a.IsDynamic).SelectMany(a => a.DefinedTypes).ToArray();
 
             var profiles =
                 allTypes
-                    .Where(t => typeof(Profile).GetTypeInfo().IsAssignableFrom(t.GetTypeInfo()))
-                    .Where(t => !t.GetTypeInfo().IsAbstract);
+                    .Where(t => typeof(Profile).GetTypeInfo().IsAssignableFrom(t))
+                    .Where(t => !t.IsAbstract)
+                    .Select(t => t.AsType());
 
             foreach (var profile in profiles)
             {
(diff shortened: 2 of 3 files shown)


lbargaoanu:
I think all it's needed is to simply include internal types too. All this is too complicated in my view. @jbogard might disagree.

danielsvr:
My approach was to preserve the current behavior and have a switch to include internal types.
I was thinking about alternatives like:
- a new method called something like AddProfilesIcludingInternals
- an overload for AddProfiles that takes a flag for including internal types
but my rational that these ones will clutter the API.

What you are saying is that, AddProfiles should always include internal types and not having a choice to include or exclude?

lbargaoanu:
Yes. One line of code.

jbogard:
Yeah, I think I just picked the wrong method for getting the types out of an assembly. Reflection is different in `TypeInfo` land. I don't know why you *wouldn't* want to pick up all declared Profiles.

danielsvr:
Ok then, I'll update the PR to have only this change.

asherber:
I'm just looking at this now. I'm not sure that including **all** types like this is the way to go, because it uses reflection to get around the declared intent of the authors of the assemblies. My point in the original issue I raised was that it should be easy to include internal profiles in _my own current assembly_ (i.e., `Assembly.GetExecutingAssembly()`), since I already have access to those profiles. I don't think it's good behavior to load internal types from other assemblies.

jbogard:
Assembly.GetExecutingAssembly does not exist in any the `netstandard` targets I support. I'd have to create a `net45`-specific method just to do this. That ain't happening.

This seems reasonable - you still have to opt-in to the assemblies you want to include. I don't scan everything loaded (also, something that does not exist in `netstandard1.x`). If you as a user decide to scan assemblies for profiles that aren't things you wrote, well, that's your prerogative I guess.

asherber:
You're right, I wasn't thinking about `netstandard`. 

Still, even with assembly scanning being opt-in, it seems to go against convention and expectations to include non-public classes by default. I think it would be better to have overloads or new methods for this (as @danielsvr  indicates in his second comment above), even though this clutters the API a bit. Failing that, I would suggest leaving this out altogether, and we can continue to scan by hand if we want to pick up non-public classes.

jbogard:
Why would someone explicitly opt in to scanning an assembly for Profiles, but actually only want the public profiles to be added, are the others like double-secret profiles? I've had plenty of people surprised about this behavior (also in the MediatR/ASP.NET Core scanning) that I saw that *only* scanning public profiles when it's an explicit opt-in seemed more surprising than the opposite behavior.

asherber:
If a class is marked internal, then whoever wrote that assembly intended it only to be used from within the assembly. To turn your question around, why would I mark a profile internal if I expected it to be used from outside the assembly? Again, my initial post had to do only with internal classes in the current assembly.

Other libraries that scan for classes (for example, NUnit / xUnit) only look at public classes.

jbogard:
That's a good question, I have no idea why people mark their Profiles internal and then open GH issues to scan for private/internal assemblies when you can just mark them public.

But they do, for every project I have that explicitly scans an assembly.

asherber:
Okay, we've obviously gotten off on the wrong foot here, and I apologize if I've been unclear or argumentative.

My belief is that this kind of scanning should generally confine itself to types which are visible from the place where the scanning code is invoked. If I initiate an assembly scan from assembly A, I should see only public types from assemblies B and C -- but I should see both public and internal types from assembly A. That seems to me to be in line with how those access modifiers are intended to be used. 

My initial GH issue, which may not have been worded in the best way, was meant to refer to the fact that initiating a scan from assembly A does not see A's own internal types. I was not expecting, nor would I want, to see non-public types from B and C. I agree with you that devs should mark a class public if they want it to be externally visible.

The current PR will show internal types from all assemblies, which seems to me to be undesirable behavior, insofar as it breaches the contract implied by internal. If the choice is just between current state and this PR, my vote would be for current state.

Your opinions seem to be different from mine, so I'll leave the discussion at that. Thanks for this great library, which has saved me lots of work.

jbogard:
So.....revert this? I don't really care from my perspective - I wouldn't use this behavior, I just mark things public. I prefer whatever approach is most obvious to people. Obviously the original PR made it configurable, but I found that to be a bit ugly.

lbargaoanu:
I think we should leave it in. Wait for someone to complain.

asherber:
I would vote for taking it out.

danielsvr:
For me this was my first experience of contributing to open source, and I wanted to take part of the process. I'm don't have a strong opinion on what it will happen. 

I think that this should be the owners decision in the end.

PR #2124, opened 2017-05-25
Detect cycles at config time and set PreserveReferences accordingly

lbargaoanu:
This affects the complex types benchmark. I guess I should modify it to prevent that. It might break some people, but I think we should remove Configuration.GetMappers().

### DIFF
--- a/src/AutoMapper/Execution/TypeMapPlanBuilder.cs
+++ b/src/AutoMapper/Execution/TypeMapPlanBuilder.cs
@@ -82,33 +82,35 @@ namespace AutoMapper.Execution
             {
                 isRoot = false;
             }
-            foreach(var propertyMap in _typeMap.GetPropertyMaps().Where(pm => pm.CanResolveValue()))
-            {
-                CheckPropertyMapForCycles(propertyMap, visitedTypeMaps);
-            }
+
+            CheckPropertyMapsForCycles(visitedTypeMaps);
+
             if(isRoot && visitedTypeMaps.Contains(_typeMap))
             {
                 _typeMap.PreserveReferences = true;
             }
         }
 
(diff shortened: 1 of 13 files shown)


jbogard:
Ready to review?

lbargaoanu:
Yes.

src/AutoMapper/Execution/TypeMapPlanBuilder.cs
```diff
@@ -64,6 +66,64 @@ public LambdaExpression CreateMapperLambda()
...
+            if(propertyMap.SourceType == null)
+            {
+                return null;
+            }
+            var types = new TypePair(propertyMap.SourceType, propertyMap.DestinationPropertyType);
+            var typeMap = _configurationProvider.ResolveTypeMap(types);
+            if(typeMap == null)
+            {
+                var mapper = _configurationProvider.FindMapper(types) as IObjectMapperInfo;
+                if(mapper != null)
```
jbogard: FYI you can do this as `if (_configurationProvider.FindMapper(types) is IObjectMapperInfo mapper)`. See https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-7#is-expression

jbogard:
Do we know what GH issues this would close? I suspect a few.

src/AutoMapper/Execution/TypeMapPlanBuilder.cs
lbargaoanu: Yes, it was smth more compact in an earlier version, but for debugging, it was easier to write it more explicit.

lbargaoanu:
#2068, unless you want UseDestinationValue before somebody requests it. But for other issues, the answer was to set PreserveReferences.

lbargaoanu:
I guess we cannot deprecate PreserveReferences, it might be required in some cases, but we should let people know somehow that it's not usually required.

PR #2153, opened 2017-06-20
Fix ExpressionBuilder GetMapExpression behaviour

Tornhoof:
The typed version of ``GetMapExpression()`` has ``params MemberInfo[]``. Make certain the untyped version creates an empty array otherwise ``ExpressionRequest`` crashes in ctor.

### DIFF
--- a/src/UnitTests/Projection/MapFromTest.cs
+++ b/src/UnitTests/Projection/MapFromTest.cs
@@ -22,20 +22,6 @@ namespace AutoMapper.UnitTests.Projection.MapFromTest
             typeof(NullReferenceException).ShouldNotBeThrownBy(() => config.ExpressionBuilder.GetMapExpression<UserModel, UserDto>()); //null reference exception here
         }
 
-
-        [Fact]
-        public void Should_fail_Typed()
-        {
-            var config = new MapperConfiguration(cfg =>
-            {
-                cfg.CreateMap<UserModel, UserDto>()
-                    .ForMember(dto => dto.FullName, opt => opt.MapFrom(src => src.LastName + " " + src.FirstName));
-            });
-            typeof(ArgumentNullException).ShouldBeThrownBy(() => config.ExpressionBuilder.GetMapExpression<UserModel, UserDto>(null, new MemberInfo[0])); //ArgumentNullException here
-            typeof(ArgumentNullException).ShouldBeThrownBy(() => config.ExpressionBuilder.GetMapExpression<UserModel, UserDto>(new Dictionary<string, object>(), null)); //ArgumentNullException here
-
-        }
-
         [Fact]
         public void Should_fail_Untyped()
         {
--- a/src/UnitTests/Projection/MapFromTest.cs
+++ b/src/UnitTests/Projection/MapFromTest.cs
@@ -21,6 +21,18 @@ namespace AutoMapper.UnitTests.Projection.MapFromTest
         }
 
         [Fact]
+        public void Should_not_fail_Untyped()
+        {
+            var config = new MapperConfiguration(cfg =>
+            {
+                cfg.CreateMap<UserModel, UserDto>()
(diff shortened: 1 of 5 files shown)


lbargaoanu:
It needs a test.

src/UnitTests/Projection/MapFromTest.cs
```diff
@@ -21,6 +21,18 @@ public void Should_not_fail()
...
         [Fact]
+        public void Should_not_fail_Untyped()
+        {
+            var config = new MapperConfiguration(cfg =>
+            {
+                cfg.CreateMap<UserModel, UserDto>()
+                    .ForMember(dto => dto.FullName, opt => opt.MapFrom(src => src.LastName + " " + src.FirstName));
+            });
+
+            typeof(ArgumentNullException).ShouldNotBeThrownBy(() => config.ExpressionBuilder.GetMapExpression(typeof(UserModel), typeof(UserDto), null, null)); //no ArgumentNullException here
```
lbargaoanu: This is not a good test. I'm guessing that the test passes if you don't pass null :)

Tornhoof: I'm not sure I understand, if I comment the fix in ExpressionBuilder the test fails. That's the whole idea of that test, that next time someone refactors it, we don't break it unintentionally?

lbargaoanu: You cannot pass null and complain that it crashes with NRE :) I don't think there is anything to fix here.

lbargaoanu: The null check for parameters made sense when the default was null, as is for the other overload.

Tornhoof: But the Typed API `` Expression<Func<TSource, TDestination>> GetMapExpression<TSource, TDestination>(ParameterBag parameters = null, params MemberInfo[] membersToExpand);`` has a ``params`` for the membersToExpand and the untyped not. the ``params`` will lead to an empty array automatically, even if no ``MemberInfo``s are passed. I just fixed the difference in behaviour when I pass no ``MemberInfo``s. It's just consistent to handle it exactly like your ``parameters = parameters ?? new Dictionary<string, object>();``

lbargaoanu: The fix would be to null check all the parameters, or at least remove the one for the dictionary :)

Tornhoof: Your decision, I changed the checks and now all the args for that method are null checked, obviously had to move the ``parameters ?? new Dictionary<string, object>();`` to a different place (namely ``SourceInjectedQueryProvider`` where the ?? for the ``MemberInfos`` is).
I still don't really like the logical difference between the typed and untyped api

lbargaoanu: It seems in bad taste to have optional parameters in an interface. You do that in some API that uses that interface.

Tornhoof: Changing the optional param is a breaking change as far as I know, but if I'm touching the API there I can add an overload without the optional params (apparently it's usually called without any params at all)

lbargaoanu: This isn't much of an API. So I think it's safe to remove all the default parameters from the interface and add them somewhere else.

lbargaoanu: An extension method if there is no better place.

lbargaoanu:
I think you can remove the typed method from the interface and just use the extension method.

lbargaoanu:
LGTM. The convenience stuff doesn't belong in the interface.

jbogard:
@lbargaoanu the review thingy lets you approve the changes, are you able to do that?

lbargaoanu:
This one should be good to go.

Issue #2161, opened 2017-06-26
Allow ignoring a path in ReverseMap


jbogard:
Create `ForPath.Ignore`

jbogard:
Fixes #2151

PR #2162, opened 2017-06-26
Allow ignores on reverse map paths

jbogard:
Should really have done this with 6.1.0, whoops.

Allow you to ignore a path:

```csharp
cfg =>
    {
		cfg.CreateMap<TimesheetModel, TimesheetViewModel>()
			.ForMember(d=>d.Contact, o=>o.MapFrom(s=>s.ContactNavigation.Id))
			.ReverseMap()
                        .ForPath(s => s.ContactNavigation.Id, opt => opt.Ignore());
    }
```

### DIFF
--- a/src/AutoMapper/Configuration/PathConfigurationExpression.cs
+++ b/src/AutoMapper/Configuration/PathConfigurationExpression.cs
@@ -28,6 +28,11 @@ namespace AutoMapper.Configuration
             MapFromUntyped(sourceExpression);
         }
 
+        public void Ignore()
+        {
+            PathMapActions.Add(pm => pm.Ignored = true);
+        }
+
         public void MapFromUntyped(LambdaExpression sourceExpression)
         {
             _sourceExpression = sourceExpression;
@@ -39,14 +44,6 @@ namespace AutoMapper.Configuration
 
         public void Configure(TypeMap typeMap)
         {
-            //var destMember = DestinationMember;
-
-            //if(destMember.DeclaringType.IsGenericType())
-            //{
-            //    var destTypeInfo = typeMap.Profile.CreateTypeDetails(destMember.DeclaringType);
-            //    destMember = destTypeInfo.PublicReadAccessors.Single(m => m.Name == destMember.Name);
-            //}
-
             var pathMap = typeMap.FindOrCreatePathMapFor(_destinationExpression, MemberPath, typeMap);
 
             Apply(pathMap);
--- a/src/AutoMapper/PathMap.cs
+++ b/src/AutoMapper/PathMap.cs
@@ -6,7 +6,7 @@ namespace AutoMapper
 {
     using Internal;
 
-    [DebuggerDisplay("{" + nameof(DestinationExpression) + "}")]
+    [DebuggerDisplay("{DestinationExpression}")]
     public class PathMap
     {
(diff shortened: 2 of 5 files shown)


jbogard:
Closes #2161

jbogard:
@lbargaoanu take a look if you don't mind

src/AutoMapper/PathMap.cs
```diff
@@ -6,7 +6,7 @@ namespace AutoMapper
 {
     using Internal;
 
-    [DebuggerDisplay("{DestinationExpression}")]
+    [DebuggerDisplay("{" + nameof(DestinationExpression) + "}")]
```
lbargaoanu: This seems wrong :) It's supposed to show the expression and it works now.

src/AutoMapper/Configuration/PathConfigurationExpression.cs
```diff
@@ -39,14 +44,6 @@ public void MapFromUntyped(LambdaExpression sourceExpression)
 
         public void Configure(TypeMap typeMap)
         {
-            //var destMember = DestinationMember;
```
lbargaoanu: This was a reminder that I left smth out, but I guess somebody will eventually complain.

src/AutoMapper/Configuration/PathConfigurationExpression.cs
```diff
@@ -28,6 +28,11 @@ public void MapFrom<TSourceMember>(Expression<Func<TSource, TSourceMember>> sour
             MapFromUntyped(sourceExpression);
         }
 
+        public void Ignore()
+        {
+            PathMapActions.Add(pm => pm.Ignored = true);
```
lbargaoanu: The code is simple enough, but there is some duplication here. We could just copy Ignored to PropertyMap.Ignored and handle both cases (PathMaps and PropertyMaps) in TryPopertyMap.

src/AutoMapper/PathMap.cs
jbogard: R# again...

src/AutoMapper/Configuration/PathConfigurationExpression.cs
jbogard: Ha, should we open another GH issue then?

src/AutoMapper/PathMap.cs
lbargaoanu: :)

src/AutoMapper/Configuration/PathConfigurationExpression.cs
lbargaoanu: I'd rather wait, but if you feel it matters right now...

src/AutoMapper/Configuration/PathConfigurationExpression.cs
jbogard: That might take a bit more refactoring I think, since that method returns an expression and the PropertyMap piece above in `CreateAssignmentFunc` does a similar filter.

jbogard:
@lbargaoanu reverted that R# suggestion

src/AutoMapper/Configuration/PathConfigurationExpression.cs
lbargaoanu: Yes, you would have to return null when CanResolve is false and filter for that.

PR #2167, opened 2017-06-28
Fix for issue #2166

### LINKED ISSUE #2166: Issue with MapExpression when expression checks if a string equals null
When I try to map an expression where there is a check if a string property equals null, I receive the following error:

"InvalidOperationException: The operands for operator 'Equal' do not match the parameters of method 'op_Equality'."

Example:
===================================================================
protected virtual IMapper CreateDefaultMapper<TEntity, TMappedEntity>()
{
      return new MapperConfiguration(cfg =>
      {
                cfg.CreateMap<TEntity, TMappedEntity>().ReverseMap();

      }).CreateMapper();
}
var mapper = CreateDefaultMapper<Order, OrderDto>();
TestMapToDto(mapper, x => x.ShipAddress == null);

void TestMapToDto(IMapper mapper, Expression<Func<OrderDto, bool>> expression)
{
            var mappedExpression = mapper.MapExpression<Expression<Func<Order, bool>>>(expression);
}

===================================================================

I would like to fix it myself, but I'm not really familiar with how to do this, so I'll try and explain what (in my opinion) is the issue.

The error occurs in the class 'XpressionMapperVisitor', in the method "protected override Expression VisitBinary(BinaryExpression node)"

When is change this method so it looks like this;
===================================================================
        protected override Expression VisitBinary(BinaryExpression node)
        {
            if (node.Right is ConstantExpression constantExpression)
            {
                if (constantExpression.Value == null)
                {
                    return base.VisitBinary(node.Left.Type.GetTypeInfo().IsValueType || node.Left.Type == typeof(string)
                        ? node.Update(node.Left, node.Conversion, Expression.Constant(null, node.Left.Type))
                        : node.Update(node.Left, node.Conversion, Expression.Constant(null, typeof(object))));
                }
            }

            constantExpression = node.Left as ConstantExpression;
            if (constantExpression != null)
            {
                if (constantExpression.Value == null)
                {
                    return base.VisitBinary(node.Right.Type.GetTypeInfo().IsValueType || node.Right.Type == typeof(string)
                        ? node.Update(Expression.Constant(null, node.Right.Type), node.Conversion, node.Right)
                        : node.Update(Expression.Constant(null, typeof(object)), node.Conversion, node.Right));
                }
            }

            var newLeft = Visit(node.Left);
            var newRight = Visit(node.Right);
            if ((newLeft.Type.GetTypeInfo().IsGenericType && newLeft.Type.GetGenericTypeDefinition() == typeof(Nullable<>)) ^ (newRight.Type.GetTypeInfo().IsGenericType && newRight.Type.GetGenericTypeDefinition() == typeof(Nullable<>)))
                throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, Resource.cannotCreateBinaryExpressionFormat, newLeft.ToString(), newLeft.Type.Name, newRight.ToString(), newRight.Type.Name));

            return base.VisitBinary(node);
        }
===================================================================

Then the issue is solved.

I added:
-  || node.Left.Type == typeof(string)
and
-  || node.Right.Type == typeof(string)

in the call to the base.VisitBinary(...) method.

Snijderman:
This should be the fix for this issue
["Issue with MapExpression when expression checks if a string equals null"](https://github.com/AutoMapper/AutoMapper/issues/2166)
Closes #2166.

### DIFF
--- a/src/AutoMapper/XpressionMapper/XpressionMapperVisitor.cs
+++ b/src/AutoMapper/XpressionMapper/XpressionMapperVisitor.cs
@@ -97,7 +97,7 @@ namespace AutoMapper.XpressionMapper
             {
                 if (constantExpression.Value == null)
                 {
-                    return base.VisitBinary(node.Left.Type.GetTypeInfo().IsValueType
+                    return base.VisitBinary(node.Left.Type.GetTypeInfo().IsValueType || node.Left.Type == typeof(string)
                         ? node.Update(node.Left, node.Conversion, Expression.Constant(null, node.Left.Type))
                         : node.Update(node.Left, node.Conversion, Expression.Constant(null, typeof(object))));
                 }
@@ -108,7 +108,7 @@ namespace AutoMapper.XpressionMapper
             {
                 if (constantExpression.Value == null)
                 {
-                    return base.VisitBinary(node.Right.Type.GetTypeInfo().IsValueType 
+                    return base.VisitBinary(node.Right.Type.GetTypeInfo().IsValueType || node.Right.Type == typeof(string)
                         ? node.Update(Expression.Constant(null, node.Right.Type), node.Conversion, node.Right) 
                         : node.Update(Expression.Constant(null, typeof(object)), node.Conversion, node.Right));
                 }
--- a/src/UnitTests/XpressionMapperTests.cs
+++ b/src/UnitTests/XpressionMapperTests.cs
@@ -382,10 +382,25 @@ namespace AutoMapper.UnitTests
         }
 
         [Fact]
-        public void Map_accountModel_to_account_with_null_checks_against_string_type()
+        public void Map_accountModel_to_account_with_right_null_checks_against_string_type()
         {
             //Arrange
             Expression<Func<AccountModel, bool>> exp = f => f.Description == null;
+
+
+            //Act
+            Expression<Func<Account, bool>> expMapped = mapper.MapExpression<Expression<Func<Account, bool>>>(exp);
+            List<Account> accounts = Users.Select(u => u.Account).Where(expMapped).ToList();
+
+            //Assert
+            Assert.True(accounts.Count == 0);
+        }
+
+        [Fact]
+        public void Map_accountModel_to_account_with_left_null_checks_against_string_type()
+        {
+            //Arrange
+            Expression<Func<AccountModel, bool>> exp = f => null == f.Description;
 
 
             //Act
--- a/src/UnitTests/XpressionMapperTests.cs
+++ b/src/UnitTests/XpressionMapperTests.cs
@@ -2,11 +2,8 @@
 using Should.Core.Assertions;
 using System;
 using System.Collections.Generic;
-using System.Data;
 using System.Linq;
 using System.Linq.Expressions;
-using System.Text;
-using System.Threading.Tasks;
 using Xunit;
 
 namespace AutoMapper.UnitTests
@@ -385,6 +382,21 @@ namespace AutoMapper.UnitTests
         }
 
         [Fact]
+        public void Map_accountModel_to_account_with_null_checks_against_string_type()
+        {
+            //Arrange
+            Expression<Func<AccountModel, bool>> exp = f => f.Description == null;
+
+
+            //Act
+            Expression<Func<Account, bool>> expMapped = mapper.MapExpression<Expression<Func<Account, bool>>>(exp);
+            List<Account> accounts = Users.Select(u => u.Account).Where(expMapped).ToList();
+
+            //Assert
+            Assert.True(accounts.Count == 0);
+        }
+
+        [Fact]
         public void When_use_lambda_statement_with_typemapped_property_being_other_than_first()
         {
             //Arrange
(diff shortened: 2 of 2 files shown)


jbogard:
@BlaiseD this is your area :)

BlaiseD:
That works.

May have been better to just get rid of the value type check altogether and use `node.Update(node.Left, node.Conversion, Expression.Constant(null, node.Left.Type));` either way.  The same applies to the left side null check.  The related tests (ExpressionConversion, ExpressionMapping and XpressionMapperTests) pass. Is there a 4th group of tests for expression mapping?

lbargaoanu:
That makes sense, I don't see why strings are special.

Snijderman:
@BlaiseD is your comment a general one or do I have to respond? (sorry I'm new to this, don't know how it works :-) )

I agree with your remark, it could be just 1 line. Didn't know why there were 2, just left it so, to be on the save side.

This one 
>>> "Is there a 4th group of tests for expression mapping?" 
I don't really understand.

lbargaoanu:
It's a code review. You implement the changes you do understand and inquire about those you don't.

BlaiseD:
@Snijderman -  It's more of a comment.  My thinking is if System.String cannot be compared to object (like other classes) then it's possible some other reference type will also fail - therefore it makes sense to always convert the null type to match the other operand.

I added the value type check after value types failed - I believe this should have been the correct fix at the time.

Never mind the "4th group of tests" question - it's really addressed to @jbogard .

Snijderman:
I tried to modify it further as suggested by @BlaiseD , but then a couple of tests are failing, so didn't commit that modification.
Added an additional test, so both scenarios (left null and right null) are tested now.

BlaiseD:
@Snijderman - can you say which tests were failing please?

TylerCarlson1:
@lbargaoanu `String` implements interface `IEnumerable` and `IEnumerable<char>`, so when you do collection based things, which in AM's case include IEnumerable you have to exclude strings because you don't want them to work like a collection.

It's something I keep forgetting from time to time and have to remind myself.

lbargaoanu:
Well, yes, I've just wrote that check today, but I didn't know it applied here.

BlaiseD:
@jbogard . The update in this PR by @Snijderman  #2167 is the best approach for now.  #2177 passes (I just realized because it does nothing) but there are expressions created by some external components which will fail (reason for the initial code in VisitBinary) with the #2177 approach.

lbargaoanu:
@BlaiseD OK, but why is string special here?

BlaiseD:
@lbargaoanu - I'm not sure to be honest.  I would have thought it would behave like any other object that is not a value type.   We do know the base ExpressionVisitor throws the exception.  Need some time to investigate further.

lbargaoanu:
I'd say the reason is that is has an equals operator override. I guess this could be merged in as it is, but I'd say it needs more thought.

BlaiseD:
@lbargaoanu Looks like you're right.  To check for the override, the fix will look something like the following. For nullables we need to know if the underlying type has the operator override.

```
Type t = node.Left.Type.GetTypeInfo().IsGenericType && node.Left.Type.GetGenericTypeDefinition().Equals(typeof(Nullable<>))
                        ? Nullable.GetUnderlyingType(node.Left.Type)
                        : node.Left.Type;
MethodInfo m = t.GetDeclaredMethod(node?.Method?.Name);

return base.VisitBinary(m != null
     ? node.Update(node.Left, node.Conversion, Expression.Constant(null, node.Left.Type))
      : node.Update(node.Left, node.Conversion, Expression.Constant(null, typeof(object))));
```
Thoughts?

lbargaoanu:
Nullable.GetUnderlyingType will return null when passed smth else, so those other checks are not needed. We know it's a binary expression and it has a method. So that means it's a binary operator override? I guess we need a few more tests.

lbargaoanu:
And I guess we already have the method, we just need to check method.DeclaringType, no need to look it up. Maybe smth like

    var nodeType = node.Method?.DeclaringType ?? typeof(object);
    return base.VisitBinary(node.Update(node.Left, node.Conversion, Expression.Constant(null, nodeType)));

BlaiseD:
@lbargaoanu Thanks - that's much cleaner. It was really meant to be treated as pseudo code but the approach itself might not be correct.  I'll follow up after the weekend.

BlaiseD:
The initial problem was the following.  Take the mapping and expression below:

```
CreateMap<UserModel, User>().ReverseMap();
Expression<Func<UserModel, bool>> selection = s => s != null && s.AccountModel != null && s.AccountModel.Bal > 555.20;
```
If the null value (item != null) was typed i.e. null is of type UserModel instead of object then expression mapping failed with the exception message: **_"Reference equality is not defined for the types 'ContosoUniversity.Data.Enitities.User' and 'ContosoUniversity.Domain.School.UserModel'."_**  because I had failed to map the constant. There's a new test (Map_object_when_null_values_are_typed) in XpressionMapperTests #2177  demonstrating the issue.

I think the correct approach is to fix the underlying problem in XpressionMapperVisitor.VisitConstant:
```
	protected override Expression VisitConstant(ConstantExpression node)
        {
            Type newType;
            return this.TypeMappings.TryGetValue(node.Type, out newType)
                ? base.VisitConstant(Expression.Constant(node.Value, newType))
                : base.VisitConstant(node);
        }
```
and remove the solution in XpressionMapperVisitor.VisitBinary. Please see #2177 .
Thoughts?

lbargaoanu:
It makes sense. So the string problem was caused by the wrong fix in VisitBinary?

BlaiseD:
Exactly - without the VisitBinary "fix" the string problem goes away.

PR #2173, opened 2017-06-30
Make compatible with net40

sunnycase:
I like AutoMapper very much, but unfurtunately the last version supporting .net4,0 is 4.1.1.
I wrote a TypeInfo and some workaround methods to make AutoMapper compatible with .Net4.0.


lbargaoanu:
This is @jbogard's decision, but I have to say, I don't like it. And it would be painful to have to look at (and support) this kind of code on a regular basis. Maybe smth outside the main project, AutoMapper.Net40.

sunnycase:
@lbargaoanu I agree with you it's painful to write code with net40. What about to create a net40 brunch and publish a AutoMapper.Net40 nupkg?

lbargaoanu:
And how would that help? I mean a branch as opposed to another repository?

sunnycase:
It's up to you. I just said the way ButchersBoy/MaterialDesignInXamlToolkit did.

jbogard:
I'm not completely against it (we already have to have those extension methods anyway). Not as bad as I thought it would be.

One problem we have tho is the tests don't run across all platforms, so I don't know if this actually works on .NET 4.0

jbogard:
After #2180 gets merged, can you rebase and add `net40` as a target in the tests?

sunnycase:
It seems Microsoft.NET.Test.Sdk and xunit don't support net40. :(

jbogard:
sad trombone

sunnycase:
Before #2180 merged, I added net40 AutoMapper.dll reference to unit tests and passed tests. But after merged I can't find a way to add reference net40 output to net452 unit tests targeting net456.

jbogard:
I'm working on your branch, I might have to open a new PR, stay tuned...

TylerCarlson1:
@sunnycase @jbogard There's an option to allow your PR to be editable by maintainer, so you don't have to PR a PR or anything like that.

jbogard:
I rebased :)

jbogard:
PUSH FORCE

jbogard:
So this change, which I think runs OK, runs the xUnit 1.9.2 version against the tests. xUnit 2.x doesn't work, so 4.0 target is skipped.

jnm2:
@sunnycase This is not about AutoMapper but I'm curious as it relates to other libraries. What was keeping a `net45` build of AutoMapper from working for you?

PR #2177, opened 2017-06-30
Fix for #2166 and #2167.

### LINKED ISSUE #2166: Issue with MapExpression when expression checks if a string equals null

BlaiseD:
Replacement PR for #2167.
Fixes #2166.

### DIFF
--- a/src/AutoMapper/XpressionMapper/XpressionMapperVisitor.cs
+++ b/src/AutoMapper/XpressionMapper/XpressionMapperVisitor.cs
@@ -84,39 +84,47 @@ namespace AutoMapper.XpressionMapper
                     ? visitor.Visit(last.CustomExpression.Body.AddExpressions(afterCustExpression))
                     : visitor.Visit(last.CustomExpression.Body);
 
+                this.TypeMappings.AddTypeMapping(node.Type, ex.Type);
                 return ex;
             }
             fullName = BuildFullName(propertyMapInfoList);
             var me = InfoDictionary[parameterExpression].NewParameter.BuildExpression(fullName);
+
+            this.TypeMappings.AddTypeMapping(node.Type, me.Type);
             return me;
+        }
+
+        protected override Expression VisitConstant(ConstantExpression node)
+        {
+            Type newType;
+            if (this.TypeMappings.TryGetValue(node.Type, out newType))
+                return base.VisitConstant(Expression.Constant(node.Value, newType));
+
+            return base.VisitConstant(node);
         }
 
         protected override Expression VisitBinary(BinaryExpression node)
         {
-            if (node.Right is ConstantExpression constantExpression)
-            {
-                if (constantExpression.Value == null)
-                {
-                    node.Update(node.Left, node.Conversion, Expression.Constant(null, node.Left.Type));
-                    //return base.VisitBinary(node.Update(node.Left, node.Conversion, Expression.Constant(null, node.Left.Type)));
-                    //return base.VisitBinary(node.Left.Type.GetTypeInfo().IsValueType
-                        //? node.Update(node.Left, node.Conversion, Expression.Constant(null, node.Left.Type))
-                        //: node.Update(node.Left, node.Conversion, Expression.Constant(null, typeof(object))));
-                }
-            }
+            //if (node.Right is ConstantExpression constantExpression)
+            //{
+            //    if (constantExpression.Value == null)
+            //    {
+            //        return base.VisitBinary(node.Left.Type.GetTypeInfo().IsValueType
+            //            ? node.Update(node.Left, node.Conversion, Expression.Constant(null, node.Left.Type))
+            //            : node.Update(node.Left, node.Conversion, Expression.Constant(null, typeof(object))));
+            //    }
+            //}
 
-            constantExpression = node.Left as ConstantExpression;
-            if (constantExpression != null)
-            {
-                if (constantExpression.Value == null)
-                {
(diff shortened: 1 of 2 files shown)


src/UnitTests/XpressionMapperTests.cs
```diff
@@ -147,6 +147,39 @@ public void Map__object_including_child_and_grandchild()
 
             //Assert
             Assert.True(users.Count == 2);
+        }
+
+        [Fact]
+        public void Map_object_when_null_values_are_typed()
+        {
+            //Arrange
+            //Expression<Func<UserModel, bool>> selection = s => s != null && s.AccountModel != null && s.AccountModel.Bal > 555.20;
```
lbargaoanu: Casting nulls doesn't do the job?

    Expression<Func<UserModel, bool>> selection = s => s != (UserModel) null && s.AccountModel != (AccountModel) null && s.AccountModel.Bal > 555.20;

src/AutoMapper/XpressionMapper/XpressionMapperVisitor.cs
```diff
@@ -34,7 +34,7 @@ protected override Expression VisitParameter(ParameterExpression parameterExpres
         }
 
         protected override Expression VisitMember(MemberExpression node)
-        {
+        {
             if (node.NodeType == ExpressionType.Constant)
```
lbargaoanu: It can be a member expression and a constant?

src/UnitTests/XpressionMapperTests.cs
BlaiseD: No - you need the Expression API to set the type.

lbargaoanu: Aha, so I guess someone generates these expressions.

src/AutoMapper/XpressionMapper/XpressionMapperVisitor.cs
BlaiseD: Quite likely that code has no value.  The first curly brace below never gets hit stepping through the tests.

```
            if (node.NodeType != ExpressionType.MemberAccess)
            {
            }
```


 don't recall why I wrote it 2 years ago.

lbargaoanu:
I closed that other PR. Will you clean this up or make a new one?

src/UnitTests/XpressionMapperTests.cs
BlaiseD: Yes - mostly some meta programming component.

BlaiseD:
Done

lbargaoanu:
LGTM.

BlaiseD:
Thanks guys.
@Snijderman  - your tests have been included in the #2177 .

PR #2180, opened 2017-06-30
Sdk style tests

### LINKED ISSUE #2179: Make the test projects SDK style and xplat
And also execute tests on all supported platforms.

This will mean we need to get off of Fixie and onto xUnit.

jbogard:
Fixes #2179


lbargaoanu:
As far as I can tell, in VS tests run only on .NET 4.5, but much slower.

jbogard:
Yeah, that's just for people that don't have any other test runner. As punishment I guess, it also only supports .NET 4.5, not .NET Core

lbargaoanu:
All right :) So what do you use, ReSharper?

jbogard:
TestDriven.net but R# when I want to do test sessions. That package is just
to enable xUnit at least from their docs

lbargaoanu:
I'm unable to get any significant speed up by running the tests in parallel, but maybe, for testing purposes, it would still be a better idea to run them in parallel.

jbogard:
Ah yes I guess we don't use the static API in the tests

PR #2185, opened 2017-07-04
Allow simple MapFrom-s to pass validation

### LINKED ISSUE #2172: AssertConfigurationIsValid with MemberList.Source doesn't recognize inherited ForMember mapping
Given the scenario in this gist https://gist.github.com/michaelaird/df1c79d944a8e9ce0fde51f149879bf0 
(note the source field name is an intentional mis-spelling, this is a real example from our code)

### Expected behavior

Configuration should be valid

### Actual behavior

AssertConfigurationIsValid claims Prepopulate is not mapped

lbargaoanu:
Fixes #2172. This won't work for more complicated expressions. If you feel it's inconsistent, just close it.

### DIFF
--- a/src/AutoMapper/TypeMap.cs
+++ b/src/AutoMapper/TypeMap.cs
@@ -110,16 +110,16 @@ namespace AutoMapper
 
         public string[] GetUnmappedPropertyNames()
         {
-            string GetFunc(PropertyMap pm) => ConfiguredMemberList == MemberList.Destination
+            string GetProperty(PropertyMap pm) => ConfiguredMemberList == MemberList.Destination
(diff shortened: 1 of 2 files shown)


jbogard:
Not following how this fixes the original issue - shouldn't the MapFrom have fixed the configuration problem in the derived map? Or it's because it's looking at source members.

lbargaoanu:
The mapping works. It's not valid because it validates the source and that extra check that I removed. I guess a more general solution would be to consider mapped all source members referenced in the MapFrom.

jbogard:
Well, let's see if this works for folks. I suspect that most want just the `MemberList.None` in any case.

PR #2190, opened 2017-07-06
Use Resolve instead of Find for expression generation

TylerCarlson1:
Conventions can't use UseAsDataSource without calling map first because when generate expressions it doesn't try to resolve type map, just find if it exists.

### DIFF
--- a/src/AutoMapper/QueryableExtensions/Impl/QueryMapperHelper.cs
+++ b/src/AutoMapper/QueryableExtensions/Impl/QueryMapperHelper.cs
@@ -24,7 +24,7 @@ namespace AutoMapper.QueryableExtensions.Impl
 
         public static TypeMap CheckIfMapExists(this IConfigurationProvider config, Type sourceType, Type destinationType)
         {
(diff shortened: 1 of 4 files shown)


src/AutoMapper/QueryableExtensions/Impl/QueryMapperHelper.cs
```diff
@@ -24,7 +24,7 @@ public static PropertyMap GetPropertyMap(this IConfigurationProvider config, Mem
 
         public static TypeMap CheckIfMapExists(this IConfigurationProvider config, Type sourceType, Type destinationType)
         {
-            var typeMap = config.FindTypeMapFor(sourceType, destinationType);
+            var typeMap = config.ResolveTypeMap(sourceType, destinationType);
```
lbargaoanu: I don't know about this one...

lbargaoanu: But it seems to be used only for QueryableExtensions, so I was probably thinking of smth else :) Clearly you usually want Resolve, not Find.

PR #2200, opened 2017-07-10
Type map plan builder refactor

jbogard:
Making the TypeMapPlanBuilder only build type map expressions and nothing else. Broke out other concerns to a separate static class we can do `using static` with.

### DIFF
--- a/src/AutoMapper/Execution/ExpressionBuilder.cs
+++ b/src/AutoMapper/Execution/ExpressionBuilder.cs
@@ -1,0 +1,150 @@
+namespace AutoMapper.Execution
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+    using System.Linq.Expressions;
+    using AutoMapper.Configuration;
+    using AutoMapper.Internal;
+    using AutoMapper.Mappers.Internal;
+    using static System.Linq.Expressions.Expression;
+
+    public static class ExpressionBuilder
+    {
+        private static readonly Expression<Func<IRuntimeMapper, ResolutionContext>> CreateContext =
+            mapper => new ResolutionContext(mapper.DefaultContext.Options, mapper);
+
+        public static Expression MapExpression(IConfigurationProvider configurationProvider,
+            ProfileMap profileMap,
+            TypePair typePair,
+            Expression sourceParameter,
+            Expression contextParameter,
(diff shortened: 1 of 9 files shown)


TylerCarlson1:
So is this the start of consolidating all the expression generation logic into a single set of places, so they can be refactored and broken up into EF6, EF.Core, ect mapping types?

Also like the `using static` better then calling static type, or trying to make them extension methods.

jbogard:
Kinda. Really I just wanted to understand what we have today and I couldn't
tell.

TylerCarlson1:
Yea it's nearly impossible right now to read this stuff as a whole.  Also possibility of mixing up ProjectTo with MapPlanBuilder both being expression based.  I don't blame yea.  Whatever it takes to make expressions easier is well worth it IMHO.

src/AutoMapper/Execution/ExpressionBuilder.cs
```diff
@@ -62,29 +64,29 @@ Expression ClearDestinationCollection()
...
                 var destinationElementType = ElementTypeHelper.GetElementType(destinationParameter.Type);
                 var destinationCollectionType = typeof(ICollection<>).MakeGenericType(destinationElementType);
-                var destinationVariable = Expression.Variable(destinationCollectionType, "collectionDestination");
+                var destinationVariable = Variable(destinationCollectionType, "collectionDestination");
                 var clearMethod = destinationCollectionType.GetDeclaredMethod("Clear");
-                var clear = Expression.Condition(Expression.Property(destinationVariable, "IsReadOnly"),
-                    Expression.Empty(), Expression.Call(destinationVariable, clearMethod));
-                return Expression.Block(new[] {destinationVariable},
-                    Expression.Assign(destinationVariable,
+                var clear = Condition(Property(destinationVariable, "IsReadOnly"),
```
TylerCarlson1: I would suggest for static calls with 1-2 maybe 3 variables use Extension methods like in my cleanup/optimize [PR](https://github.com/AutoMapper/AutoMapper/pull/1346/files#diff-d58ee48fd6246bad6783dc201ff089aaR68)

Then like will look like
```
var clear = destinationVariable.Property("IsReadOnly").Condition(Empty(), destinationVariable.Call(clearMethod))
```
And hopefully that fluent API will actually be easier to read when people look back at this code.

PR #2218, opened 2017-07-17
Add static ResolutionContext.Map MethodInfo field in ExpressionBuilder

Tornhoof:
This is just a minor refactoring, currently ``ContextMap`` uses ``.First`` with the method name to get the proper methodinfo, this depends on the order of methods in the ``ResolutionContext`` class, there are already two methods called ``Map``, now it uses the ``Method`` idea from ``ExpressionFactory``, this should prevent suprises when refactoring ``ResolutionContext``.
I added another of the ``Method`` methods to the ``ExpressionFactory`` to support non-static methods.

No extra tests were added as this method is called in many unit tests, if you want to see how it fails, use the old code and swap the ``Map`` methods in ``ResolutionContext``, I get ~150 failed tests then.

### DIFF
--- a/src/AutoMapper/Internal/ExpressionFactory.cs
+++ b/src/AutoMapper/Internal/ExpressionFactory.cs
@@ -19,9 +19,11 @@ namespace AutoMapper.Internal
                         null;
         }
 
-        public static MethodInfo Method<T>(Expression<Func<T>> expression) => ((MethodCallExpression) expression.Body).Method;
+        public static MethodInfo Method<T>(Expression<Func<T>> expression) => GetExpressionBodyMethod(expression);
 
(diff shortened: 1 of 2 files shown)


src/AutoMapper/Internal/ExpressionFactory.cs
```diff
@@ -21,6 +21,8 @@ public static Expression GetSetter(MemberExpression memberExpression)
 
         public static MethodInfo Method<T>(Expression<Func<T>> expression) => ((MethodCallExpression) expression.Body).Method;
 
+        public static MethodInfo Method<TType, TResult>(Expression<Func<TType, TResult>> expression) => ((MethodCallExpression)expression.Body).Method;
+
```
lbargaoanu: There is some duplication here. Both methods could forward to another one that does the job.

Tornhoof: Aye, I'll add another method which does the body stuff for both,

PR #2222, opened 2017-07-18
Validate included types

### LINKED ISSUE #2192: Validation success when TSourceBase or TDestinationBase in IncludeBase<TSourceBase, TDestinationBase>() are not actual base types
### Source/destination classes
```csharp
class A { }
class A1 : A { }
class B { }

class ADto { }
class A1Dto : ADto { }
class BDto { }
```
### Mapping configuration
```csharp
CreateMap<A1, A1Dto>().IncludeBase<B, BDto>(); // B is not base class for A1 and BDto is not for A1Dto
```

### Expected behavior
Validation error if `TSourceBase` is not `TSource` or base class for `TSource` (same for `TDestinationBase` and `TDestination`), in this case, `B` is not base for `A1` and `BDto` is not base for `A1Dto`.

### Actual behavior
Validation successfully passed, target object successfully mapped.

### Steps to reproduce

```csharp
{
    cfg.CreateMap<A1, A1Dto>().IncludeBase<B, BDto>();
});

var source_a = new A1();
var target_a = mapper.Map<A1Dto>(source_a);
```

lbargaoanu:
Fixes #2192.

### DIFF
--- a/src/AutoMapper/Configuration/MappingExpression.cs
+++ b/src/AutoMapper/Configuration/MappingExpression.cs
@@ -239,12 +239,18 @@ namespace AutoMapper.Configuration
         public IMappingExpression<TSource, TDestination> Include<TOtherSource, TOtherDestination>()
             where TOtherSource : TSource
             where TOtherDestination : TDestination 
-            => Include(typeof(TOtherSource), typeof(TOtherDestination));
+            => IncludeCore(typeof(TOtherSource), typeof(TOtherDestination));
 
         public IMappingExpression<TSource, TDestination> Include(Type otherSourceType, Type otherDestinationType)
         {
-            TypeMapActions.Add(tm => tm.IncludeDerivedTypes(otherSourceType, otherDestinationType));
+            CheckIsDerived(otherSourceType, typeof(TSource));
+            CheckIsDerived(otherDestinationType, typeof(TDestination));
+            return IncludeCore(otherSourceType, otherDestinationType); ;
+        }
 
+        IMappingExpression<TSource, TDestination> IncludeCore(Type otherSourceType, Type otherDestinationType)
+        {
+            TypeMapActions.Add(tm => tm.IncludeDerivedTypes(otherSourceType, otherDestinationType));
             return this;
         }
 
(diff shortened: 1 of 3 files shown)


src/AutoMapper/Configuration/MappingExpression.cs
```diff
@@ -239,22 +239,29 @@ private void IgnoreDestinationMember(MemberInfo property)
...
             where TOtherSource : TSource
             where TOtherDestination : TDestination 
-            => Include(typeof(TOtherSource), typeof(TOtherDestination));
+            => IncludeCore(typeof(TOtherSource), typeof(TOtherDestination));
 
         public IMappingExpression<TSource, TDestination> Include(Type otherSourceType, Type otherDestinationType)
+        {
+            CheckIsDerived(otherSourceType, typeof(TSource));
+            CheckIsDerived(otherDestinationType, typeof(TDestination));
+            return IncludeCore(otherSourceType, otherDestinationType); ;
```
jbogard: Whoops ;) ;)

src/AutoMapper/Configuration/MappingExpression.cs
```diff
@@ -239,22 +239,29 @@ private void IgnoreDestinationMember(MemberInfo property)
...
+        IMappingExpression<TSource, TDestination> IncludeCore(Type otherSourceType, Type otherDestinationType)
         {
-            TypeMapActions.Add(tm => tm.IncludeDerivedTypes(otherSourceType, otherDestinationType));
-
+            TypeMapActions.Add(tm => tm.IncludeDerivedTypes(otherSourceType, otherDestinationType));
             return this;
-        }
+        }
 
         public IMappingExpression<TSource, TDestination> IncludeBase<TSourceBase, TDestinationBase>()
```
jbogard: Can we do the reverse here too? Where TSource : TSourceBase?

lbargaoanu: I've tried :) Unless I'm missing some syntactic stuff, we cannot.

src/AutoMapper/Configuration/MappingExpression.cs
lbargaoanu: I don't see it :)

lbargaoanu: OK, _now_ I see it :)

src/AutoMapper/Configuration/MappingExpression.cs
lbargaoanu: We need to constrain TSourceBase, not TSource, and there is no constraint is a base class of. That seems to be the idea.

jbogard: It seems we cannot bend the laws of C# to our will then.

lbargaoanu: Not today :)

PR #2229, opened 2017-07-20
Ignore member paths

### LINKED ISSUE #2199: Ignore a set of paths based on ignoring a matching member
See #2195 

Given our types:

```c#
// source type

       public class Dto
        {
            public int AddressId { get; set; }
            public string AddressLine1 { get; set; }
            public string AddressLine2 { get; set; }
        }

// destination types

        public class Model
        {
            public int AddressId { get; set; }
            public Address Address { get; set; }
        }
        public class Address
        {
            public int Id { get; set; }
            public string Line1 { get; set; }
            public string Line2 { get; set; }
        }
```

If we ignore a destination member:

```c#
            Mapper.Initialize(
                cfg =>
                {
                    cfg.CreateMap<Model, Dto>()
                        .ReverseMap()
                        .ForMember(x => x.Address, x => x.Ignore());
                }
            );
```

Ignore all matching paths for that member.

Open question - should we also allow hierarchical path ignoring? Ignore a parent path, all child paths are ignored?

lbargaoanu:
Fixes #2199.
Kind of convoluted. Maybe it's the implementation, but I'd say it's the requirements.
- ForPath should make the first member pass validation
- ignoring the first member ignores all the paths that start with it
- an explicit ForPath should override the ignored first member

Maybe we should give up a few of these :)

### DIFF
--- a/src/AutoMapper/Configuration/MappingExpression.cs
+++ b/src/AutoMapper/Configuration/MappingExpression.cs
@@ -179,11 +179,14 @@ namespace AutoMapper.Configuration
                 throw new ArgumentOutOfRangeException(nameof(destinationMember), "Only member accesses are allowed.");
             }
             var expression = new PathConfigurationExpression<TSource, TDestination>(destinationMember);
-            IgnoreDestinationMember(expression.MemberPath.First);
+            var firstMember = expression.MemberPath.First;
+            var firstMemberConfig = GetDestinationMemberConfiguration(firstMember);
+            if(firstMemberConfig == null)
+            {
+                IgnoreDestinationMember(firstMember, ignorePaths: false);
+            }
             _memberConfigurations.Add(expression);
-
             memberOptions(expression);
-
             return this;
         }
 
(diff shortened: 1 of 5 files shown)


jbogard:
@Tornhoof how does this look?

Tornhoof:
From a user perspective, I don't understand the new public ``Ignore(bool)`` method, why would call that externally with ``false``?
Does a version without allowing an explicit ForPath to unignore the path make the code much easier or less convoluted? If yes then I'd say skip this requirement and let people map it all explicitly.

lbargaoanu:
That method is not part of the interface, that's the API. I'll make it internal just to be more explicit (I'm not exactly a fan of internal). 
About the second question, I didn't try it, but I think that's secondary. The core of the conflict seems to be between ignoring the first member because ForPath maps it somehow and making Ignore also ignore paths that start with it. Another idea is to handle differently paths we create, but I'm not so sure about that either.

Tornhoof:
Ah yes, sorry about the ``Ignore(bool)`` I didn't notice that it's not in the interface.
Thanks for the more detailed explanation about the FirstMember problem, I still think (from a user perspective) that hierarchical ignore is quite useful, but I understand your conflict.

PR #2232, opened 2017-07-21
Consolidate destination nullable mappings

### LINKED ISSUE #1987: Consolidate nullable logic out of IObjectMappers
We have nullable logic all over the place. Both in terms of checking types and doing mapping junk. I want to move the nullable stuff that's junking up our IObjectMappers out.

Here's the basic logic for two values:

Source | Dest | Src Value | Dest Value
--------|-------|-------|------
T|U?|T|Map(T, U)
T|U?|default(T)|Map(T, U)
T?|U|T|Map(T, U)
T?|U|null|default(U)
T?|U?|T|Map(T, U)
T?|U?|null|default(U?)
V|U|V|Map(V, U)
V|U|null|default(U)
V|U?|V|Map(V, U)
V|U?|null|default(U?)

where T: struct, V: class

I think we could wrap this logic in one place for top-level maps and member-level maps.

jbogard:
Fixes #1987


lbargaoanu:
Nice :)

TylerCarlson1:
Who knew it would be so simple to refactor, and not break anything? :)

jbogard:
Well, there are some more places to consolidate. Gonna set this to WIP

PR #2233, opened 2017-07-21
We have to set everything in the plan, we don't get a second chance

lbargaoanu:
It's like...be aggressive :)

### DIFF
--- a/src/AutoMapper/Execution/TypeMapPlanBuilder.cs
+++ b/src/AutoMapper/Execution/TypeMapPlanBuilder.cs
@@ -44,7 +44,7 @@ namespace AutoMapper.Execution
 
         public ParameterExpression Context { get; }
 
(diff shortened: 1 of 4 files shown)


jbogard:
Does this fix another issue?

lbargaoanu:
https://stackoverflow.com/questions/45215011/self-referencing-loop-causing-stack-overflow-with-map/

lbargaoanu:
This is too aggressive I think :) The original idea was not bad, if you start from a map and you need it again while building the plan, you have recursion. But Seal is called only once, that's why it didn't work. Now just because you use a map twice in different levels, it would call it recursion. But it isn't. If I want to do this in one pass I think I need to also store the maps that lead to where I am in the plan.

lbargaoanu:
I think this is done.

PR #2252, opened 2017-08-05
Add PreCondition overload that takes source and ResolutionContext as parameters

### LINKED ISSUE #2251: Add PreCondition overload
In IMemberConfigurationExpression there is a PreCondition method that passes source, and there is a PreCondition method that passes ResolutionContext, however there isn't one that passes both.

____

Internally, in the MemberConfigurationExpression implementation, both the existing PreCondition methods set an expression property that takes both the source and ResolutionContext as parameters.

It also appears that TypeMapPlanBuilder passes both parameters to the internal PreCondition expression property.

So I don't think much would need to change to implement this.

ESPNSTI:
Added an overload to PreCondition that had both TSource and ResolutionContext as parameters.
Closes #2251.

### DIFF
--- a/src/AutoMapper/Configuration/MemberConfigurationExpression.cs
+++ b/src/AutoMapper/Configuration/MemberConfigurationExpression.cs
@@ -222,6 +222,17 @@ namespace AutoMapper.Configuration
             });
         }
 
+        public void PreCondition(Func<TSource, ResolutionContext, bool> condition)
+        {
+            PropertyMapActions.Add(pm =>
+            {
+                Expression<Func<TSource, ResolutionContext, bool>> expr =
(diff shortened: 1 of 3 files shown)


lbargaoanu:
The test is missing.

ESPNSTI:
There's no test for the ResolutionContext overload either. :)

I'll see what I can do.

lbargaoanu:
Maybe it's GitHub's fault, but it would be nice if you could fix the diff.

ESPNSTI:
I'm not sure what's going on, I thought I had set autocrlf to false.
I'll check and see if I can fix it.
Some of the files seem to have inconsistent line endings it seems.

PR #2256, opened 2017-08-07
Fixing issue #2255

### LINKED ISSUE #2255: Expression that uses a member of an implemented interface
In my project, I am trying to map from a interface to an object.
To make it easier for you to understand the problem, I will explain it with the models from XpressionMapperTests.cs

### Source/destination types

```csharp

    public class User
    {
        public int UserId { get; set; }
        public string FirstName { get; set; }
        public string LastName { get; set; }
        public bool IsLoggedOn { get; set; }
        public int Age { get; set; }
        public bool Active { get; set; }
        public Account Account { get; set; }
    }


    public interface IIdentifiable
    {
        int Id { get; set; }
    }

    public interface IUserModel : IIdentifiable
    {
        string FullName { get; set; }
        string AccountName { get; set; }
        bool IsOverEighty { get; set; }
        string LoggedOn { get; set; }
        int AgeInYears { get; set; }
        bool IsActive { get; set; }
    }

```

### Mapping configuration

```csharp

CreateMap<User, IUserModel>()
                .ForMember(d => d.Id, opt => opt.MapFrom(s => s.UserId))
                .ForMember(d => d.FullName, opt => opt.MapFrom(s => string.Concat(s.FirstName, " ", s.LastName)))
                .ForMember(d => d.LoggedOn, opt => opt.MapFrom(s => s.IsLoggedOn ? "Y" : "N"))
                .ForMember(d => d.IsOverEighty, opt => opt.MapFrom(s => s.Age > 80))
                .ForMember(d => d.AccountName, opt => opt.MapFrom(s => s.Account == null ? string.Empty : string.Concat(s.Account.Branch.Name, " ", s.Account.Branch.Id.ToString())))
                .ForMember(d => d.AgeInYears, opt => opt.MapFrom(s => s.Age))
                .ForMember(d => d.IsActive, opt => opt.MapFrom(s => s.Active));

            CreateMap<IUserModel, User>()
                .ForMember(d => d.UserId, opt => opt.MapFrom(s => s.Id))
                .ForMember(d => d.FirstName, opt => opt.MapFrom(s => s.FullName))
                .ForMember(d => d.IsLoggedOn, opt => opt.MapFrom(s => s.LoggedOn.ToUpper() == "Y"))
                .ForMember(d => d.Age, opt => opt.MapFrom(s => s.AgeInYears))
                .ForMember(d => d.Active, opt => opt.MapFrom(s => s.IsActive));
```

### Version: x.y.z
AutoMapper 6.1.1

### Expected behavior
When I map an expression that converts from a direct IUserModel interface member, the mapping works:
`Expression<Func<IUserModel, bool>> selection = s => s.LoggedOn == "Y"`

if I however use the member from the implemented interface "IIdentifiable", ...
`Expression<Func<IUserModel, bool>> selection = s => s.Id == 14;`

### Actual behavior

I get an InvalidOperationException, because the XpressionMappingVisitor cannot find the target property.

### Steps to reproduce

* Add the above interfaces IUsesrModel and IIdentifiable, to the XpressionMapperTests.
* Make UserModel implement IUserModel
* add the following unittests:

```csharp

        [Fact]
        public void Map_works_with_members_from_interfaces()
        {
            //Arrange
            Expression<Func<IUserModel, bool>> selection = s => s.LoggedOn == "Y";

            //Act
            Expression<Func<User, bool>> selectionMapped = mapper.Map<Expression<Func<User, bool>>>(selection);
            List<User> users = Users.Where(selectionMapped).ToList();

            //Assert
            Assert.True(users.Count == 1);
        }

        [Fact]
        public void Map_works_with_members_from_base_interfaces()
        {
            //Arrange
            Expression<Func<IUserModel, bool>> selection = s => s.Id == 14;

            //Act
            Expression<Func<User, bool>> selectionMapped = mapper.Map<Expression<Func<User, bool>>>(selection);
            List<User> users = Users.Where(selectionMapped).ToList();

            //Assert
            Assert.True(users.Count == 1);
        }

```

gentledepp:
Fixes #2255.
You can not map an expression from an interface to a class and thereby even use members of base-interfaces.


src/AutoMapper/TypeExtensions.cs
```diff
@@ -36,18 +36,35 @@ public static Type CreateType(this TypeBuilder type)
 
         public static IEnumerable<MemberInfo> GetAllMembers(this Type type)
         {
-            while (true)
+            if (type.IsInterface())
```
lbargaoanu: This doesn't look good. One idea is to always check for implemented interfaces.

gentledepp: What do you mean by that?
I just tried to keep the amount of changes as little as possible.

lbargaoanu: I mean an extra test will always fix an issue, but that's not good code. Try to write general code that will work in all cases.

gentledepp: Ok, so what would you suggest? Adding another extensionmethod to the TypeExtensions class like 
`public static IEnumerable<MemberInfo> GetAllMembersIncludingInterfaces( this Type type)`

Any precise suggestions are very welcome. I did not develop this feature, I just made a bugfix for something we need in production ;-)

src/AutoMapper/XpressionMapper/XpressionMapperVisitor.cs
```diff
@@ -231,7 +231,7 @@ protected void FindDestinationFullName(Type typeSource, Type typeDestination, st
             if (sourceFullName.IndexOf(period, StringComparison.OrdinalIgnoreCase) < 0)
             {
                 var propertyMap = typeMap.GetPropertyMaps().SingleOrDefault(item => item.DestinationProperty.Name == sourceFullName);
-                var sourceMemberInfo = typeSource.GetMember(propertyMap.DestinationProperty.Name).First();
+                var sourceMemberInfo = typeSource.GetAllMembers().First(mi => mi.Name == propertyMap.DestinationProperty.Name);
```
BlaiseD: There appears to be a conflict in XpressionMapperVisitor  between Type.GetMember and TypeExtensions.GetMember.  Consider calling TypeExtensions explicitly i.e. `var sourceMemberInfo = TypeExtensions.GetMember(typeSource, propertyMap.DestinationProperty.Name).First();` instead of changing the logic.

lbargaoanu:
Well, I went ahead and messed up your PR. Sorry about that :) If all else fails, we can use the git history to get it back. Hopefully we won't have to.

src/AutoMapper/XpressionMapper/XpressionMapperVisitor.cs
```diff
@@ -254,7 +254,7 @@ protected void FindDestinationFullName(Type typeSource, Type typeDestination, st
                 var propertyName = sourceFullName.Substring(0, sourceFullName.IndexOf(period, StringComparison.OrdinalIgnoreCase));
                 var propertyMap = typeMap.GetPropertyMaps().SingleOrDefault(item => item.DestinationProperty.Name == propertyName);
                 
-                var sourceMemberInfo = typeSource.GetAllMembers().First(mi => mi.Name == propertyMap.DestinationProperty.Name);
+                var sourceMemberInfo = typeSource.GetInheritedMember(propertyMap.DestinationProperty.Name);
```
gentledepp: This might be confusing, as the original "TypeHelper.GetAllMembers()" function also returns inherited members.
What I really changed is, that it also returned interface members
So maybe it should be called `GetClassAndInterfaceMembers` ?

src/AutoMapper/Internal/TypeHelper.cs
```diff
@@ -1,169 +0,0 @@
...
-        public static MethodInfo GetSetMethod(PropertyInfo propertyInfo, bool ignored) => propertyInfo.GetSetMethod();
-#else
-        public static MethodInfo GetGetMethod(PropertyInfo propertyInfo, bool ignored) => propertyInfo.GetMethod;
-
-        public static MethodInfo GetSetMethod(PropertyInfo propertyInfo, bool ignored) => propertyInfo.SetMethod;
-#endif
-
-        public static FieldInfo GetField(Type type, string name) => type.GetRuntimeField(name);
-    }
-}
```
gentledepp: I don't see what this has to do with my PR - I guess you should have created another branch/PR for this, but ok ;-)

src/AutoMapper/TypeExtensions.cs
```diff
@@ -58,16 +58,13 @@ public static IEnumerable<Type> GetTypeInheritance(this Type type)
         public static MethodInfo GetDeclaredMethod(this Type type, string name) => type.GetAllMethods().FirstOrDefault(mi => mi.Name == name);
 
         public static MethodInfo GetDeclaredMethod(this Type type, string name, Type[] parameters) =>
-                type.GetAllMethods()
-                .Where(mi => mi.Name == name)
-                .Where(mi => mi.GetParameters().Length == parameters.Length)
-                .FirstOrDefault(mi => mi.GetParameters().Select(pi => pi.ParameterType).SequenceEqual(parameters));
+                type.GetAllMethods().Where(mi => mi.Name == name).MatchParameters(parameters);
```
gentledepp: Again, what does this have to do with my PR?
Just curious :-|

src/AutoMapper/MapperConfiguration.cs
```diff
@@ -18,7 +18,7 @@ namespace AutoMapper
 
     public class MapperConfiguration : IConfigurationProvider
     {
-        private static readonly ConstructorInfo ExceptionConstructor = typeof(AutoMapperMappingException).GetConstructors().Single(c => c.GetParameters().Length == 3);
+        private static readonly ConstructorInfo ExceptionConstructor = typeof(AutoMapperMappingException).GetDeclaredConstructors().Single(c => c.GetParameters().Length == 3);
```
gentledepp: ... same question :-)

lbargaoanu:
@BlaiseD You might want to take a look.

BlaiseD:
That's pretty impressive - did not know about Expression.MakeMemberAccess.  Also - good call removing the MethodInfo check from AddPropertyMapInfo.  The sourceFullName always come from a MemberExpression.

gentledepp:
merge it merge it merge it

PR #2261, opened 2017-08-09
Value transformers in global, profiles, type maps, and members

jbogard:


jbogard:
cc/ @asherber

src/AutoMapper/Execution/TypeMapPlanBuilder.cs
```diff
@@ -522,6 +522,10 @@ private Expression BuildValueResolverFunc(PropertyMap propertyMap, Expression de
                     );
             }
 
+            valueResolverFunc = _typeMap.Profile.ValueTransformers
+                .Where(vt => vt.IsMatch(propertyMap))
+                .Aggregate(valueResolverFunc, (current, vtConfig) => ToType(ReplaceParameters(vtConfig.TransformerExpression, ToType(current, vtConfig.ValueType)), propertyMap.DestinationPropertyType));
```
TylerCarlson1: Can this change to having ValueTransformerConfiguration have function likeMapExpression?  Kinda like what IObjectMapper does now where it's an interface like IValueTransformerConfiguration.

jbogard: Yeah that's a further task on the list

jbogard: I wasn't going to make it like IObjectMapper though, more like IValueResolver and ITypeConverter. You can go the expression route, which can be used in LINQ etc. or the runtime route that can get access to ResolutionContext. That would also allow a DI-able version.

jbogard:
This one's ready to review.

TylerCarlson1:
What do you think about making `ValueTransformerConfiguration` an interface with that being an implementation of it.  That way it's more versatile to transform things in the future.  Also people will be able to make their own custom ones based on whatever they want if they so desire.

The only issue with this if you want to keep the same public function calls `ApplyTransform` is should probably be an extension method and you would have to supply the List of transforms so the extension method can access and add to the list.

Unless this is meant for another PR in the future, then can discuss it there.

jbogard:
Hrm that's a big departure from the rest of the configuration API. I think you already have the ability to do similar though with ForAllTypeMaps?

TylerCarlson1:
But for all type maps doesn't apply to individual property maps and can't be profile specific.  That's the whole reason for transforms.

Also your current implementation is only for property maps of same type.  What about different types being mapped what about comparing by some kind of naming convention.  If you keep the API as it is for each one of them you'll have to keep adding more and more functions to the interface, as well as their implementation in multiple places away from the base class.  Where as if you use extension methods they can be confined to their own classes near their respected implementations of `IValueTransformerConfiguration` for profiles and member configurations.

`ValueTransformerConfiguration` can have a static extensions class in the same location as it, maybe even a sub class in the same file, with all the method calls that use it.  Instead of every configuration interface having an implementation in 2-3 separate files that you have to search for if you want to change something.

Yes it's a big change in how the API is structured, but one thing I don't like about working with the current one is that the MapperConfiguration, Profile, and everything else have interfaces that have like 20-30 properties/functions that do almost the exact same things using a class and keep adding functionality with more and more function calls is going to make navigating them even worse to wade through.

I think this makes the code more easily expandable, without having to hack around in private properties or submit PR's and wait for the next version if you want to do something extra.  But then it exposes things that you might not want the user to worry about in configuration.  It's a trade off, the big one being change how API is structured, but from the end user experience I don't think it will change, unless there's some weird thing between Extension methods that I'm not aware of.

jbogard:
PR for a PR?

I can see some benefits to the flexibility for sure, I do get a little wary
of just extension methods since they aren't nearly as discoverable. I would
be interested in a concrete example of what you're suggesting though.

src/AutoMapper/ValueTransformerConfiguration.cs
```diff
@@ -0,0 +1,22 @@
+using System;
+using System.Linq.Expressions;
+
+namespace AutoMapper
+{
+    public class ValueTransformerConfiguration
```
TylerCarlson1: It has function calls for IsMatch.  Would have to change it to a `Func<PropertyMap, bool>` for it to be a struct I think.

lbargaoanu: I don't follow.

TylerCarlson1: If it's a struct instead of a class `public bool IsMatch(PropertyMap propertyMap)` shouldn't exist, was my understanding of things.

Therefore you would need `public Func<PropertyMap, bool> IsMatch { get; }` because it's not a function but a reference to one.

lbargaoanu: No, structs can have member functions.

src/UnitTests/ValueTransformers.cs
```diff
@@ -0,0 +1,247 @@
...
+            public class Dest
+            {
+                public string Value { get; set; }
+            }
+
+            protected override MapperConfiguration Configuration { get; } = new MapperConfiguration(cfg =>
+            {
+                cfg.CreateMap<Source, Dest>();
+                cfg.ApplyTransform<string>(dest => dest + " is straight up dope");
+                cfg.ApplyTransform<string>(dest => dest + "! No joke!");
```
jbogard: Maybe the wording can be better? But yes, they'd stack is the idea.

lbargaoanu: I think it's easy to make a mistake. You have a lot of config code and you have one ApplyTransform and a few pages below, oops, another one. Not exactly the pit of success :)

lbargaoanu: I think AddTransform would be better and exposing a list even clearer.

jbogard: Then in that case I think the extension method would make a lot more sense. Expose the list directly of value transform configs, and an extension method to add a transform with an expression:

```
cfg.AddTransform<string>(dest => dest + "!!!");
cfg.ValueTransformers.Add(new ValueTransformer { typeof(string), expr });
```

lbargaoanu: I see your point, but I don't mind
   
    cfg.ValueTransformers.Add<string>(dest => dest + "!!!");

It is two dots. But it makes more obvious what's going on. I'm not sure if you wanted to do other things with that list.

asherber: I've been following this at a little bit of a distance. FWIW, from a functionality standpoint I like the idea of stacking transforms, but it seems like a departure from the rest of the API, which I think just does overrides. On the whole I think I would vote for `ApplyTransform()` and overrides, rather than `AddTransform()` and stacking, but I don't feel strongly about it.

jbogard: @lbargaoanu I gotcha, add the extension method to the list of value transformers rather than the root. I don't want to create a custom collection type.

@asherber I think we could add conditions to transforms that would allow for turning things off (or don't apply things globally).

TylerCarlson1: @jbogard @lbargaoanu I kinda like extending off the list better.  Less mass options from Intelisense and groups functionality by property.

Also can make it fluent where you add and then return list, so you can stack, and know you are stacking.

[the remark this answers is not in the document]
jbogard: OK I made some changes here lemme know what you think. I couldn't go extension methods all the way because it would have required forcing generic arguments or a weird chaining.

src/AutoMapper/Execution/TypeMapPlanBuilder.cs
```diff
@@ -522,6 +522,12 @@ private Expression BuildValueResolverFunc(PropertyMap propertyMap, Expression de
                     );
             }
 
+            valueResolverFunc = propertyMap.ValueTransformers
+                .Concat(_typeMap.ValueTransformers)
+                .Concat(_typeMap.Profile.ValueTransformers)
```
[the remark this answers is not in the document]
lbargaoanu: A few hundred mappings, maybe a few thousand properties, it seems reasonable. For all the existing transformers. This will hurt eventually :)

src/UnitTests/ValueTransformers.cs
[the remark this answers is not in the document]
lbargaoanu: The list extension method thing works for profiles, but not for type maps. It breaks the DSL. We could chain through the list, but that seems wrong because we could compose with ForAllMembers or As, for example, right?

src/AutoMapper/IMemberConfigurationExpression.cs
```diff
@@ -199,6 +199,13 @@ void ResolveUsing<TValueResolver>()
         /// The destination member being configured.
         /// </summary>
         MemberInfo DestinationMember { get; }
+
+        /// <summary>
+        /// Apply a transformation function after any resolved destination member value with the given type
+        /// </summary>
+        /// <typeparam name="TValue">Value type to match and transform</typeparam>
+        /// <param name="transformer">Transformation expression</param>
+        void AddTransform<TValue>(Expression<Func<TValue, TValue>> transformer);
```
lbargaoanu: Maybe a non generic overload that just uses TMember.

src/AutoMapper/Execution/TypeMapPlanBuilder.cs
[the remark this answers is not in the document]
jbogard: What would we cache? The list could be different since you can define at the property map level.

Alternatively we could push the transformers down from the parent levels, kinda like we do for Include.

src/AutoMapper/IMemberConfigurationExpression.cs
jbogard: I don't know if we need an overload as much as just get rid of this version. I don't remember why I put a transform where you could specify a different type. Transformers get applied after resolving/mapping so it's always going to be of the destination member type.

src/AutoMapper/Execution/TypeMapPlanBuilder.cs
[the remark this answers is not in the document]
lbargaoanu: Yes, but not the transformers per member are the problem, all the rest are. And yes, I thought about smth like your second suggestion, but I'm not sure how that would look like, caching by type seemed more obvious.

src/AutoMapper/IMemberConfigurationExpression.cs
lbargaoanu: After resolving and before mapping. And I think you meant for TValue to be the type of the resolved value. Which doesn't have to be TMember. Right? :)

jbogard: Ah then I wonder if I have value transformations in the right place. I think it's really before you assign it, you transform it but I have it as part of resolving.

src/AutoMapper/Execution/TypeMapPlanBuilder.cs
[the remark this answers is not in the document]
jbogard: I think this will only be an issue with large amounts of value transformers. And I'm not sure that would really happen

src/AutoMapper/IMemberConfigurationExpression.cs
jbogard: I moved the value transformations to after mapping and right before setting.

src/AutoMapper/Execution/TypeMapPlanBuilder.cs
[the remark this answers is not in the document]
lbargaoanu: I guess we can wait :)

src/AutoMapper/IMemberConfigurationExpression.cs
lbargaoanu: I like it. I've always thought AfterPropertyMap is a good idea :)

src/UnitTests/ValueTransformers.cs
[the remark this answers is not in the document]
lbargaoanu: I think this one is blocking.

jbogard:
Anything else on this one?

src/AutoMapper/IMemberConfigurationExpression.cs
jbogard: Ha well maybe this will take care of people's scenarios.

src/UnitTests/ValueTransformers.cs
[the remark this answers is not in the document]
jbogard: What's still blocked here? I made the extension method on Profile/Mapper's ValueTransformers, but had to keep AddTransformer in TypeMap. I think that's at least consistent with each's DSL.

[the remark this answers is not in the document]
lbargaoanu: Yes 😊 I got confused somehow. I thought it was not chaining properly. Sorry about that.

asherber:
Thank you!

Issue #2266, opened 2017-08-10
ArgumentOutOfRangeException if you put IgnoreMap attribute on a destination type property that there is no equivalent property to it in the source type


mjalil:
This might be 2 issues, but since both of them are from a single line of code, I mention them together:
1- Silently ignores mapping in the reverse direction (although, this behavior has some advocates)
2- throws an exception if you put `IgnoreMap` attribute on a destination type property that there is no equivalent property to it in the source type
The cause: https://github.com/AutoMapper/AutoMapper/blame/49e81dff7801259be2779f3054161c3a8b6cedaf/src/AutoMapper/Internal/MappingExpression.cs#L616
### Source/destination types

```csharp
public class BlogPost
{
    public long Id {get; set;}
    public string Body {get; set;}
    public DateTime PublishDate {get; set;}
    public string Publisher {get; set;}
}
public class BlogPostDto
{
    public long Id {get; set;}
    public string Body {get; set;}
    public DateTime PublishDate {get; set;}
    [IgnoreMap]
    public string Publisher {get; set;}
    [IgnoreMap]
    public List<BlogPostDto> RelatedPosts {get; set;}
}
```

### Mapping configuration

```csharp
Mapper.Initialize(cfg => {
    cfg.CreateMap<BlogPost, BlogPostDto>();
});
```

### Expected behavior

Ignore map `BlogPost.Publisher -> BlogPostDto.Publisher` but not ignore `BlogPostDto.Publisher -> BlogPost.Publisher`. also, IgnoreMap on `RelatedPosts` should not throw an exception, silently ignore it.
I just want to permanently ignore mapping of RelatedPosts now and at the future.
It was working In version 2.2 of the library as I expected. I think it is ok in 4.2.1 too. but not after v5.0.0-beta-1
### Actual behavior

silently ignores mapping of Publisher in the reverse direction.
and throws an ArgumentOutOfRangeException: 'Cannot find a field or property named RelatedPosts'
Now, after upgrading from 2.2 to 6.1.1 I have to search the source code and delete `IgnoreMap`s or initialize the mapper like this:
```csharp
Mapper.Initialize(cfg => {
    cfg.CreateMap<BlogPost, BlogPostDto>()
    .ReverseMap()
    .ForMember(entity => entity.Publisher, opt => opt.MapFrom(dto => dto.Publisher));
});
```
both solutions are search and edit work for me. but in old version it was not required
### Steps to reproduce

```csharp
// use the sample above ;)
[...]

PR #2273, opened 2017-08-12
Don't throw ArgumentOutOfRangeException when there is no reverse property to IgnoreMap. issue #2266

mjalil:
I want to Ignore a property mapping even if currently there is no equivalent property in the source type to ignore. I want to Ignore it in destination type permanently, now and in the future. so I put an `IgnoreMap` attribute on the destination property

### DIFF
--- a/src/AutoMapper/Configuration/Internal/PrimitiveHelper.cs
+++ b/src/AutoMapper/Configuration/Internal/PrimitiveHelper.cs
@@ -17,16 +17,13 @@ namespace AutoMapper.Configuration.Internal
         private static IEnumerable<MemberInfo> GetAllMembers(this Type type) =>
             type.GetTypeInheritance().Concat(type.GetTypeInfo().ImplementedInterfaces).SelectMany(i => i.GetDeclaredMembers());
 
-        private static MemberInfo GetInheritedMember(this Type type, string name) => type.GetAllMembers().FirstOrDefault(mi => mi.Name == name);
+        public static MemberInfo GetInheritedMember(this Type type, string name) => type.GetAllMembers().FirstOrDefault(mi => mi.Name == name);
 
         public static MethodInfo GetInheritedMethod(Type type, string name)
             => type.GetInheritedMember(name) as MethodInfo ?? throw new ArgumentOutOfRangeException(nameof(name), $"Cannot find method {name} of type {type}.");
 
         public static MemberInfo GetFieldOrProperty(Type type, string name) 
             => type.GetInheritedMember(name) ?? throw new ArgumentOutOfRangeException(nameof(name), $"Cannot find member {name} of type {type}.");
-
-        public static bool HasFieldOrProperty(Type type, string name)
-            => type.GetInheritedMember(name) != null;
 
         public static bool IsNullableType(Type type) 
             => type.IsGenericType(typeof(Nullable<>));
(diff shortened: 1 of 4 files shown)


src/AutoMapper/Configuration/MappingExpression.cs
```diff
@@ -541,7 +541,14 @@ public void Configure(TypeMap typeMap)
                 if (attrs.Any(x => x is IgnoreMapAttribute))
                 {
                     IgnoreDestinationMember(destProperty);
-                    _reverseMap?.ForMember(destProperty.Name, opt => opt.Ignore());
+                    try
+                    {
+                        _reverseMap?.ForMember(destProperty.Name, opt => opt.Ignore());
```
jbogard: Is there a better way to do this than try-catch? Like checking to see if the member exists first?

mjalil: yes, it is better. I just wrote a quick fix. try-catch smells bad. I will change that as you said

src/AutoMapper/Configuration/Internal/PrimitiveHelper.cs
```diff
@@ -25,6 +25,9 @@ public static MethodInfo GetInheritedMethod(Type type, string name)
         public static MemberInfo GetFieldOrProperty(Type type, string name) 
             => type.GetInheritedMember(name) ?? throw new ArgumentOutOfRangeException(nameof(name), $"Cannot find member {name} of type {type}.");
 
+        public static bool HasFieldOrProperty(Type type, string name)
```
mjalil: @lbargaoanu where would be the test class?
a new class in the `AutoMapper.UnitTests.Bug` namesapce 
or in the `AutoMapper.UnitTest.ReverseMapIgnoreAttributeTests` class

and I want to name it `Sould_not_throw_exception_when_reverse_property_does_not_exist`, OK?

lbargaoanu: ReverseMapIgnoreAttributeTests seems best.

src/AutoMapper/Configuration/MappingExpression.cs
```diff
@@ -540,8 +541,11 @@ public void Configure(TypeMap typeMap)
                 var attrs = destProperty.GetCustomAttributes(true);
                 if (attrs.Any(x => x is IgnoreMapAttribute))
                 {
-                    IgnoreDestinationMember(destProperty);
-                    _reverseMap?.ForMember(destProperty.Name, opt => opt.Ignore());
+                    IgnoreDestinationMember(destProperty);
+                    if (typeMap.SourceType.GetInheritedMember(destProperty.Name) != null)
+                    {
+                        _reverseMap?.ForMember(destProperty.Name, opt => opt.Ignore());
```
mjalil: No, I don't. here the destination of (reverse) mapping is the source. I should ignore source property with the same **name** as `destProperty.Name` **if exists**.
`IgnoreDestinationMember` takes a `MemberInfo` as a parameter, not a **probably existing** property name. I can not get a `MemberInfo` from source if it does not exist there. I have to check for its existence first.
line 544 ignores in the forward direction. line 545 checks for source property existence, line 547 ignores source property. if this is wrong, what is the right way?
what do you mean by saying 

> You already have the member

lbargaoanu: GetInheritedMember returns the member you have to ignore.

mjalil: maybe you expected a code such this?
```csharp
    var sourceProperty = typeMap.SourceType.GetInheritedMember(destProperty.Name);
    if (sourceProperty != null)
    {
        _reverseMap?IgnoreDestinationMember(sourceProperty);
    }
```

PR #2279, opened 2017-08-16
[Example] Moving functionality to ExtensionMethods

TylerCarlson1:
@jbogard This is an extrapolation of the idea talked about in #2261.
I moved Condition and MaxDepth logic to single file with all the logic working with those properties for both TypeMap and PropertyMap.  Not everything can be extracted out with extension methods and have no code change for every user.

This moves some of the clutter of the Configuration and Execution Generation classes, while consolidating logic into a single space.  If this continues TypeMapPlanBuilder.cs would become more readable in the steps used to generate an expression, instead of trying to decipher Expression code when looking for something.

And for issues with MaxDepth, Condition, or other things, you can go to a single file to view all the logic, instead of weeding through the many places it's being used trying to remember what does what from memory.

The layout isn't final in any way.  It's just moving stuff around and trying to clean up some of the larger kinda god classes, TypeMapPlanBuilder.cs.


lbargaoanu:
The whole extension methods thing is just like writing C code, you have some data and you pass it to functions. If you wanted to make TypeMapPlanBuilder more readable, you could have smaller member functions that used expressions and have higher level code that read more like pseudocode for the mapping logic. You could even separate that into a separate class. I don't see any need for that myself, but at least it would be OO code and it would not pass ten parameters as it did before when everything was static.

jbogard:
Hrmmmmmm that is rather interesting.

I wonder though - we've still got the issue in the TypeMap side where I have more or less hard-coded options on what can be interpreted by type map plan builders (of which there are now many).

TylerCarlson1:
I'm not sure what the issue you are talking about.  We aren't going to change the hard coded nature of the TypeMapPlanBuilder necessarily, just move the logic to another class.  Also not everything would need to be refactored out to its own class.  Things would be like PreserveReferences, maybe before & after maps could be moved out, but other things can stay or get consolidated into a single class like custom resolvers/expressions logic.

If you are talking about TypeMapPlan vs Projection vs Expression Mapping and there differences I would try to push the logic for those out into different extension identifying there intent.

Again not sure what you mean.  An example of class or function talking about would be nice.

@lbargaoanu you are right.  That is one of the downsides to this, so the payoff from this I want is that it's easier to check functionality in code when someone has an issue.  If you have a problem with PreserveReferences you can go to a single class and see what it does in every possible use case to see if there would be faults.

Also another way to make it more object oriented might be to have some extension methods go off of TypeMapPlanBuilder therefore making it more like internal methods where it doesn't have to pass over sub properties as much.

jbogard:
Moving to extension methods means we've got a bit of decoupling from the configuration model (TypeMap/ProfileMap/PropertyMap) and the DSL. Which I don't have any problem with, I never really thought about it but with the configuration model now being applied via a bunch of actions...

The only funny thing we'd have though are configuration options in our configuration model that are explicitly for one kind of mapping. I wonder if there's something we can allow more dynamic application of configuration in our TypeMap/ProfileMap/PropertyMap etc. This is how ASP.NET Core config works for example, configs are stuffed in a dictionary for "add-ons". So you could do something like:

```
var valueTransformerConfig = typeMap.GetConfig<TransformerConfig>();

vs.

var valueTransformerConfig = typeMap.ValueTransformers;
```

TylerCarlson1:
So you would want Type/Profile/Property Maps to have some kind of dictionary that hold all the action classes and then only allow extension methods to add certain types to certain classes?  So then there's no more ValueTransformers, and you have to load them from the dictionary?

I would think going that far would make it more confusing when you start extracting it.  Yes you decouple the Configuration classes even further down to almost a struct but then reading the code might become harder again.  It also doesn't need to be THAT configurable, because it still has to follow a define set of rules.  Maybe I could be wrong, I just don't know about that right now.

Also I played with it more and making TypeMapPlanBuilder have all the fields properties means I can make smaller extension methods that only have to pass 1-2 parameters in.  Haven't pushed yet.

lbargaoanu:
Yes, but I think the main benefit of that is that it's exposed externally.

TylerCarlson1:
@jbogard [ConstructorMap.cs](https://github.com/AutoMapper/AutoMapper/blob/71e1190bd55bf4409d527c91fd672406f35cf1ab/src/AutoMapper/ConstructorMap.cs) is kinda an idea of moving it all to one file.  Most, if not all, major constructor logic is here, so if you have to look at constructor code it will all be here and be organized by functionality.  Names for function calls probably need work though.

Have namespaces for MapPlan and QueryableExtensions that do all the functionality for those sections.

Regular Extensions are the Configuration Functionality and stay in the AutoMapper namespace so that the API remains unchanged.  As well as the comments for the public API.

src/AutoMapper/Execution/TypeMapPlanBuilder.cs
```diff
@@ -5,548 +5,88 @@ namespace AutoMapper.Execution
...
-
-            var mapperFunc = CreateMapperFunc(assignmentFunc);
-
-            var checkContext = CheckContext(_typeMap, Context);
+                return Lambda(customExpression.ReplaceParameters(Source, InitialDestination, Context), Source,
+                    InitialDestination, Context);
+
+            this.CheckForCycles(typeMapsPath);
+
+            var mapperFunc = Block(
```
TylerCarlson1: This block is now much more readable in what goes on and all the function calls are located in a class that consolidates the functionality.

TylerCarlson1:
@jbogard made some more updates on this to move the expression building for the maps out of TypeMapPlanBuilder and into files that have classes under AutoMapper.Execution namespace with that functionality.

Same treatment would apply to AutoMapper.QueryableExtensions and moving functionality out of ExpressionBuilder.cs

Things like if you want to look at before map functionality you can go to BeforeMap.cs and see it has an AutoMapper and AutoMapper.Execution namespace so it applies to Configuration and Execution, where as if you go to MaxDepth.cs you will see AutoMapper.QueryableExtensions as well, so you know max depth is being applied to Query generation like ProjectTo

jbogard:
OK so in the next major release I want to move non-core mapping out into individual repos, I think that would be a good time to consider this change as we'll see immediately what individual mappers need for configuration

TylerCarlson1:
Moving things out would be beyond the scope of what this does.  And a lot of new hassles.

But when you get to that point you are going to have to make a more concise configuration interface that can be extended and added onto

PR #2280, opened 2017-08-17
Create proxies on .Net Core

### LINKED ISSUE #2278: Interface proxy on .Net Core 1.1
When mapping a concrete class to an interface wihtout a concrete implementation, Automapper is not resolving the dynamic proxy in a .Net Core 1.1 Project (In full framework is working). I've see the excecution plan and everithing is ok, it only seems like the proxy object is null.

### Source/destination types

```csharp
public interface ISuperClass
    {
        int Id { get; set; }
        string Name { get; set; }
    }

    public class ConcreteClass
    {
        public int Id { get; set; }
    }
```

### Mapping configuration

```csharp
Mapper.Initialize(cfg => { cfg.CreateMap<ConcreteClass, ISuperClass>(); }); 
```

### Expected behavior

A new proxy instance is created with properties mapped.

### Actual behavior

NullReferenceException

### Steps to reproduce

```csharp
[Fact]
//The same code in FullFramework is working
        public void Test1()
        {
            Mapper.Initialize(cfg => { cfg.CreateMap<ConcreteClass, ISuperClass>(); });
            var a = new ConcreteClass() { Id = 1, Name = "test" };
            Assert.NotNull(Mapper.Map<ISuperClass>(a));
        }
```

lbargaoanu:
Fixes #2278.

### DIFF
--- a/src/AutoMapper/XpressionMapper/Extensions/MapperExtensions.cs
+++ b/src/AutoMapper/XpressionMapper/Extensions/MapperExtensions.cs
@@ -23,12 +23,6 @@ namespace AutoMapper.XpressionMapper.Extensions
             if (expression == null)
                 return default(TDestDelegate);
 
-            if (expression.GetType().GetGenericTypeDefinition() != typeof(Expression<>)
-                || typeof(TDestDelegate).GetGenericTypeDefinition() != typeof(Expression<>))
-            {
-                throw new ArgumentException(Resource.mustBeExpressions);
-            }
-
             var typeSourceFunc = expression.GetType().GetGenericArguments()[0];
             var typeDestFunc = typeof(TDestDelegate).GetGenericArguments()[0];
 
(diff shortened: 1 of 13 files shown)


lbargaoanu:
@jbogard Can I update the UnitTestProject to .Net Core 2.0?

src/AutoMapper/XpressionMapper/Extensions/MapperExtensions.cs
```diff
@@ -23,12 +23,6 @@ public static TDestDelegate MapExpression<TDestDelegate>(this IMapper mapper, La
             if (expression == null)
                 return default(TDestDelegate);
 
-            if (expression.GetType().GetGenericTypeDefinition() != typeof(Expression<>)
```
lbargaoanu: This fails in .Net Core 2.0 and fixing it seemed too much work for parameter validation.

jbogard: Yeah agree. I was working on the upgrade too but if you've got it going then 👍

PR #2282, opened 2017-08-18
Null check static methods too

### LINKED ISSUE #2276: AutoMapper calls extension methods with absent (null) source object
### Source/destination types
```
public class Foo
{
	public Bar Bar { get; set; }
}

public class Bar
{
}

public class FooDto
{
	public string BarSimpleName { get; set; }
}

public static class BarExtensions
{
	public static string GetSimpleName( this Bar source )
	{
		if (source == null)
			throw new ArgumentNullException(nameof(source));
		return "SimpleName";
	}
}
```
### Mapping configuration
```
Mapper.Initialize(v => {
	v.IncludeSourceExtensionMethods(typeof(BarExtensions));
	v.CreateMap<Foo, FooDto>();
});
```
### Expected behavior
Receive null (default value) in property BarSimpleName of the type FooDto.

### Actual behavior
ArgumentNullException in BarExtensions.GetSimpleName method (source equals to null). Why AutoMapper (AM) calls this method for the absent (null) source object? In AM 3.2.1 (I'm not checked any higher versions after it) it is worked all fine.

### Steps to reproduce
```
var input = new Foo {
	/* Works fine when uncommented.
	Bar = new Bar {
		Name = "Name"
	}
	*/
};
var output = Mapper.Map<FooDto>(input);
```

lbargaoanu:
Fixes #2276.
Sorry for the whitespace thing, the test is at the top.

### DIFF
--- a/src/AutoMapper/Execution/ExpressionBuilder.cs
+++ b/src/AutoMapper/Execution/ExpressionBuilder.cs
@@ -44,8 +44,8 @@ namespace AutoMapper.Execution
             }
             var objectMapperExpression = ObjectMapperExpression(configurationProvider, profileMap, typePair,
                 sourceParameter, contextParameter, propertyMap, destinationParameter);
-            return NullCheckSource(profileMap, sourceParameter, destinationParameter, objectMapperExpression,
-                propertyMap);
+            var nullCheckSource = NullCheckSource(profileMap, sourceParameter, destinationParameter, objectMapperExpression, propertyMap);
+            return ExpressionFactory.ToType(nullCheckSource, typePair.DestinationType);
         }
 
         public static Expression NullCheckSource(ProfileMap profileMap,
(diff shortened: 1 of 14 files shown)


lbargaoanu:
This needs more work.

src/AutoMapper/Execution/ExpressionBuilder.cs
```diff
@@ -75,45 +74,22 @@ Expression ClearDestinationCollection()
...
-                return defaultValue;
-            if (destinationType.IsArray)
+            if(profileMap.AllowNullCollections || destinationType == typeof(string) || !destinationType.IsEnumerableType())
             {
-                var destinationElementType = destinationType.GetElementType();
-                return NewArrayBounds(destinationElementType,
-                    Enumerable.Repeat(Constant(0), destinationType.GetArrayRank()));
+                return Default(declaredDestinationType);
             }
-            if (destinationType.IsDictionaryType())
```
lbargaoanu: This is mostly unrelated but it's something that was bothering me. It makes sense to move the null check for the source here, but the mappers know about the destination types. It's messy to repeat all that here.

src/UnitTests/NullBehavior.cs
```diff
@@ -50,27 +50,6 @@ public void Should_map_to_non_null()
         }
     }
 
-    public class When_mapping_untyped_null_to_IEnumerable : AutoMapperSpecBase
```
lbargaoanu: No mapper would apply here, so I think this should throw after all. Casting the source should work as expected.

PR #2286, opened 2017-08-22
Issue 2246 - UseAsDataSource throwing ArgumentNullException.

### LINKED ISSUE #2246: NullReferenceException with UseAsDataSource
Hi. I've having some trouble using "UseAsDataSource" to query a model queryable with a DTO based where clause. These are the details ([GIST provided](https://gist.github.com/darkcamper/9b6ec4109662a24d3b09069579991090))

### Source/destination types

```csharp
class DTO
{
  public class DTONested
  {
    public bool AnotherBoolean { get; set; }
  }

  public DTONested Nested { get; set; }
}

class Model
{
  public bool ABoolean { get; set; }
}
```

### Mapping configuration

```csharp
cfg.CreateMap<Model, DTO>()
       .ForMember(d => d.Nested, opt => opt.MapFrom(s => s));
cfg.CreateMap<Model, DTO.DTONested>()
       .ForMember(d => d.AnotherBoolean, opt => opt.MapFrom(s => s.ABoolean));
cfg.CreateMap<DTO, Model>()
       .ForMember(d => d.ABoolean, opt => opt.MapFrom(s => s.Nested.AnotherBoolean));
```

### What's happening
If I create an Expression<Func<DTO,bool>> , map it to Expression<Func<Model,bool>> and use it in a Where() call it succeeds as expected

```csharp
Expression<Func<DTO, bool>> expOverDTO = (dto) => dto.Nested.AnotherBoolean;
Expression<Func<Model, bool>> expOverModel = 
     mapper.Map<Expression<Func<DTO, bool>>, Expression<Func<Model, bool>>>(expOverDTO);

var result1 = queryable.Where(expOverModel).ToList();
```
But if I try a similar approach with "UseAsDataSource"  in this way:

```csharp
var result2 = queryable.UseAsDataSource(mapper).For<DTO>().Where(expOverDTO).ToList();
```
A NullReference exception is thrown with the following stack trace:

```
AutoMapper.HashCodeCombiner.Combine[T1,T2](T1 obj1, T2 obj2)
   en System.Collections.Generic.GenericEqualityComparer`1.GetHashCode(T obj)
   en System.Collections.Generic.Dictionary`2.FindEntry(TKey key)
   en System.Collections.Generic.Dictionary`2.TryGetValue(TKey key, TValue& value)
   en AutoMapper.MapperConfiguration.FindTypeMapFor(Type sourceType, Type destinationType)
   en AutoMapper.Mappers.ExpressionMapper.MappingVisitor.GetConvertedSubMemberCall(MemberExpression node)
...
```

I've tried looking in the docs and browsing some old issues but I haven't found any clue about it. Any idea what I'm doing wrong?

Thanks.

dnikolovv:
Fixes #2246.

### DIFF
--- a/src/AutoMapper/Mappers/ExpressionMapper.cs
+++ b/src/AutoMapper/Mappers/ExpressionMapper.cs
@@ -243,21 +243,14 @@ namespace AutoMapper.Mappers
                 return newExpression;
             }
 
-            private Type GetSourceType(PropertyMap propertyMap)
-            {
-                if (propertyMap.SourceMember == null)
-                {
-                    return propertyMap.SourceType ??
-                        throw new AutoMapperMappingException(
-                            "Could not determine object source type.", 
-                            null, 
-                            new TypePair(propertyMap.SourceType, propertyMap.DestinationPropertyType), 
-                            propertyMap.TypeMap, 
-                            propertyMap);
-                }
-
-                return propertyMap.SourceMember.GetMemberType();
-            }
+            private Type GetSourceType(PropertyMap propertyMap) =>
+                propertyMap.SourceType ??
+                throw new AutoMapperMappingException(
+                    "Could not determine source property type. Make sure the property is mapped.", 
+                    null, 
+                    new TypePair(propertyMap.SourceType, propertyMap.DestinationPropertyType), 
+                    propertyMap.TypeMap, 
+                    propertyMap);
 
             private PropertyMap FindPropertyMapOfExpression(MemberExpression expression)
             {
--- a/src/AutoMapper/Mappers/ExpressionMapper.cs
+++ b/src/AutoMapper/Mappers/ExpressionMapper.cs
@@ -248,7 +248,7 @@ namespace AutoMapper.Mappers
                 throw new AutoMapperMappingException(
                     "Could not determine source property type. Make sure the property is mapped.", 
                     null, 
-                    new TypePair(propertyMap.SourceType, propertyMap.DestinationPropertyType), 
+                    new TypePair(null, propertyMap.DestinationPropertyType), 
                     propertyMap.TypeMap, 
                     propertyMap);
 
--- a/src/AutoMapper/Mappers/ExpressionMapper.cs
+++ b/src/AutoMapper/Mappers/ExpressionMapper.cs
@@ -232,7 +232,7 @@ namespace AutoMapper.Mappers
                 var propertyMap = FindPropertyMapOfExpression(node.Expression as MemberExpression);
                 if (propertyMap == null)
                     return node;
-                var sourceType = propertyMap.SourceMember.GetMemberType();
+                var sourceType = GetSourceType(propertyMap);
                 var destType = propertyMap.DestinationPropertyType;
                 if (sourceType == destType)
                     return MakeMemberAccess(baseExpression, node.Member);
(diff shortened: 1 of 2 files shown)


dnikolovv:
Okay I've added a test. I'm a bit concerned on the class and test name, but feel free to move them around. Is diff broken now?

lbargaoanu:
You can check that yourself, still broken. The test belongs with the other UseAsDataSource tests.

src/AutoMapper/Mappers/ExpressionMapper.cs
```diff
@@ -226,21 +226,31 @@ protected override Expression VisitConstant(ConstantExpression node)
...
+                var newExpression = subVisitor.Visit(node);
+                _destSubTypes = _destSubTypes.Concat(subVisitor._destSubTypes).ToArray();
+                return newExpression;
+            }
+
+            private Type GetSourceType(PropertyMap propertyMap)
+            {
+                if (propertyMap.SourceMember == null)
+                {
+                    return propertyMap.SourceType;
```
lbargaoanu: SourceType can still be null here.

src/AutoMapper/Mappers/ExpressionMapper.cs
```diff
@@ -247,7 +247,13 @@ private Type GetSourceType(PropertyMap propertyMap)
             {
                 if (propertyMap.SourceMember == null)
                 {
-                    return propertyMap.SourceType;
+                    return propertyMap.SourceType ??
```
lbargaoanu: @BlaiseD What to do here?

dnikolovv: For now I made it throw an exception. Not sure if it's the best thing to do.

BlaiseD: @lbargaoanu  I believe @TylerCarlson1 knows this code better - or were you just asking my advice generally?

lbargaoanu: Yeah, you're right. And yes, chip in :)

lbargaoanu: @TylerCarlson1 Thoughts?

BlaiseD: If it's similar to XpressionMapper then either the custom expression or the source member should be available - otherwise how can the objects be mapped i.e. `var sourceType = propertyMap.CustomExpression == null
                    ? propertyMap.SourceMember.GetMemberType()
                    : propertyMap.CustomExpression.ReturnType;`

It does not seem like the code needs the source member except to find the type - may want to follow up with @TylerCarlson1

lbargaoanu: It can be mapped through a resolver.

BlaiseD: you're right - XpressionMapper throws an exception when propertyMap.ValueResolverConfig != null.  I would do something similar - that's my two cents :)

TylerCarlson1: I would think `var sourceType = propertyMap.SourceType;` would be sufficient.  If not similar to what @BlaiseD has with for source type but include CustomResolver as it's a LambdaExpression as well.

`PropertyMap.SourceType` has conditions for all types of resolvers types, so that should be used.  If it's null it will return null, but it shouldn't happen unlike what @lbargaoanu says.

The problem is it's a custom expression and `SourceMember.GetMemberType()` doesn't account for custom expressions.  I would think CustomResolver could work as well because it's a lambda but might have to be translated.

Might suggest what @BlaiseD says for setting sourceType for now and can latter test Custom Resolvers and ValueResolverConfig to see if they are supported in UseAsDataSource.

lbargaoanu: When it's mapped through a resolver it returns typeof(object) (I think I put that one in there :)). Probably it can be null only when it's not mapped.

TylerCarlson1: Yea I'm not 100% sure what would happen, but it might fail somewhere else.  Not sure if UseAsDataSource would work with a value resolver, but I would guess it wouldn't since it's similar to projection in what it does.  Well even if it did projection wouldn't support it anyways :)

PR #2299, opened 2017-08-29
Throwing InvalidOperationException when mapping for property is missed during map expression

VadimGatsura:
Added null checking of propertyMap in XpressionMapperVisitor


src/AutoMapper/XpressionMapper/XpressionMapperVisitor.cs
```diff
@@ -233,10 +233,15 @@ protected void FindDestinationFullName(Type typeSource, Type typeDestination, st
                 throw QueryMapperHelper.MissingMapException(sourceType: typeDestination, destinationType: typeSource);
             }
                 
-            
             if (sourceFullName.IndexOf(period, StringComparison.OrdinalIgnoreCase) < 0)
             {
                 var propertyMap = typeMap.GetPropertyMaps().SingleOrDefault(item => item.DestinationProperty.Name == sourceFullName);
+                if (propertyMap == null)
+                {
+                    throw new InvalidOperationException(
```
lbargaoanu: We already have a [helper](https://github.com/AutoMapper/AutoMapper/search?utf8=%E2%9C%93&q=MissingMapException&type=) for that. But I'm not sure whether that's the right message. Maybe it cannot be mapped because of a name mismatch for example. cc @BlaiseD

VadimGatsura: I know about helper, but in this case type mapping isn't missed. In this case property map is missed.
As for the message, I got it from GetPropertyMap method in [QueryMapperHelper](https://github.com/AutoMapper/AutoMapper/blob/34a237713dbd2d6fe06c42fc74e65e0c934820e0/src/AutoMapper/QueryableExtensions/Impl/QueryMapperHelper.cs) class.

lbargaoanu: Yes, that's what I was implying :) So extract that to a similar helper and use that. And also change the message. Look at your test, the map exists, right? :) Maybe usually a missing map is the issue, so change the wording.

BlaiseD: @lbargaoanu the wording does seem consistent with QueryMapperHelper.GetPropertyMap. What am I missing? I agree it's better in a helper.  @VadimGatsura if you could also use the helper to check for the property map in the else clause?
```
            if (sourceFullName.IndexOf(period, StringComparison.OrdinalIgnoreCase) < 0)
            {
                var propertyMap = typeMapOrConfigurationProviderHelper.GetPropertyMap();

                //missing code
            }
            else
            {
                var propertyName = sourceFullName.Substring(0, sourceFullName.IndexOf(period, StringComparison.OrdinalIgnoreCase));
                var propertyMap = typeMapOrConfigurationProviderHelper.GetPropertyMap();
                //missing code
            }
```

lbargaoanu: Yes, and I'm saying it's wrong in both cases :) (unless I'm missing smth). The property map missing means that property is not mapped. Usually because the the map between the respective types is missing, other times because the names don't match. So the wording should reflect that. Don't just say create the map, because the map clearly is there.

BlaiseD: Clear - some more detail  like **"Add a property map for Thing.Foo from a ThingModel member using Mapper.CreateMap<ThingModel, Thing>.FromMember()."**

lbargaoanu: Well, that's too specific, I think :) But smth helpful. There must be such a message somewhere in AM.

VadimGatsura:
I updated my PR. Could you look again, please?

src/AutoMapper/QueryableExtensions/Impl/QueryMapperHelper.cs
```diff
@@ -34,5 +34,9 @@ public static TypeMap CheckIfMapExists(this IConfigurationProvider config, Type
 
         public static Exception MissingMapException(Type sourceType, Type destinationType) 
             => new InvalidOperationException($"Missing map from {sourceType} to {destinationType}. Create using Mapper.CreateMap<{sourceType.Name}, {destinationType.Name}>.");
+
+        public static Exception MissingPropertyMapException(string sourceFullName, Type sourceType, Type destinationType)
```
lbargaoanu: The idea was to use the same helper above in GetPropertyMap.  And the message is misleading. I say throw AutoMapperConfigurationException. That seems to have the right message.

VadimGatsura:
Sorry for delayed changes, but I believe it is last commit for the PR.

VadimGatsura:
Close due to having another PR #2323

PR #2301, opened 2017-08-30
Remove ifdefs

lbargaoanu:
I guess it could be IEnumerable, but it's no biggie.


jbogard:
Meh, it could I guess. This is fine, we don't exactly enforce immutability for these config objects.

PR #2323, opened 2017-09-16
Suggestions for PR# 2299 Throwing InvalidOperationException when mapp…

BlaiseD:
@VadimGatsura @lbargaoanu 

This PR is to help resolve #2299.  I think we're stuck on some questions regarding best practices.  If we agree then please feel free to move the code into #2299.
Closes #2299.

### DIFF
--- a/src/AutoMapper/QueryableExtensions/Impl/QueryMapperHelper.cs
+++ b/src/AutoMapper/QueryableExtensions/Impl/QueryMapperHelper.cs
@@ -15,7 +15,7 @@ namespace AutoMapper.QueryableExtensions.Impl
                                       pm.SourceMember != null && pm.SourceMember.Name == sourceMemberInfo.Name);
 
             if (propertyMap == null)
-                throw new AutoMapperConfigurationException(new AutoMapperConfigurationException.TypeMapConfigErrors[] { new AutoMapperConfigurationException.TypeMapConfigErrors(typeMap, new string[] { sourceMemberInfo.Name }, true) });
+                throw PropertyConfigurationException(typeMap, new string[] { sourceMemberInfo.Name });
 
             return propertyMap;
         }
@@ -24,7 +24,7 @@ namespace AutoMapper.QueryableExtensions.Impl
         {
             var propertyMap = typeMap.GetPropertyMaps().SingleOrDefault(item => item.DestinationProperty.Name == destinationPropertyName);
             if (propertyMap == null)
-                throw new AutoMapperConfigurationException(new AutoMapperConfigurationException.TypeMapConfigErrors[] { new AutoMapperConfigurationException.TypeMapConfigErrors(typeMap, new string[] { destinationPropertyName }, true) });
+                throw PropertyConfigurationException(typeMap, new string[] { destinationPropertyName });
 
             return propertyMap;
         }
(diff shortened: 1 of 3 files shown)


src/AutoMapper/QueryableExtensions/Impl/QueryMapperHelper.cs
```diff
@@ -15,10 +15,17 @@ public static PropertyMap GetPropertyMap(this IConfigurationProvider config, Mem
                                       pm.SourceMember != null && pm.SourceMember.Name == sourceMemberInfo.Name);
 
             if (propertyMap == null)
-            {
-                var message = $"Missing property map from {sourceMemberInfo.DeclaringType.Name} to {destinationMemberType.Name} for {sourceMemberInfo.Name} property. Create using Mapper.CreateMap<{sourceMemberInfo.DeclaringType.Name}, {destinationMemberType.Name}>.";
-                throw new InvalidOperationException(message);
-            }
+                throw new AutoMapperConfigurationException(new AutoMapperConfigurationException.TypeMapConfigErrors[] { new AutoMapperConfigurationException.TypeMapConfigErrors(typeMap, new string[] { sourceMemberInfo.Name }, true) });
```
lbargaoanu: I think the constructor call needs to be reused below.

src/AutoMapper/XpressionMapper/XpressionMapperVisitor.cs
```diff
@@ -225,15 +224,10 @@ protected void FindDestinationFullName(Type typeSource, Type typeDestination, st
...
-
-            if (typeMap == null)
-            {
-                throw QueryMapperHelper.MissingMapException(sourceType: typeDestination, destinationType: typeSource);
-            }
+            var typeMap = ConfigurationProvider.CheckIfMapExists(sourceType: typeDestination, destinationType: typeSource);//The destination becomes the source because to map a source expression to a destination expression,
+            //we need the expressions used to create the source from the destination 
 
-            PathMap pathMap = typeMap.FindPathMapByDestinationPath(sourceFullName);
+            PathMap pathMap = typeMap.FindPathMapByDestinationPath(destinationFullPath: sourceFullName);
```
lbargaoanu: I don't think specifying the parameter name is helpful here (and below). The function name already makes the intention clear.

src/AutoMapper/QueryableExtensions/Impl/QueryMapperHelper.cs
BlaiseD: which constructor call? below?????.  Please be clear.

src/AutoMapper/XpressionMapper/XpressionMapperVisitor.cs
BlaiseD: ok - we can take it out.

Having said that it does make it clear that using the sourceFullName is not an error.

src/AutoMapper/QueryableExtensions/Impl/QueryMapperHelper.cs
lbargaoanu: The same code appears [here](https://github.com/AutoMapper/AutoMapper/pull/2323/files#diff-31f5e95b00518c1126cd29aef4b975f3R27).

BlaiseD: Done

BlaiseD:
Maybe you want to proceed with this PR instead of #2299 - once you're happy with it - no response yet from @VadimGatsura

Thanks

Issue #2347, opened 2017-10-02
Upgrading from 2.1.267 to 6.1.1 KeyValuePair to DTO error


cblaze22:
For some reason the code below wont work while upgrading from 2.1.267 to 6.1.1.  I get this error below?

```

            Mapper.Initialize(cfg =>
            {
              
                cfg.CreateMap<KeyValuePair<string, object>, MetaModel>()
                    .ForMember(dest => dest.Id, opt => opt.Ignore())
                    .ForMember(dest => dest.Type, opt => opt.Ignore())
                    .ForMember(dest => dest.Required, opt => opt.Ignore())
                    .ForMember(dest => dest.Hidden, opt => opt.Ignore())
                    .ForMember(dest => dest.Name, opt => opt.MapFrom(src => src.Key))
                    .ForMember(dest => dest.Value, opt => opt.MapFrom(src => src.Value));
```

```
Expression of type 'System.Collections.Generic.KeyValuePair`2[System.String,System.Object]' cannot be used for parameter of type 'System.Object' of method 'Void CacheDestination(System.Object, System.Type, System.Object)'

[ArgumentException: Expression of type 'System.Collections.Generic.KeyValuePair`2[System.String,System.Object]' cannot be used for parameter of type 'System.Object' of method 'Void CacheDestination(System.Object, System.Type, System.Object)']
   System.Linq.Expressions.Expression.ValidateOneArgument(MethodBase method, ExpressionType nodeKind, Expression arg, ParameterInfo pi) +314
   System.Linq.Expressions.Expression.Call(Expression instance, MethodInfo method, Expression arg0, Expression arg1, Expression arg2) +102
   AutoMapper.Execution.TypeMapPlanBuilder.CreateDestinationFunc(Boolean& constructorMapping) +229
   AutoMapper.Execution.TypeMapPlanBuilder.CreateMapperLambda(HashSet`1 visitedTypeMaps) +315
   AutoMapper.TypeMap.Seal(IConfigurationProvider configurationProvider, HashSet`1 visitedTypeMaps) +366
   AutoMapper.MapperConfiguration.Seal() +1569
   AutoMapper.MapperConfiguration..ctor(MapperConfigurationExpression configurationExpression) +743
   AutoMapper.Mapper.Initialize(Action`1 config) +36
   Tournaments.Models.RegisterMaps.Register() in C:\Development\Exposure\AAUIntegration\Projects\Tournaments.Models\RegisterMaps.cs:89
[...]

PR #2348, opened 2017-10-03
Preserve references doesn't make sense for value types

lbargaoanu:
Obviously the test is contrived :) (inspired by #2347) The alternative is to wait for it to happen in real code.


lbargaoanu:
This seems stuck for some reason.

jbogard:
Webhook timed out looks like.

PR #2350, opened 2017-10-04
Reuse the compiled expression

lbargaoanu:
Clearly you lose something by calling a delegate. It's not much from what I can see. But you avoid compiling things twice. I initially thought the typed version would be used more often. But I don't know. It seems most people call ```Map<Destination>(object)```, so the Lazy doesn't help and we end up compiling twice anyway.

### DIFF
--- a/src/AutoMapper/MapperConfiguration.cs
+++ b/src/AutoMapper/MapperConfiguration.cs
@@ -365,19 +365,17 @@ namespace AutoMapper
 
         internal struct MapperFuncs
         {
-            private readonly Lazy<UntypedMapperFunc> _untyped;
-
             public Delegate Typed { get; }
 
-            public UntypedMapperFunc Untyped => _untyped.Value;
+            public UntypedMapperFunc Untyped { get; }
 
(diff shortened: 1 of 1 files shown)


lbargaoanu:
I'm not sure why, but it seems even the mapping itself is slightly faster (maybe because there's no Lazy anymore). For compiling, the gain is around 30%. I used [this](https://github.com/AutoMapper/AutoMapper/blob/master/src/UnitTests/Bug/MapAtRuntime/MapAtRuntime.cs).

lbargaoanu:
Thoughts? :)

jbogard:
Yeah works for me.

I was also curious if the FastExpressionCompiler project would make much of a difference too https://github.com/dadhi/FastExpressionCompiler

lbargaoanu:
I don't know. That stuff about the hosted appdomain should be solved by applying APTCA. About compiling, in the worst cases, the JIT took longer than compiling expressions. And I'm not sure we want to depend on that package. So maybe it doesn't matter much.

PR #2358, opened 2017-10-09
Migrate documentation from github to readthedocs

### LINKED ISSUE #1982: Move all documentation from the wiki to readthedocs.io
No one reads the wiki and it's hard to organize.

stehlikio:
I took a stab a moving all of the docs from the wiki over to read the docs (which is built from master). 

You can see the docs in action [here](http://automapperdocs.readthedocs.io/en/latest/). When merged, as long as everything is configured correctly, the docs should be built and available in the project @jbogard setup on [Read the Docs](https://readthedocs.org/projects/automapper/).

I had to make a few tweaks to the markup to make it work with Read the Docs, but most of the files received only small tweaks (I'll admit...I made some whitespace changes). Down the road, it might be wise to look into migrating to reStructuredText files over markdown (here's a [blog post](http://ericholscher.com/blog/2016/mar/15/dont-use-markdown-for-technical-docs/) about that).

Closes #1982


jbogard:
Looks great to me!

jbogard:
I think the readme.md needs to change to point to this, right?

stehlikio:
Good catch, didn't even look for that. I also noticed two source files with references in the comments. I can get those in a separate commit.

jbogard:
Hmmm the built docs don't look quite right. Any ideas? http://automapper.readthedocs.io/en/latest/

stehlikio:
Interesting, it could possibly be the way that you have the readthedocs project configured. In the project if you go to Admin > Settings, is `Sphinx Html` turned on? I think you might have it set to `Mkdocs (Markdown)`.

image

PR #2360, opened 2017-10-11
Update references to wiki in comments

stehlikio:
Two source files were referencing the "old" github wiki in the comments.


stehlikio:
Also, side note for @jbogard. 

The settings for the readthedocs may set to default to the `stable` docs if you visit [automapper.readthedocs.io](http://automapper.readthedocs.io), and until the next release of AutoMapper, those docs won't get built (since it looks at release tags to determine `stable`). It might be beneficial to set readthedocs to default to `latest` for a while.

screenshot 2017-10-10 22 40 36

PR #2370, opened 2017-10-21
ResolutionContext SetDestinationValue uses indexer for adding destination in cache

### LINKED ISSUE #2363: ArgumentException: An item with the same key has already been added.
[Gist https://gist.github.com/SteveHobs/fd6ecfeabfe7ca0b116b4a8df27d2fc3](https://gist.github.com/SteveHobs/fd6ecfeabfe7ca0b116b4a8df27d2fc3)

### Source/destination types

```csharp
    public class ParentDTO
    {
        public List<ChildDTO> Children { get; set; } = new List<ChildDTO>();
        public int Id { get; set; }
    }

    public class ChildDTO
    {
        public int Id { get; set; }
        public ParentDTO Parent { get; set; }
    }

    public class ParentModel
    {
        public List<ChildModel> Children { get; set; } = new List<ChildModel>();
        public int Id { get; set; }
    }

    public class ChildModel
    {
        public ChildModel(ParentModel parent) // This causes the exception.
        {
            Parent = parent;
        }

        public int Id { get; set; }
        public ParentModel Parent { get; set; }
    }
```

### Mapping configuration

```csharp
             Mapper.Initialize(
                cfg =>
                {
                    cfg.CreateMap<ParentDTO, ParentModel>()
                        .PreserveReferences();

                    cfg.CreateMap<ChildDTO, ChildModel>()
                        .ForMember(c => c.Parent, o => o.Ignore())
                        .PreserveReferences();
                });

```

### Actual behavior

Throws an exception when using a constructor on the ChildModel class.

### Steps to reproduce

```csharp
var parentDto = new ParentDTO {Id = 1};
            for (var i = 0; i < 5; i++)
            {
                parentDto.Children.Add(new ChildDTO {Id = i, Parent = parentDto});
            }

            var children = Mapper.Map<List<ChildDTO>, List<ChildModel>>(parentDto.Children);
```

g-un--:
Fixes #2363

### DIFF
--- a/src/AutoMapper/ResolutionContext.cs
+++ b/src/AutoMapper/ResolutionContext.cs
@@ -24,7 +24,7 @@ namespace AutoMapper
 
         internal void CacheDestination(object source, Type destinationType, object destination)
         {
-            InstanceCache.Add(new ContextCacheKey(source, destinationType), destination);
+            InstanceCache[new ContextCacheKey(source, destinationType)] = destination;
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


lbargaoanu:
I was wrong. I guess we cannot solve this without changing the execution plan.

Issue #2317, opened 2017-09-13
Dynamic adding of mappings


siberianguy:
For years I've been hesitant about using AutoMapper because of its indirect nature: somewhere in one big file there're all the mappings and it's not clear what's going on when you read the method that uses Map. Introducing Profiles improved this a lot as now we can at least group mappings into groups.

Recently I've been looking into Jimmy's [ContosoUniversityCore project](https://github.com/jbogard/ContosoUniversityCore) and realised mappings are not getting reused as for every action there's one command/query and one handler. So it makes no sense to store mappings together in a Profile file common for the whole feature. Mapping belongs to the file with command/query and handler. Jimmy seems to agree with this: https://github.com/jbogard/ContosoUniversityCore/issues/19

But what is the best way to implement it? Of course we could add a Profile class right next to Command/Query and Handler but it seems to be an overkill for simple scenarios:
```
public class MappingProfile : Profile
    {
        public MappingProfile()
        {
            CreateMap<Course, Index.Result.Course>();
        }
    }
```
So 6 decorative lines just to add a one-line mapper.

What might be a better solution to this specific problem is adding inline mappers.
```
       public class Handler : IRequestHandler<Command>
        {
            private readonly SchoolContext _db;

            public Handler(SchoolContext db)
            {
                _db = db;
            }

            public void Handle(Command message)
            {
                **Mapper.CreateMap<Course, Index.Result.Course>();** //the only new line
                var course = Mapper.Map<Command, Course>(message);
                course.Id = message.Number;

                _db.Courses.Add(course);
            }
        }
```
In the example above I just took [this handler](https://github.com/jbogard/ContosoUniversityCore/blob/master/src/ContosoUniversityCore/Features/Course/Create.cs) and added one line that CreatesMap.

This approach (specific syntax is debatable) would solve the indirection that bothered me about AutoMapper for years. Here's the mapping definition, right where it gets used. It's as clear as writing manual mappings but you don't have to write repetitive code: the best from both worlds.

I assume different kinds of inline mappings were discussed multiple times during AutoMapper development and there should be good reasons why it hasn't been implemented (or am I missing something?). Implementations thoughts: CreateMap should check if this mapping has already added and don't add it again, so there should not be a huge overhead; Mapper.CreateMap doesn't have to be static: we could inject mapper in some form.

lbargaoanu:
[...]

PR #2379, opened 2017-11-02
Inline config and auto-auto-mapping

jbogard:
So far moved the dynamic mapping out of conventions and into a last-resort-mapping step. Conventions had their issues of being in a weird place in the mapping pipeline, this moves it to a more last-resort step.

### DIFF
--- a/docs/Inline-Mapping.md
+++ b/docs/Inline-Mapping.md
@@ -1,0 +1,66 @@
+# Inline Mapping
+
+AutoMapper creates type maps on the fly (new in 6.2.0). When you call `Mapper.Map` for the first time, AutoMapper will create the type map configuration and compile the mapping plan. Subsequent mapping calls will use the compiled map.
+
+## Inline configuration
+
+To configure an inline map, use the mapping options:
+
+```c#
+var source = new Source();
+
+var dest = Mapper.Map<Source, Dest>(source, opt => opt.ConfigureMap().ForMember(dest => dest.Value, m => m.MapFrom(src => src.Value + 10)));
+```
+
+You can use local functions to make the configuration a little easier to read:
+
+```c#
+var source = new Source();
+
+void ConfigureMap(IMappingOperationOptions<Source, Dest> opt) {
+    opt.ConfigureMap()
+       .ForMember(dest => dest.Value, m => m.MapFrom(src => src.Value + 10))
+};
+
+var dest = Mapper.Map<Source, Dest>(source, ConfigureMap);
+```
+
+You can use closures in this inline map as well to capture and use runtime values in your configuration:
+
+```c#
+int valueToAdd = 10;
+var source = new Source();
+
+void ConfigureMap(IMappingOperationOptions<Source, Dest> opt) {
+    opt.ConfigureMap()
+       .ForMember(dest => dest.Value, m => m.MapFrom(src => src.Value + valueToAdd))
+};
+
+var dest = Mapper.Map<Source, Dest>(source, ConfigureMap);
+```
+
+## Inline validation
+
+The first time the map is used, AutoMapper validates the map using the default validation configuration (destination members must all be mapped). Subsequent map calls skip mapping validation. This ensures you can safely map your objects.
+
+You can configure the member list used to validate, to validate the source, destination, or no members to validate per map:
+
+```c#
+var source = new Source();
+
+var dest = Mapper.Map<Source, Dest>(source, opt => opt.ConfigureMap(MemberList.None);
+```
+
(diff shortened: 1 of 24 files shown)


jbogard:
I didn’t move conventions, just dynamic mapping being a convention.

jbogard:
Fixes #2317

mat-mcloughlin:
when you don't have enough auto

hikalkan:
@jbogard as I understand there is only one setting (named ``CreateMissingTypeMaps ``) can be enabled/disabled to control both of inline mapping (on the fly mapping with provided mapping configuration) and dynamic mapping (no mapping configuration provided, we just try to execute a mapping that we haven't created before).

If so, this PR enables dynamic mapping by default, which is not considered as a good practice before. Why didn't you create two different setting for inline and dynamic mappings? Now do you think dynamic mapping is acceptable (or considered as inline mapping without custom configuration)?

andreasnilsen:
> this PR enables dynamic mapping by default, which is not considered as a good practice before. Why didn't you create two different setting for inline and dynamic mappings? Now do you think dynamic mapping is acceptable (or considered as inline mapping without custom configuration)?

I'm left wondering the same, after upgrading to AutoMapper 6.2.x, AssertConfigurationIsValid doesn't throw on unmapped properties/types any more, was it intentional to change the defaults of automatically creating dynamic mapping?

jbogard:
It throws on unmapped properties, not unmapped types..

andreasnilsen:
> It throws on unmapped properties, not unmapped types..

Ok, so AssertConfigurationIsValid will only throw on unmapped properties of the explicitly configured types, and not on unmapped properties of type maps that will be dynamically created, even when those types are used for the properties of the explicitly mapped types. Those errors will only surface runtime when doing the actual mapping.

I.e. it is expected that the last assert here fails :
``` csharp
using AutoMapper;
using Xunit;

namespace Automapping.Tests
{
    public class UnmappedPropertiesTests
    {
        public class Source
        {
            public SourceItem Item { get; set; }
        }

        public class SourceItem
        {
            public string SomeProperty { get; set; }
        }

        public class Destination
        {
            public DestinationItem Item { get; set; }
        }

        public class DestinationItem
        {
            public string SomeDifferentlyNamedProperty { get; set; }
        }

        [Fact]
        public void AssertConfigurationIsValid_Should_Throw_On_Unmapped_Properties_Of_Dynamic_Type_Maps()
        {
            //Top level types explictly mapped, the rest of the graph gets dynamically created type maps
            var mapperConfiguration = new MapperConfiguration(cfg => cfg.CreateMap<Source, Destination>());
            var mapper = mapperConfiguration.CreateMapper();

            //OK
            var runtimeMappingException = Assert.ThrowsAny<AutoMapperMappingException>(() =>
                mapper.Map<Destination>(new Source {Item = new SourceItem {SomeProperty = "abc"}}));
            //OK
            Assert.IsAssignableFrom<AutoMapperConfigurationException>(runtimeMappingException.InnerException);
            //Fails
            Assert.ThrowsAny<AutoMapperConfigurationException>(() =>
                mapper.ConfigurationProvider.AssertConfigurationIsValid());
        }
    }
}
```

The reason I ask Is because looking through our code base, it seems we have developed the anti-pattern of not specifying type maps anymore for new types introduced after we upgraded to AutoMapper 6.2+, which leads to AssertConfigurationIsValid giving a false sense of security when modifiying the mapped types.

lbargaoanu:
Maybe just set CreateMissingTypeMaps to false? :) [The docs](http://docs.automapper.org/en/latest/Inline-Mapping.html).

andreasnilsen:
Yes I am aware, I was just surprised that the defaults were changed "silently" and of the implications this has for not being able to trust AssertConfigurationIsValid.

It would be better if the dynamic maps for the missing type maps for properties of the explicitly configured types were created/compiled when invoking AssertConfigurationIsValid IMHO.

lbargaoanu:
No, that was fixed and things work according to the docs. Anything else you need, a PR is welcome.

andreasnilsen:
You mean fixed in the sense that the mapping will throw runtime, then yes, but the AssertConfigurationIsValid still doesn't throw in a repro of that issue, which is contrary to the expected behaviour described in the issue.

But ok, I take the hint, if I want changes to this, I need to implement them myself and send a PR :)

jbogard:
Perhaps one change could be if you’ve configured a map but it contains an
unmapped child, then we should assert the child?

lbargaoanu:
I think it could be done the same as for [the closed generic maps](https://github.com/AutoMapper/AutoMapper/blob/fd596f445a072a2bb8a9cfb7ae9b751cf62e344c/src/AutoMapper/MapperConfiguration.cs#L221).

PR #2400, opened 2017-11-14
added IgnoreMembers implementation

szalapski:
Added IgnoreMembers.  

I've found a version of this new method to be very helpful to simplify and shorten my mapping configuration code, and to avoid repeating myself.  Would you like to make it part of the API?  An example is shown in the XML comments on IMappingExpression.

### DIFF
--- a/src/AutoMapper/Configuration/MappingExpression.cs
+++ b/src/AutoMapper/Configuration/MappingExpression.cs
@@ -51,6 +51,9 @@ namespace AutoMapper.Configuration
         public new IMappingExpression IgnoreAllSourcePropertiesWithAnInaccessibleSetter() 
             => (IMappingExpression)base.IgnoreAllSourcePropertiesWithAnInaccessibleSetter();
 
+        public new IMappingExpression IgnoreMembers(params Expression<Func<object, object>>[] memberExpressionTrees)
+            => (IMappingExpression)base.IgnoreMembers(memberExpressionTrees);
+
         public new IMappingExpression IncludeBase(Type sourceBase, Type destinationBase) 
             => (IMappingExpression)base.IncludeBase(sourceBase, destinationBase);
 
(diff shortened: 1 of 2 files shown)


src/AutoMapper/IMappingExpression.cs
```diff
@@ -252,6 +260,21 @@ public interface IMappingExpression<TSource, TDestination>
...
+        /// <returns>Itself</returns>
+        /// <remarks>This method is syntactic sugar for calling .ForMember and .Ignore repeatedly.</remarks>
+        /// <example>
+        /// To ignore several properties when mapping, specify each in a lambda expression.
+        /// <code>
+        /// config.CreateMap&lt;Foo, FooDto&gt;()
+        ///       .IgnoreMembers(target => target.Number, target => target.Name, target => target.CreateDate);
+        /// </code> 
+        /// </example> 
+        IMappingExpression<TSource, TDestination> IgnoreMembers(params Expression<Func<TDestination, object>>[] memberExpressionTrees);
```
lbargaoanu: I think IgnoreDestinationMembers is a better name, as you can also ignore source members.

src/AutoMapper/Configuration/MappingExpression.cs
```diff
@@ -245,6 +248,14 @@ public void ForAllMembers(Action<IMemberConfigurationExpression<TSource, TDestin
             return this;
         }
 
+        public IMappingExpression<TSource, TDestination> IgnoreMembers(
+            params Expression<Func<TDestination, object>>[] memberExpressionTrees
+        ) =>
+            memberExpressionTrees.Length < 1
+                ? this
+                : ForMember(memberExpressionTrees[0], options => options.Ignore())
+                    .IgnoreMembers(memberExpressionTrees.Skip(1).ToArray());
```
lbargaoanu: There is no need for recursion here, a foreach would do the job.

lbargaoanu:
I believe this kind of thing belongs in an extension method, but @jbogard might disagree.

src/AutoMapper/IMappingExpression.cs
jbogard: +1

jbogard:
If we made it an extension method, I'd want to convert the other similar ones to extension methods too.

lbargaoanu:
It makes sense.

szalapski:
If you think you will likely accept these changes into the product, I'd be happy to rename it and add some unit tests.

jbogard:
That’s generally the purpose of these reviews. We’d say if we don’t like
the idea :)

lbargaoanu:
A new PR is welcome.

PR #2403, opened 2017-11-15
Moving inline map creation back to inside type map resolving

### LINKED ISSUE #2394: Inline maps broke missing type map creation for ProjectTo.
Moving missing type map creation out of a convention broke ProjectTo (and likely some other scenarios).

jbogard:
Fixes #2394

### DIFF
--- a/src/AutoMapper/Configuration/Conventions/SourceToDestinationNameMapperAttributesMember.cs
+++ b/src/AutoMapper/Configuration/Conventions/SourceToDestinationNameMapperAttributesMember.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
 
@@ -6,13 +7,16 @@ namespace AutoMapper.Configuration.Conventions
 {
     public class SourceToDestinationNameMapperAttributesMember : ISourceToDestinationNameMapper
     {
+        private static readonly SourceMember[] Empty = new SourceMember[0];
+        private readonly Dictionary<TypeDetails, SourceMember[]> _allSourceMembers = new Dictionary<TypeDetails, SourceMember[]>();
+
         public MemberInfo GetMatchingMemberInfo(IGetTypeInfoMembers getTypeInfoMembers, TypeDetails typeInfo, Type destType, Type destMemberType, string nameToSearch)
         {
-            var sourceMembers = getTypeInfoMembers.GetMemberInfos(typeInfo)
-                .Select(sourceMember => new SourceMember(sourceMember))
-                .Where(s => s.Attribute != null)
-                .ToArray();
-
+            if (!_allSourceMembers.TryGetValue(typeInfo, out SourceMember[] sourceMembers))
+            {
+                sourceMembers = getTypeInfoMembers.GetMemberInfos(typeInfo).Select(sourceMember => new SourceMember(sourceMember)).Where(s => s.Attribute != null).ToArray();
+                _allSourceMembers[typeInfo] = sourceMembers.Length == 0 ? Empty : sourceMembers;
+            }
             return sourceMembers.FirstOrDefault(d => d.Attribute.IsMatch(typeInfo, d.Member, destType, destMemberType, nameToSearch)).Member;
         }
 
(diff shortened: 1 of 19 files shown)


src/AutoMapper/Configuration/Conventions/SourceToDestinationNameMapperAttributesMember.cs
```diff
@@ -8,15 +7,18 @@ namespace AutoMapper.Configuration.Conventions
     public class SourceToDestinationNameMapperAttributesMember : ISourceToDestinationNameMapper
     {
         private static readonly SourceMember[] Empty = new SourceMember[0];
-        private readonly Dictionary<TypeDetails, SourceMember[]> _allSourceMembers = new Dictionary<TypeDetails, SourceMember[]>();
+        private LockingConcurrentDictionary<TypeDetails, SourceMember[]> _allSourceMembers
+            = new LockingConcurrentDictionary<TypeDetails, SourceMember[]>(_ => Empty);
```
lbargaoanu: I don't think we should lock here. It's difficult to write correct thread safe code in a lot of places. I think we should lock when calling CreateMap.

src/AutoMapper/MapperConfiguration.cs
```diff
@@ -239,6 +247,16 @@ private TypeMap GetTypeMap(TypePair initialTypes)
...
             }
 
+            if (!hasMapper.Value 
+                && Configuration.CreateMissingTypeMaps
+                && !(initialTypes.SourceType.IsAbstract() && initialTypes.SourceType.IsClass())
+                && !(initialTypes.DestinationType.IsAbstract() && initialTypes.DestinationType.IsClass())
+                && !ExcludedTypes.Contains(initialTypes.SourceType)
+                && !ExcludedTypes.Contains(initialTypes.DestinationType))
+            {
+                return Configuration.CreateInlineMap(_typeMapRegistry, initialTypes);
```
lbargaoanu: I would lock(this) here, so all CreateMap calls are serialized, closed generics, conventions and inline.

src/AutoMapper/Configuration/Conventions/SourceToDestinationNameMapperAttributesMember.cs
jbogard: We already lock there but this is another place that's a static cache. Another option would be just to get rid of the cache.

src/AutoMapper/MapperConfiguration.cs
jbogard: Sounds good. I got rid of the cache, it's just not obvious.

jbogard: Ah but that's the value factory for the locking concurrent dictionary. The lock is already there.

src/AutoMapper/Configuration/Conventions/SourceToDestinationNameMapperAttributesMember.cs
lbargaoanu: The idea was to not look for the attribute every time it tries to match the members to map. So it will be slower without the cache. But my point point is more general. A coarser grained lock would make the code easier to write and understand. That's why I added a global lock before each CreateMap. So we don't have to worry about concurrency at every point.

src/AutoMapper/MapperConfiguration.cs
lbargaoanu: Yes, I know there's a lock for the type pair, I'm just not sure that's enough. I'll have to think about it some more.

lbargaoanu: I don't think it's enough. A GetType map for some other types might call CreateMap for these initialTypes. And then you have concurrent CreateMap calls again.

lbargaoanu: Another thing, I don't think we need a Lazy here. A regular bool? should do. It's a local, only one thread will access it and the mappers collection is immutable.

jbogard: Ah you're right, I missed the other locks.

jbogard:
Good to go?

lbargaoanu:
It's not a big deal, [but...](https://github.com/AutoMapper/AutoMapper/pull/2403#discussion_r151125531)

src/AutoMapper/MapperConfiguration.cs
jbogard: I use the `hasMapper` in two places though? That would look a little odd, I might need a func or something.

lbargaoanu: I don't have a problem with
     
     var hasMapper = (FindMapper(initialTypes) != null);

Or simply calling it twice as before.

jbogard: Done!

PR #2406, opened 2017-11-16
Adding ability to configure the member list to validate; closes #2402

### LINKED ISSUE #2402: 6.2.0 Broke Conventions-Based Mappings With Missing Target Members
### Version: 6.2.0

This works fine in v6.1.1

### Expected behavior

The mapping happens successfully and `FooViewModel.Extra` is `null`

### Actual behavior

Exception:

```
AutoMapper.AutoMapperConfigurationException : 
Unmapped members were found. Review the types and members below.
Add a custom mapping expression, ignore, add a custom resolver, or modify the source/destination type
For no matching constructor, add a no-arg ctor, add optional arguments, or map all of the constructor parameters
================================================================
Foo -> FooViewModel (Destination member list)
AutoMapper.Bug.Class2+Foo -> AutoMapper.Bug.Class2+FooViewModel (Destination member list)

Unmapped properties:
Extra

```

### Steps to reproduce

```csharp
public class Foo
{
}

public class FooViewModel
{
    public string Extra { get; set; }
}

public class ViewModelConvention : Profile
{
    /// <summary>
    /// Initialises a new instance of the <see cref="ViewModelConvention"/> class.
    /// </summary>
    public ViewModelConvention()
    {
        AddConditionalObjectMapper().Where((source, dest) => dest.Name == source.Name + "ViewModel");
    }
}

[Fact]
public void TestMappingWithMissingTargetProperty()
{
    Mapper.Initialize(cfg => cfg.AddProfile<ViewModelConvention>());

    var result = Mapper.Map<FooViewModel>(new Foo { Bar = "Baz" });

    result.Bar.Should().Be("Baz");
}
```

jbogard:


jbogard:
@lbargaoanu thoughts?

lbargaoanu:
I guess it would make more sense if it were specified somewhere closer to where the map is defined. I guess it doesn't apply to the regular maps.

lbargaoanu:
Took a quick look at the code and it seems it's the same property as for the regular maps, so it fits in IMappingExpression. My first point remains. I guess most people don't care, they don't want validation anyway.

jbogard:
Yeah, I looked at changing the conventions API itself but that'd be a breaking change to `IConditionalObjectMapper` so this seems to be the most straightforward way.

I don't use conventions so I don't really know what would be appropriate there.

TylerCarlson1:
For Conventions I usually don't want validation.  Conventions are a little less lazy then CreateMissingTypeMaps, but it's still pretty lazy.

jbogard:
Why wouldn’t you want validation?

TylerCarlson1:
I am accustomed to the expected case of #2402.  It's simple it just works and I don't have to think about missing property maps.  If they don't match then I didn't want them mapped, and I would make sure I would have the ones I wanted.

PR #2407, opened 2017-11-16
Add Mapper.Reset to reset static mapping

jbogard:


jbogard:
@lbargaoanu thoughts?

lbargaoanu:
Do we really have to? :) I mean you solved the issue with the DI package, right?

jbogard:
Kinda solved it, but I think these are two different issues - you can't reset the static mapping, and the ability to configure the DI extension's behavior.

If you wanted to keep the static behavior but still do `services.AddAutoMapper` then `Reset` would give you that.

lbargaoanu:
Maybe some warning in the XML docs? Not intended for production?

jbogard:
Sounds good.

PR #2421, opened 2017-11-22
Better error message when the creation of the instantiation expression fails

frblondin:
While defining a new mapping between two classes, I was getting a mysterious `Can't resolve this to Queryable Expression` error message. The message was not meaningful enough so I had to debug AutoMapper to understand what the problem was. I simply had to tell AutoMapper which constructor had to be used...

This PR intends to generate a more explanatory error message.

### DIFF
--- a/src/AutoMapper/ConstructorMap.cs
+++ b/src/AutoMapper/ConstructorMap.cs
@@ -43,7 +43,7 @@ namespace AutoMapper
                     ExpressionResultConverters.FirstOrDefault(c => c.CanGetExpressionResolutionResult(result, map));
 
                 result = matchingExpressionConverter?.GetExpressionResolutionResult(result, map)
-                    ?? throw new AutoMapperMappingException($"Unable to generate the instantiation expression for the constructor {Ctor}: no expression could be mapped for constructor parameter '{map.Parameter}'.", null, TypeMap.Types, null, null);
+                    ?? throw new AutoMapperMappingException($"Unable to generate the instantiation expression for the constructor {Ctor}: no expression could be mapped for constructor parameter '{map.Parameter}'.", null, TypeMap.Types);
 
                 return result;
             });
--- a/src/AutoMapper/ConstructorMap.cs
+++ b/src/AutoMapper/ConstructorMap.cs
@@ -42,8 +42,8 @@ namespace AutoMapper
                 var matchingExpressionConverter =
                     ExpressionResultConverters.FirstOrDefault(c => c.CanGetExpressionResolutionResult(result, map));
 
-                result = matchingExpressionConverter?.GetExpressionResolutionResult(result, map) 
-                    ?? throw new Exception("Can't resolve this to Queryable Expression");
+                result = matchingExpressionConverter?.GetExpressionResolutionResult(result, map)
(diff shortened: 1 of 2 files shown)


src/AutoMapper/ConstructorMap.cs
```diff
@@ -35,19 +35,26 @@ public ConstructorMap(ConstructorInfo ctor, TypeMap typeMap)
 
         public Expression NewExpression(Expression instanceParameter)
         {
-            var parameters = CtorParams.Select(map =>
+            try
```
lbargaoanu: I don't think we need the extra try/catch. Some exceptions are really bugs and need no user friendly messages.

src/AutoMapper/QueryableExtensions/BuilderException.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace AutoMapper.QueryableExtensions
+{
+    public class BuilderException : Exception
```
lbargaoanu: I don't think we need a new exception type here. AutoMapperMappingException should do. Just pass it the types so it's easy to understand what's going on.

src/UnitTests/Projection/MoreExplanatoryExceptionTests.cs
```diff
@@ -0,0 +1,32 @@
...
+    public class MoreExplanatoryExceptionTests
+    {
+        [Fact]
+        public void ConstructorWithUnknownParameterTypeThrowsExplicitException()
+        {
+            var config = new MapperConfiguration(cfg =>
+            {
+                cfg.CreateMap<EntitySource, EntityDestination>();
+            });
+            var exception = Assert.Throws<BuilderException>(() => config.ExpressionBuilder.GetMapExpression<EntitySource, EntityDestination>());
```
lbargaoanu: ProjectTo is the API, so the test should call that. GetMapExpression is just an internal thing.

src/AutoMapper/ConstructorMap.cs
frblondin: Right

src/AutoMapper/QueryableExtensions/BuilderException.cs
frblondin: Uses now `AutoMapperMappingException`

src/UnitTests/Projection/MoreExplanatoryExceptionTests.cs
frblondin: `ProjectTo` is now used by unit test.

src/AutoMapper/ConstructorMap.cs
```diff
@@ -42,8 +42,8 @@ public Expression NewExpression(Expression instanceParameter)
                 var matchingExpressionConverter =
                     ExpressionResultConverters.FirstOrDefault(c => c.CanGetExpressionResolutionResult(result, map));
 
-                result = matchingExpressionConverter?.GetExpressionResolutionResult(result, map) 
-                    ?? throw new Exception("Can't resolve this to Queryable Expression");
+                result = matchingExpressionConverter?.GetExpressionResolutionResult(result, map)
+                    ?? throw new AutoMapperMappingException($"Unable to generate the instantiation expression for the constructor {Ctor}: no expression could be mapped for constructor parameter '{map.Parameter}'.", null, TypeMap.Types, null, null);
```
lbargaoanu: The last two nulls are not needed.

PR #2450, opened 2017-12-05
Collection bug

jbogard:
This fixes the newly introduced issue of casting collection types. I didn't think forcing everyone to use `UseDestinationValue` was a good solution, this change centers around the interfaces and the concrete types are only used for instantiation.

### DIFF
--- a/src/AutoMapper/Execution/DelegateFactory.cs
+++ b/src/AutoMapper/Execution/DelegateFactory.cs
@@ -1,11 +1,15 @@
 using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Linq.Expressions;
+using AutoMapper.Configuration;
+using AutoMapper.Mappers.Internal;
 
 namespace AutoMapper.Execution
 {
     using static Expression;
-    using static ExpressionBuilder;
+    using static Internal.ExpressionFactory;
+    using static ElementTypeHelper;
 
     public static class DelegateFactory
     {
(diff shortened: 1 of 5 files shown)


src/AutoMapper/Mappers/Internal/CollectionMapperExpressionFactory.cs
```diff
@@ -46,7 +47,7 @@ public static Expression MapCollectionExpression(IConfigurationProvider configur
                     Assign(passedDestination, destExpression),
                     assignNewExpression,
                     Call(destination, clearMethod),
-                    ToType(mapExpr, createInstance.Type)
```
TylerCarlson1: Stupid question.  Wouldn't this be sufficient not to cast the result back?  Not sure why anything else would be needed.  The problem wasn't that collection was a List when passed down it's cause it was casted as such instead of ICollection.

src/AutoMapper/Execution/DelegateFactory.cs
```diff
@@ -44,6 +48,15 @@ public static Expression GenerateConstructorExpression(Type type)
                 return Constant(null, typeof(string));
             }
 
+            if (type.IsInterface())
```
TylerCarlson1: Is this really necessary?  CollectionMapper returns ICollection it's just casts to List instead of left as instance type.

jbogard: This code is called in a few places so I can't assume what the mappers do.

jbogard:
Here's the before/after execution plan:

https://gist.github.com/jbogard/481811674c348e2b42b958c261d68fa0/revisions#diff-b414850bc101567b83295360411fb383

TylerCarlson1:
Ok looks good enough to me.

Also I don't feel like debugging Expressions atm. :)

src/AutoMapper/Execution/DelegateFactory.cs
lbargaoanu: I guess it's not that bad :) What about ISet? Maybe move [arrays](https://github.com/AutoMapper/AutoMapper/blob/605941b79b1f5cf6663c1876660bb9c8bc338cb4/src/AutoMapper/Execution/ExpressionBuilder.cs#L85) here too?

jbogard: I'll get another PR for that for now lgtm

PR #2451, opened 2017-12-06
Rephrasing some sentences

julienasp:


julienasp:
Hi, i was reading the doc and i had trouble with some of it sentences. So i decided to change them.

I'm planning to read it all this weekend, so you may have a couple more of these later on ^^

docs/Flattening.md
```diff
@@ -64,7 +64,9 @@ public class OrderDto
 }
 ```
 
-When you configure a source/destination type pair in AutoMapper, the configurator attempts to match properties and methods on the source type to properties on the destination type.  If for any property on the destination type a property, method, or a method prefixed with "Get" does not exist on the source type, AutoMapper splits the destination member name into individual words (by PascalCase conventions).
+When you configure a source/destination type pair in AutoMapper, the configurator attempts to match properties and methods from the source type to properties on the destination type. 
+
+When no property, method or method prefixed with "Get" in the source type match the property of the destination type, AutoMapper will splits the destination member name into individual words (by PascalCase conventions).
```
jbogard: splits -> "split"

docs/Flattening.md
```diff
@@ -96,6 +98,6 @@ dto.CustomerName.ShouldEqual("George Costanza");
 dto.Total.ShouldEqual(74.85m);
 ```
 
-We configured the type map in AutoMapper with the CreateMap method.  AutoMapper can only map type pairs it knows about, so we have explicitly register the source/destination type pair with CreateMap.  To perform the mapping, we use the Map method.
+We configured the type map in AutoMapper with the **_CreateMap_** method.  AutoMapper can only map type pairs it knows about, so we have explicitly register the source/destination type pair with CreateMap.  To perform the mapping, we use the **_Map_** method.
```
jbogard: Should do ticks for inline code (`) instead of bolded etc.

bonsailimited: ... "so we have explicitly register" ... -> ... "so we have to explicitly register "...

PR #2469, opened 2017-12-22
Make mapping at runtime closer to inline mapping

### LINKED ISSUE #2466: Mapping from ExpandoObject to existing destination
I use automapper to map dynamic types to entity types. However, when the property type in the dynamic type is empty, the automapper will not override the property value of the entity type. The original value is reserved。Here is my code

``` 
    public class abc 
    {
        public Guid t { get; set; }
        public string a { get; set; }
        public DateTime? TimeTest { get; set; } 
        public DateTime? TimeTest1 { get; set; }
        public DateTime? TimeTest2 { get; set; }
        public DateTime? TimeTest3 { get; set; }
        public int? FSelcritedID { get; set; } 
        public int FParentID { get; set; }
 }

  dynamic d = new ExpandoObject();
            d.FParentID = 99;
            d.FSelcritedID = null;
            d.TimeTest = null;
            d.TimeTest1 = 1513612800000;
            d.TimeTest2 = 1513612800000;
            d.TimeTest3 = 1513612800000;`

      abc testmodel = new abc();
            testmodel.TimeTest = DateTime.Now;
            testmodel.FSelcritedID = 88;
                cfg.CreateMap<long?, DateTime?>().ConvertUsing(new DateTimeNullTypeConverter().Convert);
                cfg.CreateMap<long, DateTime?>().ConvertUsing(new DateTimeTypeConverter().Convert);
                cfg.CreateMap<long, DateTime>().ConvertUsing(new DateTimeTTypeConverter().Convert);
            });
            var map = config.CreateMapper();
            
            map.Map(d, testmodel);
```
After the mapping is complete I want testmodel.FSelcritedID = null, testmodel.TimeTest = null but it has not changed.

Please tell me what to do!

lbargaoanu:
Fixes #2466.
The diff for FillingExistingDestination.cs is mostly whitespace. The new test is at the top.
If I change the default for Inline in PropertyMap, all tests pass, except the one assuming the default is true.

### DIFF
--- a/src/AutoMapper/AutoMapper.csproj
+++ b/src/AutoMapper/AutoMapper.csproj
@@ -17,13 +17,6 @@
     <PackageIconUrl>https://s3.amazonaws.com/automapper/icon.png</PackageIconUrl>
     <PackageProjectUrl>http://automapper.org</PackageProjectUrl>
     <PackageLicenseUrl>https://github.com/AutoMapper/AutoMapper/blob/master/LICENSE.txt</PackageLicenseUrl>
-  </PropertyGroup>
-
-  <PropertyGroup Condition="'$(Configuration)|$(TargetFramework)|$(Platform)'=='Debug|netstandard1.3|AnyCPU'">
-    <LangVersion>latest</LangVersion>
-  </PropertyGroup>
-
-  <PropertyGroup Condition="'$(Configuration)|$(TargetFramework)|$(Platform)'=='Release|netstandard1.3|AnyCPU'">
     <LangVersion>latest</LangVersion>
   </PropertyGroup>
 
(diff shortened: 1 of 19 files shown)


jbogard:
Should we really be passing the `PropertyMap` around, or the correct `ProfileMap` around? It looks like we're only passing the `PropertyMap` to access the `PropertyMap.TypeMap.Profile`. Currently we assume the root `MapperConfiguration` which looks wrong, and it should be using whatever contextual `ProfileMap` based on the enclosed TypeMap.

Also, what would this imply for the scope of profiles? How do we decide when to use the root configuration versus a contextual `ProfileMap`?

lbargaoanu:
I'd say the PropertyMap. We map top level objects differently than members and also the execution plan depends on settings in the property map.
About the scope of profiles, I've just tried to to the same thing as when mapping inline.

lbargaoanu:
What I didn't do here is attempt to determine whether two property maps are equivalent, that seems difficult.

PR #2497, opened 2018-01-15
added Is Initialized method

amir734jj:
I think AutoMapper really needs this method because currently the only way to tests whether if AutoMapper instance is initialized is to try to access Configuration and if we get an exception then it is not initialized.


lbargaoanu:
This check belongs outside AM. Like in a Lazy init. It's not thread safe, right after IsInitialized another thread could call Initialize so the check doesn't help much.

jbogard:
What @lbargaoanu said. Either make this thread-safe, or just in your application code put it in a `Lazy<T>`.

PR #2508, opened 2018-01-24
Simple object mapper : When the destination object is null, it can be useful to know its type

lbargaoanu:
A breaking change, but it seems harmless.


jbogard:
This got turned on finally yesterday :)

lbargaoanu:
I've noticed :)

lbargaoanu:
Somebody wanted also [to handle nulls](https://github.com/AutoMapper/AutoMapper/compare/master...flavourous:allowMapperSourceNull), but I'm not so sure about that. With NullSubstitute and resolvers, I'd say there are other ways.

PR #2515, opened 2018-01-30
Update Custom-value-resolvers.md

jasel-lewis:
Corrections per https://github.com/AutoMapper/AutoMapper/blob/bd3894408533181f842b99d9b5d501a539932b18/src/UnitTests/ContextItems.cs#L61


jbogard:
Now that we're in the .NET Foundation, we need the CLA. @jasel-lewis can you e-sign that?

PR #2518, opened 2018-01-30
Use resolve when looking for included maps

### LINKED ISSUE #2513: Open generics in combination with mapping inheritance do not work
This is probably a very rare case but hit's us with the .net core Identity classes because they reflect just this setup.

When trying to define mapping rules for open generics and re-use them on the concrete implementation using IncludeBase, no errors appear while setting the mapping up, but the defined rules are not applied.

Given the following example:

```csharp
using AutoMapper;

namespace ConsoleApp3
{
    class Program
    {
        {
            Mapper.Initialize(cfg =>
            {
                cfg.CreateMap(typeof(BaseUserDto<>), typeof(BaseUserEntity<>))
                    .ForMember("Id", opt => opt.Ignore());

                cfg.CreateMap(typeof(ConcreteUserDto), typeof(ConcreteUserEntity))
                    .IncludeBase(typeof(BaseUserDto<>), typeof(BaseUserEntity<>));
            });

            var user = new ConcreteUserDto
            {
                Id = "my-id",
                Name = "my-User"
            };

            var userEntity = Mapper.Map<ConcreteUserEntity>(user);
        }
    }

    public abstract class BaseUserDto<TIdType>
    {
        public TIdType Id { get; set; }
    }

    public class ConcreteUserDto : BaseUserDto<string>
    {
    }

    public abstract class BaseUserEntity<TIdType>
    {
        public TIdType Id { get; set; }
    }

    public class ConcreteUserEntity : BaseUserEntity<string>
    {
    }
}
```

The expectation would be, that the resulting `userEntity` object has it's `Id` property set to null, however it is copied over from the DTO.

When changing the example to hardcode all generics to `int`, everything works as expected.

lbargaoanu:
Fixes #2513.
This feels like a bigger change than the issue warrants, but I'm not sure I have a better idea. Another FindOrCreateTypeMap API? :)


jbogard:
Oh not another one of those...

Failed build here

lbargaoanu:
No, it really fails on .Net Core. I'm investigating :)

lbargaoanu:
I guess this would break IncludeBase calls for non existing maps. But it only seems fair :)

PR #2532, opened 2018-02-13
Testing FastExpressionCompiler

jbogard:
Seems slightly slower


lbargaoanu:
But how did you test, with CompileMappings and some large model?

dadhi:
Mmm, code looks Ok. What's VisualStudioProfiler? Can we measure with BDN?

jbogard:
Sure!

dadhi:
Also does it really hit the fast path? Can you try `CompileFast(true)` and check if it returns `null`?

jbogard:
Looks about a 30% improvement in compilation speed

jbogard:
@dadhi on my local machine :P

Issue #2538, opened 2018-02-20
Expression Projection with different types


HRasch:
### Source/destination types

```csharp
// Put your source/destination types here
public class Model
{
	public Guid? Id { get; set; }
	public MyString FullDescription { get; set; }
	public MyString ShortDescription { get; set; }
}

public class Dto
{
	public Guid? Id { get; set; }
	public string FullDescription { get; set; }
	public string ShortDescription { get; set; }
}
```
### Custom Struct type

```csharp
public struct MyString : IComparable, IConvertible
{
	private string value;
	private bool isNull;

	public MyString(string value)
	{
		this.isNull = string.IsNullOrEmpty(value);
		this.value = value ?? string.Empty;
	}

	public string Value
	{
		get
		{
			if (IsNull || value == null)
				return string.Empty;
			return value;
		}
		set
		{
			if (Value == value && !IsNull)
				return;

			if (string.IsNullOrEmpty(value) && !isNull)
			{
				isNull = true;
			}
					
			this.value = value ?? string.Empty;
		}
	}

	public bool IsNull
	{
		get
		{
			return isNull;
		}
		set
		{
			if (isNull == value) return;
			isNull = value;
		}
	}

	public static implicit operator MyString(string a)
	{
		return new MyString(a);
	}

	public static implicit operator string(MyString a)
	{
		return a.Value;
	}

	public static bool operator ==(MyString a, MyString b)
	{
		return a.Value == b.Value;
	}

	public static bool operator ==(MyString a, string b)
	{
		return a.Value == b;
	}

	public static bool operator ==(string a, MyString b)
	{
		return a == b.Value;
	}

	public static bool operator !=(MyString a, MyString b)
	{
		return a.Value != b.Value;
	}

	public static bool operator !=(MyString a, string b)
	{
		return a.Value != b;
	}

	public static bool operator !=(string a, MyString b)
	{
		return a != b.Value;
	}

	public static bool operator <(MyString a, MyString b)
	{
		return a.CompareTo(b) < 0;
	}

	public static bool operator >(MyString a, MyString b)
	{
		return a.CompareTo(b) > 0;
	}

	public static bool operator <=(MyString a, MyString b)
	{
		return a.CompareTo(b) <= 0;
	}

	public static bool operator >=(MyString a, MyString b)
	{
		return a.CompareTo(b) >= 0;
	}

	public override int GetHashCode()
	{
		return this.Value.GetHashCode();
	}

[...]

Issue #1078, opened 2016-02-03
Dropped support for .NET 4.0 with 4.2.0


dittodhole:
With the recent "feature"-release, the support for .NET 4.0 was dropped. This really hurts projects that are stuck on .NET 4.0 (like the one I am working on), and furthermore ... it doesn't follow Semver.

Is there any chance to bring back the support?

jbogard:
No, there isn't.

MS dropped support for .NET 4.0 last month. I also removed Silverlight 4 support. The reflection API completely changed from 4.0 to 4.5, and the shims I had added to support 4.0 were buggy and incorrect.

dittodhole:
Unfortunately depending on Windows XP upwards, even Microsoft dropped support a while ago.
So, I am stuckon 4.1.1 :cry: 

Nevertheless, I would suggest you'd increased the major instead of the minor to reflect such breaking change. Even though the API kinda did not break.

PR #2540, opened 2018-02-21
Added type convertion on expression projection

HRasch:
Expression Projection with different types #2538

### DIFF
--- a/src/AutoMapper/Mappers/ExpressionMapper.cs
+++ b/src/AutoMapper/Mappers/ExpressionMapper.cs
@@ -212,12 +212,8 @@ namespace AutoMapper.Mappers
                 if (propertyMap.CustomExpression != null)
 				{
 					replacedExpression = propertyMap.CustomExpression.ReplaceParameters(replacedExpression);
-
-					if (!node.Type.IsAssignableFrom(replacedExpression.Type))
-					{
-						replacedExpression = ExpressionFactory.ToType(replacedExpression, node.Type);
-					}
-
+					replacedExpression = ExpressionFactory.ToType(replacedExpression, node.Type);
+					
 					return replacedExpression;
 				}
 


HRasch:
Hm, first of all, what do you mean with "not using ToType as discussed."
You just said
> cc @BlaiseD I don't know what's the right fix here, but we have a ToType helper.

so i changed the call of 
`
replacedExpression = Convert(replacedExpression, node.Type);
`
to 

`
replacedExpression = ExpressionFactory.ToType(replacedExpression, node.Type);
`
using the suggested convertion-method. (Imo, it would be nice if this method would be an extensions-method)

Sure, it would be much better if this call would also support the propertyMap, but that's another story.

On the other hand, I'm really struggled running the existing tests of automapper.
[Contributing.md](https://github.com/AutoMapper/AutoMapper/blob/master/CONTRIBUTING.md) said I should use psake.cmd, but even after upgrade my VS2017, installing .NET Core 1.1.6 SDK, having trouble because AutoMapper.UnitTests is definied to target net40, even if [support has been dropped with 4.2.0](https://github.com/AutoMapper/AutoMapper/issues/1078), using different versions of xunit (because of net40 support), xunit.runner.visualstudio (including a 2.3.0-beta3, even if 2.3.1 rtm is released), I'm not able to run the tests with the visual studio test explorer. Even running psake.cmd throws compile-errors on the existing tests: E.g.:
> No coercion operator is defined between types 'UnitTests.Bug.ExpressionMappingWithUseAsDataSource+Model' and 'UnitTests.Bug.ExpressionMappingWithUseAsDataSource+DTO+DTONested'

So, It's not hard for me to write my unit test, but i would feel much better if i could compile and test it :-)

src/AutoMapper/Mappers/ExpressionMapper.cs
```diff
@@ -209,9 +210,18 @@ protected override Expression VisitMember(MemberExpression node)
                     replacedExpression = _parentMappingVisitor.Visit(node.Expression);
 
                 if (propertyMap.CustomExpression != null)
-                    return propertyMap.CustomExpression.ReplaceParameters(replacedExpression);
+				{
+					replacedExpression = propertyMap.CustomExpression.ReplaceParameters(replacedExpression);
 
-                Func<Expression, MemberInfo, Expression> getExpression = MakeMemberAccess;
+					if (!node.Type.IsAssignableFrom(replacedExpression.Type))
```
lbargaoanu: This check is not needed.

lbargaoanu:
Yes, I guess I got distracted by the extra check :) I usually run the tests in VS, but "dotnet test" works too. You can try changing the order of frameworks in csproj. It can be a struggle to make it work, but it's possible. And yes, an extension method would be nice.

HRasch:
Ok, removed the check now. I added it before, because it's exactly the check that MS do in it's MemberAssignment.Expressions.Bind-Method and caused the error before. But the internal check of quality should do (nearly) the same.

lbargaoanu:
Maybe I'm missing smth, but the test runs just fine on master.

PR #2542, opened 2018-02-22
2. Try - Added type convertion on expression projection

HRasch:
Seems my last bugfix breaks some other tests. So i spent my whole day to make this working correctly.

My trouble with the testing-environment ended as soon I upgraded the xunit.runner.visualstudio to version 2.3.1

In addition to the discussed solution, i did the following changes:

ExpressionMapper.ExpressionVisitor implements now additionaly the VisitNew-Method. This was the Method causing trouble, if a projection to an anonymous method with a different type has been queried. In addition to this change, now i call a new GetConvertedArgument-Method to convert the projection-types, if possible, with the registered TypeMaps.

Because the visitor now is called for each transformed argument, the visitmember method get's called for these arguments, and returns a converted expression if the types differ. To prevent the "no coercion operator"-exception, I extended the ReflectionHelper class and added a method that checks if a type can be implicit converted. If not, the original expression is returned, if yes, an convert-expression is returned.

### DIFF
--- a/src/AutoMapper/QueryableExtensions/ExpressionBuilder.cs
+++ b/src/AutoMapper/QueryableExtensions/ExpressionBuilder.cs
@@ -37,10 +37,10 @@ namespace AutoMapper.QueryableExtensions
             new NullableDestinationExpressionBinder(),
             new NullableSourceExpressionBinder(),
             new AssignableExpressionBinder(),
-            new ImplicitAssignableExpressionBinder(),
             new EnumerableExpressionBinder(),
             new MappedTypeExpressionBinder(),
-            new StringExpressionBinder()
+            new StringExpressionBinder(),
+            new ImplicitAssignableExpressionBinder()
         };
 
         private readonly LockingConcurrentDictionary<ExpressionRequest, LambdaExpression[]> _expressionCache;
--- a/src/AutoMapper/QueryableExtensions/ExpressionBuilder.cs
+++ b/src/AutoMapper/QueryableExtensions/ExpressionBuilder.cs
@@ -37,6 +37,7 @@ namespace AutoMapper.QueryableExtensions
             new NullableDestinationExpressionBinder(),
             new NullableSourceExpressionBinder(),
             new AssignableExpressionBinder(),
+            new ImplicitAssignableExpressionBinder(),
             new EnumerableExpressionBinder(),
             new MappedTypeExpressionBinder(),
             new StringExpressionBinder()
--- a/src/AutoMapper/Internal/ReflectionHelper.cs
+++ b/src/AutoMapper/Internal/ReflectionHelper.cs
@@ -187,8 +187,11 @@ namespace AutoMapper.Internal
             var simpleConvertToType = convertTo.IsNullableType() ? convertTo.GetTypeOfNullable() : convertTo;
 
             return simpleFromType.GetDeclaredMethods().Any(x => (x.Name == "op_Explicit" || x.Name == "op_Implicit")
-                && x.ReturnType == simpleConvertToType
-                && x.GetParameters().First().ParameterType == simpleFromType);
+                    && x.ReturnType == simpleConvertToType
+                    && x.GetParameters().First().ParameterType == simpleFromType)
+                || simpleConvertToType.GetDeclaredMethods().Any(x => (x.Name == "op_Explicit" || x.Name == "op_Implicit")
+                    && x.ReturnType == simpleConvertToType
+                    && x.GetParameters().First().ParameterType == simpleFromType);
         }
     }
 }
--- a/src/AutoMapper/Internal/ReflectionHelper.cs
+++ b/src/AutoMapper/Internal/ReflectionHelper.cs
@@ -181,17 +181,36 @@ namespace AutoMapper.Internal
             return targetType;
         }
 
-        public static bool CanImplicitConvert(this Type convertFrom, Type convertTo)
+        public static bool CanImplicitConvert(this Type from, Type to)
         {
-            var simpleFromType = convertFrom.IsNullableType() ? convertFrom.GetTypeOfNullable() : convertFrom;
-            var simpleConvertToType = convertTo.IsNullableType() ? convertTo.GetTypeOfNullable() : convertTo;
+            if (to.IsNullableType())
+            {
+                from = from.IsNullableType() ? Nullable.GetUnderlyingType(from) : from;
+                to = to.IsNullableType() ? Nullable.GetUnderlyingType(to) : to;
+            }
 
-            return simpleFromType.GetDeclaredMethods().Any(x => (x.Name == "op_Explicit" || x.Name == "op_Implicit")
-                    && x.ReturnType == simpleConvertToType
-                    && x.GetParameters().First().ParameterType == simpleFromType)
-                || simpleConvertToType.GetDeclaredMethods().Any(x => (x.Name == "op_Explicit" || x.Name == "op_Implicit")
-                    && x.ReturnType == simpleConvertToType
-                    && x.GetParameters().First().ParameterType == simpleFromType);
+            bool CanConvert(Type type, Type typeToCheck) => valueTypeDict.ContainsKey(type) && valueTypeDict[type].Contains(typeToCheck);
+
+            bool HasImplicitOp(Type type) => type.GetDeclaredMethods().Any(x => x.Name == "op_Implicit" 
+                && x.ReturnType == to 
+                && (x.GetParameters().First().ParameterType == from 
+                    || CanConvert(x.GetParameters().First().ParameterType, from))
+            );
+
+            return to.IsAssignableFrom(from) || HasImplicitOp(from) || HasImplicitOp(to) || CanConvert(to, from);
         }
+
+        static Dictionary<Type, HashSet<Type>> valueTypeDict = new Dictionary<Type, HashSet<Type>>()
+        {
+            { typeof(decimal), new HashSet<Type> { typeof(byte), typeof(sbyte), typeof(char), typeof(short), typeof(ushort), typeof(int), typeof(uint), typeof(long), typeof(ulong) } },
+            { typeof(double), new HashSet<Type> { typeof(byte), typeof(sbyte), typeof(char), typeof(short), typeof(ushort), typeof(int), typeof(uint), typeof(long), typeof(ulong), typeof(float) } },
+            { typeof(float), new HashSet<Type> { typeof(byte), typeof(sbyte), typeof(char), typeof(short), typeof(ushort), typeof(int), typeof(uint), typeof(long), typeof(ulong) } },
+            { typeof(ulong), new HashSet<Type> { typeof(byte), typeof(char), typeof(ushort), typeof(uint) } },
+            { typeof(long), new HashSet<Type> { typeof(byte), typeof(sbyte), typeof(char), typeof(short), typeof(ushort), typeof(int), typeof(uint) } },
+            { typeof(uint), new HashSet<Type> { typeof(byte), typeof(char), typeof(ushort) } },
+            { typeof(int), new HashSet<Type> { typeof(byte), typeof(sbyte), typeof(char), typeof(short), typeof(ushort) } },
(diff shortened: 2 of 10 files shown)


HRasch:
Maybe you should wait with the merge, I found some more cases that don't work (e.g. projecting from string to custom struct.). Also my odata-api is creating some fancy stuff. I will fix them as much as I can and contribute the final result.

HRasch:
Here are my latest infos about the mapping problem. It seems, this kind of conversion has never been testet.

The background for my issue is that i try to create an odata-provider for a custom datasource (based on webapi, odata, restier). Most of the datatypes in my custom datasource are definied with custom struct types, as mentioned. The idea of mapping the odata-query directly to my custom datasource sounds fantastic to me, and I think AutoMapper is the right tool to translate the queries.

This task should not be to complicated, cause Automapper already works perfect, if the source and destination-properties have the same type.

My prototype can currently load data from my source and convert it to the odata-result. Also the select-parameter works correct after my last commit.

BUT: There are much more cases, where this change is not enough!

I updated my fork with my latest changes. There are Unit-Tests that fail, because I couldn't figure out how to handle them in the correct way.

These failing tests are not existing tests, but new tests with custom struct types. These Test-Cases are relevant, because OData builds projections on the select-param.

AutoMapper ignores the use case that the propert-types may differ. I spend 2 days to find out how the ExpressionMapper.MappingVisitor.VisitMember-Method must be modified to meet all requirements, but i think it's not possible to have a general solution, cause the visitor sometimes expect that a not-converted expression is returned (of course, the current version is not aware of different struct-types and implicit casts)

My changes on the VisitMember-Method presume that the node-expression defines the type that is expected from outside, so the return-type should be converted to this type. This is not correct in all cases. The existing code breaks, if I always return a converted expression. Can someone please help me to

find out, when the return type should be changed
what's the best way to convert the expression to the return type (have a suggest in my code)
extend the functionality to full support of odata-parameters $select, $filter, $orderby
My defined Unit-Tests may help to understand what I'm try'in to achieve.

HRasch:
I think a found a good way to handle this implicit struct mapping stuff. Please review my latest version. I think there should be more test-cases be added to ensure all different kinds of mapping work.

src/AutoMapper/QueryableExtensions/ExpressionBuilder.cs
```diff
@@ -37,6 +37,7 @@ public class ExpressionBuilder : IExpressionBuilder
             new NullableDestinationExpressionBinder(),
             new NullableSourceExpressionBinder(),
             new AssignableExpressionBinder(),
+            new ImplicitAssignableExpressionBinder(),
```
lbargaoanu: Can this be done elsewhere? Because EF6 doesn't allow just any cast, so this could break things.

HRasch: Hard to decide, 2 Test cases fail if I remove it (Project_to_custom_struct and Project_to_custom_struct_with_mapping). But, if entity framework is used, this case should not occour, cause entity framework also doesn't support custom struct types, right?

lbargaoanu: I don't know about that, but the main use case for ProjectTo is EF, so it seems to me like this code should not be here.

lbargaoanu: @TylerCarlson1 @BlaiseD Wouldn't it be possible to achieve the same effect and have this code elsewhere, not on the EF path?

BlaiseD: I think you would need to use a different ExpressionVisitor with new extension methods to avoid affecting other users.  It would also help to create an issue to clarify the goal here.

lbargaoanu: I think that extension method can be UseAsDataSource.

TylerCarlson1: @lbargaoanu My best guess right now would be make the read only array be a list like IMapper list is so you can inject the ImplicitAssignableExpressionBinder at the configuration stage.
Like you can remove assignable mapper if you want if you want explicit cloning, or adding the Collection.Mappers stuff.

Then from there can go about making extension methods saying support OData projection or something along those lines and later on can make one for EF.Core perhaps.

lbargaoanu: Sounds good to me :) @jbogard Thoughts?

HRasch: @BlaiseD Sorry that i didn't open multiple issues to this case. Everything depends on supporting the mapping of custom struct types and the implicit cast to this types if operator-methods are defined. When I find a not supported case, I create a unit-test and try to fix it as best as I can.

The ImplicitAssignableExpressionBinder checks in addition to the other binders, if a type can be convertet implicit.

BlaiseD: FYI - The following tests fails against ReflectionHelper.CanImplicitConvert.  There's a suggested fix below -  may be incomplete also.

```csharp
    public class CanImplicitConvertFailedTests
    {
        [Fact]
        public void IListOfString_is_assignable_from_ListOfString()
        {
            IList<string> list = new List<string>();
            //Assert
            Assert.True(typeof(List<string>).CanImplicitConvert(typeof(IList<string>)));
        }

        [Fact]
        public void Int_is_assignable_from_short()
        {
            int i = (short)0;
            //Assert
            Assert.True(typeof(short).CanImplicitConvert(typeof(int)));
        }

        [Fact]
        public void Nullable_custom_string_is_assignable_from_custom_string()
        {
            CustString? ss = new CustString("");
            //Assert
            Assert.True(typeof(CustString).CanImplicitConvert(typeof(CustString?)));
        }

        [Fact]
        public void Nullable_int_digit_is_assignable_from_short()
        {
            IntDigit? i = (short)0;
            //Assert
            Assert.True(typeof(short).CanImplicitConvert(typeof(IntDigit?)));
        }

        [Fact]
        public void Nullable_int_is_assignable_from_int()
        {
            int? i = int.MaxValue;
            //Assert
            Assert.True(typeof(int).CanImplicitConvert(typeof(int?)));
        }

        [Fact]
        public void Nullable_int_is_assignable_from_nullable_short()
        {
            int? i = new short?();
            //Assert
            Assert.True(typeof(short?).CanImplicitConvert(typeof(int?)));
        }

        [Fact]
        public void Nullable_int_is_assignable_from_short()
        {
            int? i = (short)0;
            //Assert
            Assert.True(typeof(short).CanImplicitConvert(typeof(int?)));
        }

        [Fact]
        public void Short_is_assignable_from_sbyte()
        {
            short s = (sbyte)0;
            //Assert
            Assert.True(typeof(sbyte).CanImplicitConvert(typeof(short)));
        }
    }

    public struct CustString
    {
        public CustString(string i) { val = i; }
        public string val;

        public static implicit operator string(CustString i)
        {
            return i.val;
        }

        public static implicit operator CustString(string i)
        {
            return new CustString(i);
        }
    }

    public struct IntDigit
    {
        public IntDigit(int i) { val = i; }
        public int val;

        public static implicit operator int(IntDigit i)
        {
            return i.val;
        }

        public static implicit operator IntDigit(int i)
        {
            return new IntDigit(i);
        }
    }

    internal static class Helpers
    {
        //suggested extension
        public static bool CanImplicitConvert(this Type from, Type to)
        {
            if (to.IsAssignableFrom(from))
                return true;

            if (!(!to.IsNullable() && from.IsNullable()))
            {//Anything but To is NOT nullable and From IS nullable
                to = to.IsNullable() ? Nullable.GetUnderlyingType(to) : to;
                from = from.IsNullable() ? Nullable.GetUnderlyingType(from) : from;

                if (dict.ContainsKey(to) && dict[to].Contains(from))
                    return true;
            }

            bool ReturnTypeValid(Type returnType) => returnType == to || (dict.ContainsKey(to) && dict[to].Contains(returnType));
            bool ParameterValid(Type parameterType) => (parameterType == from) || (dict.ContainsKey(parameterType) && dict[parameterType].Contains(from));

            return from.GetMethods(BindingFlags.Public | BindingFlags.Static)
                            .Any(m => m.Name == "op_Implicit"
                                    && ReturnTypeValid(m.ReturnType)
                                    && ParameterValid(m.GetParameters().Single().ParameterType))
                            ||
                            to.GetMethods(BindingFlags.Public | BindingFlags.Static)
                            .Any(m => m.Name == "op_Implicit"
                                    && ReturnTypeValid(m.ReturnType)
                                    && ParameterValid(m.GetParameters().Single().ParameterType));
        }

        private static bool IsNullable(this Type type)
        {
            return type.IsGenericType && type.GetGenericTypeDefinition().Equals(typeof(Nullable<>));
        }

        static Dictionary<Type, HashSet<Type>> dict = new Dictionary<Type, HashSet<Type>>()
        {
            { typeof(decimal), new HashSet<Type> { typeof(byte), typeof(sbyte), typeof(char), typeof(short), typeof(ushort), typeof(int), typeof(uint), typeof(long), typeof(ulong) } },
            { typeof(double), new HashSet<Type> { typeof(byte), typeof(sbyte), typeof(char), typeof(short), typeof(ushort), typeof(int), typeof(uint), typeof(long), typeof(ulong), typeof(float) } },
            { typeof(float), new HashSet<Type> { typeof(byte), typeof(sbyte), typeof(char), typeof(short), typeof(ushort), typeof(int), typeof(uint), typeof(long), typeof(ulong) } },
            { typeof(ulong), new HashSet<Type> { typeof(byte), typeof(char), typeof(ushort), typeof(uint) } },
            { typeof(long), new HashSet<Type> { typeof(byte), typeof(sbyte), typeof(char), typeof(short), typeof(ushort), typeof(int), typeof(uint) } },
            { typeof(uint), new HashSet<Type> { typeof(byte), typeof(char), typeof(ushort) } },
            { typeof(int), new HashSet<Type> { typeof(byte), typeof(sbyte), typeof(char), typeof(short), typeof(ushort) } },
            { typeof(ushort), new HashSet<Type> { typeof(byte), typeof(char) } },
            { typeof(short), new HashSet<Type> { typeof(byte), typeof(sbyte) } }
        };
    }
```

HRasch: Hello,
I added your tests and got an error in AutoMapper.UnitTests.Projection.NestedAndArraysTests.LinqTests.Example

> Testname:	AutoMapper.UnitTests.Projection.NestedAndArraysTests.LinqTests.Example
> Test FullName:	AutoMapper.UnitTests.Projection.NestedAndArraysTests.LinqTests.Example
> Testquelle:	C:\Git-Repositories\AutoMapper\src\UnitTests\Projection\NestedAndArraysTests.cs > Zeile 49
> Testergebnis:	Fehlgeschlagene
>Testdauer:	0:00:00,008

> Ergebnis StackTrace:	
 bei lambda_method(Closure , Entity )
   bei AutoMapper.UnitTests.Projection.NestedAndArraysTests.LinqTests.Example() in C:\Git-Repositories\AutoMapper\src\UnitTests\Projection\NestedAndArraysTests.cs:Zeile 76.
Ergebnis Meldung:	System.InvalidCastException : Das Objekt des Typs "WhereSelectEnumerableIterator`2[AutoMapper.UnitTests.Projection.NestedAndArraysTests.LinqTests+SubEntity,System.String]" kann nicht in Typ "System.String[]" umgewandelt werden.

The reason is that my ImplicitAssignableExpressionBinder is returned because of the 
```csharp
if (to.IsAssignableFrom(from))
                return true;
```

Imo this check shouldn't be part of the Check, because the assignable binding is done by the AssignableExpressionBinder. When the check is removed, ImplicitAssignableExpressionBinder is not returned for the array-type and the test succeed, but some of your new Nullable-Tests fail. So I decided to fix the error by keeping the IsAssignableFrom-check but to move the ImplizitAssignableExpressionBinder to the end of the Bindings-List. This way the required EnumerableExpressionBinder has higher priority and all tests succeed.

Issue #2543, opened 2018-02-25
XpressionMapper does not convert the delegate return type.


BlaiseD:
This is related to #2538 . (Expression Projection with different types).

### Source/destination types

```csharp
public class Model
{
	public Guid? Id { get; set; }
	public MyString FullDescription { get; set; }
	public MyString ShortDescription { get; set; }
}

public class Dto
{
	public Guid? Id { get; set; }
	public string FullDescription { get; set; }
	public string ShortDescription { get; set; }
}
```

### Mapping configuration

```csharp
            var config = new MapperConfiguration(c =>
            {
                c.CreateMap<Model, Dto>()
                .ForMember(d => d.ShortDescription, opt => opt.MapFrom(s => s.ShortDescription.Value))
                .ReverseMap()
                .ForMember(d => d.ShortDescription, opt => opt.MapFrom(s => new MyString(s.ShortDescription)));
            });

            config.AssertConfigurationIsValid();

            var mapper = config.CreateMapper();
```

### Expected behavior

Mapping succeeds

### Actual behavior

Exception message "Expression of type 'ConsoleApp1.MyString' cannot be used for return type 'System.String'".

### Steps to reproduce

```csharp
            var source = new List<Model> {
                new Model
                {
                    Id = Guid.NewGuid(),
                    ShortDescription = "Model 1",
                    FullDescription = "Deluxe Foo Manufactured by Yoyodyne, Inc.",
                },
                new Model
                {
                    Id = Guid.NewGuid(),
                    ShortDescription = "Model 2",
                    FullDescription = "Deluxe Foo Manufactured by Yoyodyne, Inc.",
                }
            };

            Expression<Func<Model, MyString>> selection3 = s => s.ShortDescription;
            Expression<Func<Dto, string>> selection3Mapped = mapper.MapExpression<Expression<Func<Dto, string>>>(selection3);
```

Issue #2378, opened 2017-11-02
Expression mapping with coalesce operator is extremely slow or just hangs up


urasaber:
### Source/destination types

```csharp
public class Entity1
{
    public int? FolderId { get; set; }
}

public class Entity2
{
    public int FolderId { get; set; }
}
```

### Mapping configuration

```csharp
CreateMap<Entity1, Entity2>()
.ForMember(d => d.FolderId, o => o.MapFrom(s => s.FolderId ?? 0));
```

### Expected behavior
Expression maps successfully. At least at 4.1.1 it worked nicely and fast enough.

### Actual behavior
Mapping of ~15 conditions in one expression(Lambda debug view in VS: https://i.imgur.com/Sabxuhw.png) takes 3-6seconds
If I add one more OrElse to lambda expression then it takes ~12seconds
One more again - ~23seconds
64 items(Lambda debug view in VS: https://i.imgur.com/GPnif14.png) - never ends

### Steps to reproduce

```csharp
Mapper.Map<Expression<Func<Entity2, bool>>, Expression<Func<Entity1, bool>>>(expression)
```

We are updating our project from 4.1.1 to 6.1.1 and faced this issue. Am I missing something?
This is how it looks in dotTracer: https://i.imgur.com/OUFDPT0.png

lbargaoanu:
cc @BlaiseD

lbargaoanu:
@urasaber What happens is you remove the MapFrom. That should be a default conversion. If that fails, maybe ProjectTo could be used instead?

urasaber:
@lbargaoanu Well, if you remove the MapFrom it'll say that the type of source and dest differs, one is Nullable<> and other is Int32.

You know what, I think that creating such an expressions using OrElse is a bad anyway.
I have tried to create Expression using MethodCall on "Contains" of FolderIds array and now it works flawlessly.

Just wonder why expression visitor is such slow for an expression that mentioned in this issue.

BlaiseD:
@urasaber thanks for bringing this up.  I've called Visit a couple of time in VisitBinary just to check for compatibility - which is causing the preformance problem. It probably makes sense to just check the TypeMappings dictionary instead.  Please leave this open @lbargaoanu .  I'll create a PR.

lbargaoanu:
OK :)

PR #2544, opened 2018-02-25
Fix for #2543 -  XpressionMapper does not convert the delegate return…

BlaiseD:
… type.

Fixes issue #2543 .

### DIFF
--- a/src/AutoMapper/XpressionMapper/Extensions/MapperExtensions.cs
+++ b/src/AutoMapper/XpressionMapper/Extensions/MapperExtensions.cs
@@ -5,6 +5,7 @@ using System.Linq.Expressions;
 using static System.Linq.Expressions.Expression;
 using System.Reflection;
 using AutoMapper.Mappers.Internal;
+using AutoMapper.Internal;
 
 namespace AutoMapper.XpressionMapper.Extensions
 {
@@ -34,7 +35,17 @@ namespace AutoMapper.XpressionMapper.Extensions
             if (remappedBody == null)
                 throw new InvalidOperationException(Resource.cantRemapExpression);
 
+            remappedBody = VerifyType(typeDestFunc.GetGenericArguments().Last(), remappedBody);
+
             return (TDestDelegate)Lambda(typeDestFunc, remappedBody, expression.GetDestinationParameterExpressions(visitor.InfoDictionary, typeMappings));
+        }
+
+        private static Expression VerifyType(Type type, Expression remappedBody)
+        {
+            if (type != remappedBody.Type)
+                remappedBody = ExpressionFactory.ToType(remappedBody, type);
+
+            return remappedBody;
         }
 
         /// <summary>
@@ -73,6 +84,8 @@ namespace AutoMapper.XpressionMapper.Extensions
             var remappedBody = visitor.Visit(expression.Body);
             if (remappedBody == null)
                 throw new InvalidOperationException(Resource.cantRemapExpression);
+
+            remappedBody = VerifyType(typeDestFunc.GetGenericArguments().Last(), remappedBody);
 
             return (TDestDelegate)Lambda(typeDestFunc, remappedBody, expression.GetDestinationParameterExpressions(visitor.InfoDictionary, typeMappings));
         }
--- a/src/AutoMapper/XpressionMapper/Extensions/MapperExtensions.cs
+++ b/src/AutoMapper/XpressionMapper/Extensions/MapperExtensions.cs
@@ -35,18 +35,14 @@ namespace AutoMapper.XpressionMapper.Extensions
             if (remappedBody == null)
                 throw new InvalidOperationException(Resource.cantRemapExpression);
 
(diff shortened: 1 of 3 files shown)


BlaiseD:
@HRasch - the tests in #2542 seem to work for both expression (Xpression) mapping and UseAsDatSource if the configuration is modified. [See CustomStructTests](https://github.com/BlaiseD/AutoMapper/blob/master/src/UnitTests/CustomStructTests.cs). Sounds like you would like to make this work by specifying the types alone.  That would be great.  Note #2378 for any performance impact.

src/AutoMapper/XpressionMapper/Extensions/MapperExtensions.cs
```diff
@@ -33,8 +34,18 @@ public static TDestDelegate MapExpression<TDestDelegate>(this IMapper mapper, La
             var remappedBody = visitor.Visit(expression.Body);
             if (remappedBody == null)
                 throw new InvalidOperationException(Resource.cantRemapExpression);
+
+            remappedBody = VerifyType(typeDestFunc.GetGenericArguments().Last(), remappedBody);
```
lbargaoanu: Simply replacing VerifyType with ToType should be enough here.

BlaiseD: Good  point.

src/AutoMapper/XpressionMapper/Extensions/MapperExtensions.cs
```diff
@@ -33,9 +34,15 @@ public static TDestDelegate MapExpression<TDestDelegate>(this IMapper mapper, La
...
             if (remappedBody == null)
                 throw new InvalidOperationException(Resource.cantRemapExpression);
+
+            return (TDestDelegate)Lambda
+                                    (
+                                        typeDestFunc,
+                                        ExpressionFactory.ToType(remappedBody, typeDestFunc.GetGenericArguments().Last()),
+                                        expression.GetDestinationParameterExpressions(visitor.InfoDictionary, typeMappings)
+                                    );
+        }
```
lbargaoanu: The two MapExpression methods are almost the same, only the visitor differs. Some code reuse seems in order.

lbargaoanu:
That convert code seems ad hoc. Because we already have that code in the mappers, so maybe it should try to use those as the normal mapping does?

src/AutoMapper/XpressionMapper/Extensions/MapperExtensions.cs
BlaiseD: I agree. The simplest method would be to pass a `Func<Dictionary<Type, Type>, XpressionMapperVisitor> func` to the consolidated method.  That belongs in a separate PR though - we did not add the two methods in this PR -  unless you want to fix it.

BlaiseD:
Please be specific about how you would do it differently.

src/AutoMapper/XpressionMapper/Extensions/MapperExtensions.cs
lbargaoanu: Nah, I'm good :)

lbargaoanu:
I was talking about the convert code in that other PR. I don't really know what's all that about. But if the types don't match, I was thinking maybe mapping to the needed type would help. As we already have mappers that consider the conversion operators that might apply. Does that make sense? :)

BlaiseD:
Ah - sorry I misread you.  It's possible. I'll have to look at the code further.  The challenge is with binary expressions e.g. int? == int, CustomString == System.String etc...  Any side could be a method call, member, variable, or constant. The base ExpressionVisitor.VisitBinary will throw an exception when the two sides do not match (unless one side is converted in the override).  From #2378 - the fix could be expensive - or maybe it is easier than I thought.

lbargaoanu:
Then I guess mapping doesn't really make sense.

HRasch:
Excuse me if I'm wrong, but the compare on the VisitBinary is just one case. OData e.g. tries to create projections with MemberInit and if the types are not equal, exceptions are thrown. I tried to fix the problem by introducing a new parameter to the MappingVisitor-Class, telling what return type is expected, so I can decide if i have to convert the member to a specific type (by a typemap-lookup or implicit cast). But i failed because of calls like _parentMappingVisitor.Visit, where I have no control over the visitor.
In the end it feelt like the MappingVisitor is just doing to much and maybe another visitor should be introduced, just for the type-convertings. Then, in case of a binary expression, the left expression should control the expected type of the right expression. This should also work with MemberInit and MemberBind.

BlaiseD:
@jbogard @lbargaoanu - Do you have reservations about this one? Seems like it's been a while.

lbargaoanu:
I could go either way :), so I just left it for @jbogard.

BlaiseD:
Clear - I think. There are related changes I'd like to make so I'm going to close this one and address #2543 in a different PR.  Thanks.

PR #2575, opened 2018-03-20
.NET Standard 2.0 support

### LINKED ISSUE #2572: Add a NetStandard 2.0 target
Even though there won't be a difference in the mapping we support, `netstandard2.0` fixes some dependency explosions. Also, as part of this, change the `#if` conditions to be feature flags instead of platform checks. Then for each platform, indicate what features are supported.

jbogard:
This PR bumps the version to 7.0. Drops support for .NET 4.0, the tooling around it was too annoying to keep up. Sorry @sunnycase 

Adds xunit tests for `netcoreapp1.1` along with `net452` and `netcoreapp2.0`.

Removes `netstandard1.1` support. This was just Windows Phone anyway. Going forward we'll have:

 - `net45`
 - `netstandard1.3`
 - `netstandard2.0`

This removed a number of `#if` preprocessor directives.

Converted as many projects as I could to SDK-style projects.

Fixes #2572


jbogard:
@AutoMapper/developers this also bumps to 7.0.0 since it affects the target frameworks. I think after this will be to put expression mapping into its own repo.

TylerCarlson1:
@jbogard Like AM.Expressions?
And will that extend even that to AM.Expressions.EF6 and AM.Expressions.EFCore in there own repositiories?
Or perhaps multiple expression translation options when adding expression translation.

jbogard:
There are a few different mappings:

- In-memory
- Dynamic/dictionary (kinda)
- QueryableExtensions
- UseAsDataSource
- Expression mapping (XpressionMapper)

The idea would be to move the non in-memory packages out. I dunno about having EF6/EFCore packages, as there are other things that use the LINQ support so maybe we just put hooks in so that EF6/EFCore packages could alter the expressions generated.

TylerCarlson1:
That can be discussed later.  If it's just to move to AM.Expressions, then you can have EF6 EFCore things that alter the mapping configuration from InMemory to suit there needs.

Another option might be to make EF6 and Core there own repo's and that way they can identify the sources of the IQueryable and then alter the in-memory mapping logic to work for what they need.
So if you do a `db.Set<T>()` it can identify it as such and alter the query to make it compatible.  But for that you need a ref to the dll's to identify more than likely.

Either way should work, and just moving the Expression mapping out from AutoMapper's repo should be enough to make it clear which way to go once it's transferred.

Issue #2578, opened 2018-03-21
Automapper should sort properties in ProjectTo consistently


dozer75:
There is an issue with automapper and generating select statement using ProjectTo, while it works most of the time, it doesn't create the select list the same way for the same destination type all the time. This causes the following exception in EF when you try to create a EF Union between two different configurations for the same destination.

`System.NotSupportedException: 'The type 'AutoMapperTest.DTO.Entity' appears in two structurally incompatible initializations within a single LINQ to Entities query. A type can be initialized in two places in the same query, but only if the same properties are set in both places and those properties are set in the same order.'`

There are several ways to reproduce this, but this [gisthub](https://gist.github.com/dozer75/073fca9b6e83fab514f75320a6af69f1) example is the simplest one.

The ideal solution would be that when ProjectTo generates the Select clause it either sorts the properties by name or by position in the destination type. It would then always be the same no matter the what the source is.

NOTE: I see an issue in the circumstances where one of the mappings to the same destnation types ignores a property and thus impossible to solve, but other than these I think that this could be solved by a simple sort of property rules before creating select expression.

### Source/destination types

```csharp
// Destination
    public class DTO
    {
        public string Col1 { get; set; }
        public string Col2 { get; set; }
        public int Id { get; set; }
    }

// Sources
    public class Entity1
    {
        public string Col1 { get; set; }
        public string Col2 { get; set; }
        public int Id { get; set; }
    }

    public class Entity2
    {
        public string Col1 { get; set; }
        public int Id { get; set; }
    }
```

### Mapping configuration

```csharp
    Mapper.Initialize(config =>
                        {
                            config.CreateMap<Entity1, DTO>();
                            config.CreateMap<Entity2, DTO>()
                                .ForMember(dest => dest.Col2, opt => opt.UseValue("abc"));
                        });
```

### Expected behavior

The generated select for Entity2 -> DTO should have the same order of property assignment as Entity1 -> DTO select has no matter what the source type and configuration are.

### Actual behavior

The generated select for Entity2 -> DTO have Col2 as the last property assigned, while the generated select for Entity1 -> DTO has Col2 in the same position as it is in the DTO type.

These are the expression debug views of the generated selects:

Entity1:

    .New AutoMapperTest.DTO(){
        Col1 = $dto.Col1,
        Col2 = $dto.Col2,
        Id = $dto.Id
    }
Entity2:

    .New AutoMapperTest.DTO(){
        Col1 = $dto.Col1,
[...]

PR #2580, opened 2018-03-22
Fix for Issue #2578

dozer75:
This pull request solves the issue around #2578 so that `DestinationProperty.Name` is a part of there ordering of property maps when `TypeMap.GetPropertyMaps()` is called on a sealed type map.

`PropertyMap.MappingOrder` will be primary sorting property, then `PropertyMap.DestinationProperty.Name` if any `PropertyMap.MappingOrder` are identical. 

All property maps with no `MappingOrder` will be sorted after all property maps with a valid `PropertyMap.MappingOrder` ordred by `PropertyMap.DestinationProperty.Name`.

If no property maps has `PropertyMap.MappingOrder` set, it will sort on `PropertyMap.Destinationproperty.Name` only.

### DIFF
--- a/src/AutoMapper/TypeMap.cs
+++ b/src/AutoMapper/TypeMap.cs
@@ -284,7 +284,8 @@ namespace AutoMapper
             _orderedPropertyMaps =
                 _propertyMaps
                     .Union(_inheritedMaps)
-                    .OrderBy(map => map.MappingOrder).ToArray();
+                    .OrderBy(map => map.MappingOrder ?? int.MaxValue).ThenBy(map => map.DestinationProperty?.Name)
+                    .ToArray();
 
             MapExpression = new TypeMapPlanBuilder(configurationProvider, this).CreateMapperLambda(typeMapsPath);
         }
--- a/src/UnitTests/MappingOrder.cs
+++ b/src/UnitTests/MappingOrder.cs
@@ -167,15 +167,13 @@ namespace AutoMapper.UnitTests
             [Fact]
             public void Should_use_not_have_any_configured_mapping_orders()
             {
-                Assert.All(_propertyMaps, pm => pm.MappingOrder = null);
+                Assert.True(_propertyMaps.All(pm => pm.MappingOrder == null));
             }
 
 
             [Fact]
             public void Should_perform_mapping_ordered_by_name()
             {
-                Assert.Equal(6, _propertyMaps.Length);
-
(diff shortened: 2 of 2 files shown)


src/AutoMapper/TypeMap.cs
```diff
@@ -284,7 +284,8 @@ public void Seal(IConfigurationProvider configurationProvider, Stack<TypeMap> ty
             _orderedPropertyMaps =
                 _propertyMaps
                     .Union(_inheritedMaps)
-                    .OrderBy(map => map.MappingOrder).ToArray();
+                    .OrderBy(map => map.MappingOrder ?? int.MaxValue).ThenBy(map => map.DestinationProperty?.Name)
+                    .ToArray();
```
lbargaoanu: The issue is about EF and ProjectTo, Map works just fine, so I don't see any reason to touch Map. Also the order by name is too obvious and people might be tempted to rely on it. I would order by DestinationProperty.MetadaToken. That should be consistent and faster.

dozer75: The reason for touching it where I did was that MappingOrder was already done the same place, and to avoid having the same logic happen different places I put it there to avoid confusion from for others. I can't really see the reason why MappingOrder should be here and not default mapping?

As for DestinationProperty it doesn't have any MetadataToken?

lbargaoanu: Yes, I see it's not easy to get a hold of that MetadataToken. Then I guess there is nothing else to sort on. DestinationProperty will always be set, so that null check is useless. I don't think MappingOrder makes much sense for ProjectTo, so I'd rather have a built in ordering, but @jbogard  might disagree.

dozer75: People may use MappingOrder already, so I think ignoring it would be a bad idea for them.

jbogard: MappingOrder is irrelevant for ProjectTo though. I think the fix is just to order by name during projection. Can you write an integration test that shows the failure? I'd really rather see a LINQ query that uses a union that actually shows a failure.

dozer75: Ok! I'll move the code to the projection! An integration test will also be made!

dozer75:
I close this PR and creates a new one to avoid unnecessary commits and creates a new one based on ordering on projection.

PR #2581, opened 2018-03-26
Set PreserveReferences for dynamic maps and constructor mapping

lbargaoanu:
Some white space in the tests file, the new tests are at the top.


src/AutoMapper/Execution/TypeMapPlanBuilder.cs
```diff
@@ -72,71 +78,86 @@ public LambdaExpression CreateMapperLambda(Stack<TypeMap> typeMapsPath)
...
-                    SetPreserveReferences(propertyTypeMap);
-                    foreach(var derivedTypeMap in propertyTypeMap.IncludedDerivedTypes.Select(ResolveTypeMap))
+                if(typeMapsPath.Contains(propertyTypeMap))
+                {
+                    if(propertyTypeMap.SourceType.IsValueType())
                     {
-                        SetPreserveReferences(derivedTypeMap);
+                        if(propertyTypeMap.MaxDepth == 0)
+                        {
+                            propertyTypeMap.MaxDepth = 10;
```
lbargaoanu: Because 10 should be enough for everybody :)

Issue #2593, opened 2018-04-06
ProjectTo error with aggregate functions on IEnumerable<T>


KSid:
ProjectTo fails with to an unhandled exception when trying to project to a destination containing an aggregate function (e.g. Count) where the matching source is an ```IEnumerable<T>```. This works as expected when the aggregate function is Count and the source is an ```ICollection<T>``` (e.g. ```List<T>```) because the name matches a property on the destination object. Other aggregate functions cause the same error with ```ICollection<T>``` (but not Sum which is dicussed below).

The [documentation](https://automapper.readthedocs.io/en/latest/Queryable-Extensions.html#aggregations) suggests this should work.

This error is due to the correct extension method being located but then treated as a field/property accessor in [MemberGetterExpressionResultConverter.cs](https://github.com/AutoMapper/AutoMapper/blob/aa1bc8600b07b6d73f3126011b6987d38edcf4dc/src/AutoMapper/QueryableExtensions/Impl/MemberGetterExpressionResultConverter.cs#L21):
```csharp
private static ExpressionResolutionResult ExpressionResolutionResult(
    ExpressionResolutionResult expressionResolutionResult, MemberInfo getter)
{
    var member = Expression.MakeMemberAccess(expressionResolutionResult.ResolutionExpression, getter);
    return new ExpressionResolutionResult(member, member.Type);
}
```

I replaced the method above with the code below while investigating and it passes existing tests. This also allows other aggregate functions (e.g. Sum, Max, Min, etc.) to work with ```IEnumerable<T>``` but not all aggregate functions work with ```ICollection<T>``` (e.g. Sum) and I haven't found the cause. This second behaviour can be seen by changing the ```Points``` property to an ```ICollection<T>``` and then back to an ```IEnumerable<T>```.

```csharp
private static ExpressionResolutionResult ExpressionResolutionResult(
    ExpressionResolutionResult expressionResolutionResult, MemberInfo getter)
{
    var member = (getter is MethodInfo method && method.IsStatic)
        ? Expression.Call(method, expressionResolutionResult.ResolutionExpression) as Expression
        : Expression.MakeMemberAccess(expressionResolutionResult.ResolutionExpression, getter);
    return new ExpressionResolutionResult(member, member.Type);
}
```

I have created a [gist with a full example](https://gist.github.com/KSid/45040b68b97eae3b6769b8ef3b685be1).

[...]

PR #2602, opened 2018-04-11
Fix projection to destinations using extension methods #2593

KSid:
Fixes the main issue in #2593.

### DIFF
--- a/src/AutoMapper/QueryableExtensions/Impl/MemberGetterExpressionResultConverter.cs
+++ b/src/AutoMapper/QueryableExtensions/Impl/MemberGetterExpressionResultConverter.cs
@@ -21,7 +21,9 @@ namespace AutoMapper.QueryableExtensions.Impl
         private static ExpressionResolutionResult ExpressionResolutionResult(
             ExpressionResolutionResult expressionResolutionResult, MemberInfo getter)
         {
-            var member = Expression.MakeMemberAccess(expressionResolutionResult.ResolutionExpression, getter);
+            var member = (getter is MethodInfo method)
+                ? (Expression)Expression.Call(method, expressionResolutionResult.ResolutionExpression)
+                : Expression.MakeMemberAccess(expressionResolutionResult.ResolutionExpression, getter);
             return new ExpressionResolutionResult(member, member.Type);
         }
 
--- a/src/UnitTests/Bug/CannotMapICollectionToAggregateSumDestination.cs
+++ b/src/UnitTests/Bug/CannotMapICollectionToAggregateSumDestination.cs
@@ -1,0 +1,53 @@
+﻿using Shouldly;
+using Xunit;
+using System.Linq;
+using AutoMapper.QueryableExtensions;
+using System.Collections.Generic;
+
+namespace AutoMapper.UnitTests.Bug
+{
+    public class CannotMapICollectionToAggregateSumDestination
+    {
+        class DummySource
+        {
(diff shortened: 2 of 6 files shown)


src/AutoMapper/QueryableExtensions/Impl/MemberGetterExpressionResultConverter.cs
```diff
@@ -21,7 +21,9 @@ public ExpressionResolutionResult GetExpressionResolutionResult(ExpressionResolu
         private static ExpressionResolutionResult ExpressionResolutionResult(
             ExpressionResolutionResult expressionResolutionResult, MemberInfo getter)
         {
-            var member = Expression.MakeMemberAccess(expressionResolutionResult.ResolutionExpression, getter);
+            var member = (getter is MethodInfo method && method.IsStatic && method.GetParameters().Count() == 1)
+                ? Expression.Call(method, expressionResolutionResult.ResolutionExpression) as Expression
```
jbogard: Consider just using a cast here instead of "as" which returns null if the value is not of that type (and we know that's impossible here).

src/AutoMapper/QueryableExtensions/Impl/MemberGetterExpressionResultConverter.cs
```diff
@@ -21,7 +21,9 @@ public ExpressionResolutionResult GetExpressionResolutionResult(ExpressionResolu
         private static ExpressionResolutionResult ExpressionResolutionResult(
             ExpressionResolutionResult expressionResolutionResult, MemberInfo getter)
         {
-            var member = Expression.MakeMemberAccess(expressionResolutionResult.ResolutionExpression, getter);
+            var member = (getter is MethodInfo method && method.IsStatic && method.GetParameters().Count() == 1)
```
lbargaoanu: The methods are already filtered, so just checking whether it's a method should be enough.

KSid:
Thank you @jbogard and @lbargaoanu for reviewing the changes. I've tweaked the code [with](https://github.com/AutoMapper/AutoMapper/pull/2602#pullrequestreview-111340826) [your](https://github.com/AutoMapper/AutoMapper/pull/2602#pullrequestreview-111625958) [suggestions](https://github.com/AutoMapper/AutoMapper/pull/2603#pullrequestreview-111623684) and added some integration tests. I've rolled #2603 into this PR as that test needed the initial fix here. I'll rebase and squash into the two main commits (or whatever you prefer) once you are happy.

Would you like the integration tests to be renamed or placed somewhere else (e.g. under AutoMapper.IntegrationTests.Projections which doesn't exist at the moment)?

src/UnitTests/Bug/CannotMapICollectionToAggregateSumDestination.cs
```diff
@@ -0,0 +1,34 @@
...
+            // arrange
+            var config = new MapperConfiguration(cfg =>
+            {
+                cfg.CreateMap<DummySource, DummyDestination>();
+            });
+
+            // act
+            // do nothing
+
+            // assert
```
lbargaoanu: Maybe do the same as the other test, check that it really works?

KSid:
No problem! Would you like me to rebase and squash now or wait for the PR to be fully approved first?

lbargaoanu:
I think you can do it now.

lbargaoanu:
@jbogard I think this is good to go.

lbargaoanu:
Merge? :)

PR #2603, opened 2018-04-11
Fix TypeDetails ignoring non-generic extension methods #2593

KSid:
Fixes the secondary issue in #2593.

### DIFF
--- a/src/AutoMapper/TypeDetails.cs
+++ b/src/AutoMapper/TypeDetails.cs
@@ -107,14 +107,17 @@ namespace AutoMapper
             {
                 genericInterfaces = genericInterfaces.Union(new[] { Type });
             }
+
             return explicitExtensionMethods.Union
             (
-                from genericMethod in sourceExtensionMethodSearch
-                where genericMethod.IsGenericMethodDefinition
                 from genericInterface in genericInterfaces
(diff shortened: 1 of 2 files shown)


src/AutoMapper/TypeDetails.cs
```diff
@@ -107,14 +107,17 @@ private IEnumerable<MethodInfo> BuildPublicNoArgExtensionMethods(IEnumerable<Met
...
                 from genericInterface in genericInterfaces
                 let genericInterfaceArguments = genericInterface.GetTypeInfo().GenericTypeArguments
-                where genericMethod.GetGenericArguments().Length == genericInterfaceArguments.Length
-                let methodMatch = genericMethod.MakeGenericMethod(genericInterfaceArguments)
+                from extensionMethod in sourceExtensionMethodSearch
+                where !extensionMethod.IsGenericMethodDefinition
+                    || extensionMethod.GetGenericArguments().Length == genericInterfaceArguments.Length
+                let methodMatch = extensionMethod.IsGenericMethodDefinition
+                    ? extensionMethod.MakeGenericMethod(genericInterfaceArguments)
+                    : extensionMethod
```
lbargaoanu: I think two queries linked by Concat make more sense here. One for generic methods and one for the others.

KSid:
Merged into #2602 as the integration test relies on that fix

PR #2604, opened 2018-04-12
Fix to issue #2578

dozer75:
Second suggestion for fix to issue #2578.

Added OrderBy when getting the property maps when creating memberbindings, sorting first on `PropertyMap.MappingOrder`, then on the `DestinationProperty.Name` to ensure that any usage of `PropertyMap.MappingOrder` is honored before default sorting on `DestinationProperty.Name` to avoid break any existing usages of AutoMapper in these circumstances for other scenarios.

Added both integration test for this specific issue and unit tests.


src/AutoMapper/QueryableExtensions/ExpressionBuilder.cs
```diff
@@ -209,15 +209,18 @@ protected override Expression VisitNew(NewExpression node)
         private List<MemberBinding> CreateMemberBindings(ExpressionRequest request, TypeMap typeMap, Expression instanceParameter, TypePairCount typePairCount, LetPropertyMaps letPropertyMaps)
         {
             var bindings = new List<MemberBinding>();
-            foreach (var propertyMap in typeMap.GetPropertyMaps().Where(pm => 
-                (!pm.ExplicitExpansion || request.MembersToExpand.Contains(pm.DestinationProperty)) && pm.CanResolveValue() && ReflectionHelper.CanBeSet(pm.DestinationProperty)))
+            foreach (var propertyMap in typeMap.GetPropertyMaps()
+                                               .Where(pm => (!pm.ExplicitExpansion || request.MembersToExpand.Contains(pm.DestinationProperty)) && 
+                                                            pm.CanResolveValue() && ReflectionHelper.CanBeSet(pm.DestinationProperty))
+                                               .OrderBy(pm => pm.MappingOrder ?? int.MaxValue).ThenBy(pm => pm.DestinationProperty.Name))
```
lbargaoanu: The ordering by MappingOrder is not needed.

dozer75: Removed! However I think that it  would be better that it remained here as mentioned to avoid breaking already implementations using MappingOrder.

PR #2650, opened 2018-05-25
Condition vs PreCondition docs

mrpmorris:
1: Changed "must be met before that property will be mapped" to "must be met for that property to be mapped" - so that the reader doesn't infer order of events from the sentence due to the use of the word "before".

2: Changed case of "precondition" to "PreCondition" to make it clear it is a method (I thought you was talking about a practice, not a method).

3: Added a more detailed explanation of the differences between the two, including test cases.

### DIFF
--- a/docs/Conditional-mapping.md
+++ b/docs/Conditional-mapping.md
@@ -90,17 +90,3 @@ Mapper.Initialize(cfg =>
       });
 });
 ```
-
-### Case 3 - Implicit null checking
-
-When using the `MapFrom` method it is often not necessary to use either a `Condition` or a `PreCondition` as AutoMapper will perform null-checking along the way.
-
-```c#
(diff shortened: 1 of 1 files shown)


mrpmorris:
I've removed case 3. Now it is only about Condition vs PreCondition. You currently don't have any documentation regarding PreCondition at all that I can see.

lbargaoanu:
The problem remains.

mrpmorris:
What is "too much information" about it? It describes the difference between Condition and PreCondition in a section that is about Conditional mapping. What do you think still needs to come out?

lbargaoanu:
You can leave the page as it is, except adding a link to the new content, with some short description perhaps.

mrpmorris:
I don't understand. This file is the documentation topic for Conditions, why would the information belong elsewhere?

PR #2651, opened 2018-05-26
Add a section about conditions and resolvers

lbargaoanu:
Closes #2650.


mrpmorris:
IMHO 2650 is far clearer. It's much easier to understand examples you are likely to write and show why they are right/wrong, and the section on the flow of events (numbered sequence) is better too.

lbargaoanu:
@jbogard I don't see these changes in the latest docs.

jbogard:
No? I can see them now.

lbargaoanu:
Well, [no](http://automapperdocs.readthedocs.io/en/latest/Custom-value-resolvers.html) :) If it's just smth on my end, it doesn't matter.

jbogard:
http://docs.automapper.org/en/stable/

:)

lbargaoanu:
:) Yes, I never know which one of those url-s is the good one. And I'm guessing most people don't :)

PR #2658, opened 2018-06-01
Break out expression mapper

### LINKED ISSUE #2070: Break out "special" mappers from main repo
And introduce new repos and packages for them.

This would include:

- LINQ projection (ProjectTo)
- ExpressionMapper
- Dynamic/dictionary

This would cut down the core repo to just the in-memory object-to-object mapping. A couple of reasons to do this:

- Not every uses the above features
- Having all in the same repo is confusing - we've got 3-4 different "ways" to map from A-B
- Make it more obvious that these mappings go outside the normal configuration, so your MapFrom etc won't necessarily work

I already did this with IDataReader. The packages could be:

- AutoMapper.Linq
- AutoMapper.ExpressionMapping
- AutoMapper.DynamicMapping

jbogard:
Fixes #2070 

Heads up @BlaiseD


jbogard:
See https://github.com/AutoMapper/AutoMapper.Extensions.ExpressionMapping

BlaiseD:
ok @jbogard  -  will move changes for #2622 to the new repo.  Thanks.

PR #2677, opened 2018-06-12
Parameterized queries

### LINKED ISSUE #2673: Parameterized projections do not parameterize the query
From [a comment](https://lostechies.com/jimmybogard/2014/12/30/automapper-3-3-feature-parameterized-projections/#comment-3935180242), parameterized projections do not result in a parameterized query. Doing this manually in EF6 *does* result in a parameterized query.

Given our models:

```c#
public class Entity
{
    public int Id { get; set; }
}

public class EntityDto
{
    public int Id { get; set; }
}

private class ClientContext : DbContext
{
    static ClientContext()
    {
        Database.SetInitializer(new DatabaseInitializer());
    }

    public DbSet<Entity> Entities { get; set; }
}
```

And this configuration:

```c#
new MapperConfiguration(cfg =>
{
    string username = null;
    cfg.CreateMap<Entity, EntityDto>()
        .ForMember(d => d.UserName, opt => opt.MapFrom(s => username));
});
```

This test passes:

```c#
[Fact]
public async Task Should_parameterize_value()
{
    using (var db = new ClientContext())
    {
        string username = "Joe";
        var dtos = await db.Entities.ProjectTo<EntityDto>(Configuration, new {username}).ToListAsync();

        dtos.All(dto => dto.UserName == username).ShouldBeTrue();

        username = "Mary";
        dtos = await db.Entities.ProjectTo<EntityDto>(Configuration, new { username }).ToListAsync();
        dtos.All(dto => dto.UserName == username).ShouldBeTrue();

        username = "Jane";
        dtos = await db.Entities.Select(e => new EntityDto
        {
            Id = e.Id,
            Value = e.Value,
            UserName = username
        }).ToListAsync();
        dtos.All(dto => dto.UserName == username).ShouldBeTrue();
    }
}
```

However, the execution plan is different for the last query vs. the first two. The first two hardcode the parameter value:

```sql
SELECT 
    [Extent1].[Id] AS [Id], 
    N'Joe' AS [C1], 
    [Extent1].[Value] AS [Value]
    FROM [dbo].[Entities] AS [Extent1]

SELECT 
    [Extent1].[Id] AS [Id], 
    N'Mary' AS [C1], 
    [Extent1].[Value] AS [Value]
    FROM [dbo].[Entities] AS [Extent1]
```

But the last one parameterizes it:

```sql
exec sp_executesql N'SELECT 
    [Extent1].[Id] AS [Id], 
    [Extent1].[Value] AS [Value], 
    @p__linq__0 AS [C1]
    FROM [dbo].[Entities] AS [Extent1]',N'@p__linq__0 nvarchar(4000)',@p__linq__0=N'Jane'
```

jbogard:
Fixes #2673

Made an additional overload to `ProjectTo` to take an expression for the parameters. The underlying expression builder then simply swaps one member access for the other based on name.

### DIFF
--- a/src/AutoMapper/QueryableExtensions/ExpressionBuilder.cs
+++ b/src/AutoMapper/QueryableExtensions/ExpressionBuilder.cs
@@ -19,6 +19,7 @@ namespace AutoMapper.QueryableExtensions
     public interface IExpressionBuilder
     {
         LambdaExpression[] GetMapExpression(Type sourceType, Type destinationType, ParameterBag parameters, MemberInfo[] membersToExpand);
+        LambdaExpression[] GetMapExpression(Type sourceType, Type destinationType, Expression<Func<object>> parameterExpression, MemberInfo[] membersToExpand);
         LambdaExpression[] CreateMapExpression(ExpressionRequest request, TypePairCount typePairCount, LetPropertyMaps letPropertyMaps);
         Expression CreateMapExpression(ExpressionRequest request, Expression instanceParameter, TypePairCount typePairCount, LetPropertyMaps letPropertyMaps);
     }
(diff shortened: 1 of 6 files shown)


lbargaoanu:
Tricky :)

jbogard:
Dictionaries won't work, EF won't support those expressions. I wonder if it's worth deprecating the dictionary-based params, since it can allow SQL injection.

jbogard:
@lbargaoanu This is slightly less clever, but I think it's more obvious than adding an overload. Instead of switching out member access value, I just point to the new object through a constant passed in.

lbargaoanu:
Yes, it seems clearer. I'll take a look.

src/AutoMapper/QueryableExtensions/ExpressionBuilder.cs
```diff
@@ -302,7 +345,29 @@ private Expression NullCheck(Expression input)
...
+
+            protected override Expression VisitMember(MemberExpression node)
+            {
+                if (!node.Member.DeclaringType.Has<CompilerGeneratedAttribute>())
+                {
+                    return base.VisitMember(node);
+                }
+                var parameterName = node.Member.Name;
+
+                var matchingMember = _parameters.GetType().GetDeclaredProperty(parameterName);
```
lbargaoanu: There is some reflection here on very call, but we already have that with CompilerGeneratedAttribute above.

lbargaoanu:
Dictionary based parameters are based on strings and vulnerable to SQL injection. But they are also more flexible. You can pass them at runtime. The object parameters seem useful at compile time.

jbogard:
Actually looking again, I don't think it's vulnerable to SQL injection.
It's just that it's passed in as a value to the projection instead of a
parameter, but it still has to go through the query provider.

The only reason I added the dictionary was because it was what MVC did at
the time. Anyway, at least the object one uses parameterized queries now.

lbargaoanu:
Yes, I see it quotes apostrophes for example. After all, the provider takes the expression and generates the SQL, so it does its job.

PR #2682, opened 2018-06-13
Remove package references to System.Reflection.Emit

### LINKED ISSUE #2680: System.Reflection.Emit is a deprecated package
We should remove it and replace with the correct packages:

https://www.nuget.org/packages/System.Reflection.Emit/

jbogard:
Fixes #2680 

I had to target `netcoreapp2.0` directly. This will be an issue for folks using referencing AutoMapper transitively.

See https://github.com/dotnet/corefx/issues/29365

I think based on this we may want to consider removing any references to `System.Reflection.Emit`. This would mean losing support for mapping to interfaces, as well as the way we're doing `let` expressions for complex MapFrom's.

### DIFF
--- a/src/AutoMapper/AutoMapper.csproj
+++ b/src/AutoMapper/AutoMapper.csproj
@@ -3,7 +3,7 @@
   <PropertyGroup>
     <Summary>A convention-based object-object mapper.</Summary>
     <Description>A convention-based object-object mapper.</Description>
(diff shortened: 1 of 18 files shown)


lbargaoanu:
About dropping Emit altogether. The queries for the sub maps were terrible. So I think it would be fair to fail by default. Opt in maybe. I think we should just keep the define, but then again, I would say that, wouldn't I? :)

jbogard:
lol well the mapping to interfaces is interesting but I don't understand any of that code

PR #2698, opened 2018-06-23
 Implement ShouldUseConstructor

### LINKED ISSUE #2691: Limit which constructors can be used
AutoMapper will quite happily use any/all constructors when determining how to construct an object, even private/internal ones.

I propose a setting (however deeply buried), that allows the setting of a required minimum scope. There are times where I have an empty private/internal constructor and really don't want that to be called.

I don't want to have explicity set the constructor to use in all of my mappings.

Is there any appetite for this, if so, I'll take a look at putting a PR together.

lxalln:
Allows for constructors to be excluded in the same way as properties
and fields (ShouldMapField & ShouldMapProperty).
Fixes #2691.


src/AutoMapper/Configuration/IProfileConfiguration.cs
```diff
@@ -39,6 +39,12 @@ public interface IProfileConfiguration
         /// </summary>
         Func<FieldInfo, bool> ShouldMapField { get; }
 
+        /// <summary>
+        /// Specify which constructors should be mapped.
+        /// By default all non-static constructors are mapped.
+        /// </summary>
+        Func<ConstructorInfo, bool> ShouldMapCtor { get; }
```
lbargaoanu: I think ShouldUseConstructor is a better name.

docs/Configuration.md
```diff
@@ -176,9 +176,9 @@ Mapper.Initialize(cfg => {
 });
 ```
 
-## Global property/field filtering
+## Global property/field/constructor filtering
 
-By default, AutoMapper tries to map every public property/field. You can filter out properties/fields with the property/field filters:
+By default, AutoMapper tries to map every public property/field, and all non-static declared constructors. You can filter out properties/fields/constructors with the property/field/constructor filters:
```
lbargaoanu: I think the wording is strange, AM doesn't map constructors, it uses them to construct destination objects. Also there is a special page for constructors, so the content should be there, not here.

src/AutoMapper/ProfileMap.cs
```diff
@@ -1,273 +1,275 @@
...
-                Configure(typeMapConfiguration, configurationProvider);
-                if (typeMapConfiguration.ReverseTypeMap != null)
-                {
-                    Configure(typeMapConfiguration.ReverseTypeMap, configurationProvider);
-                }
-            }
-        }
-
-        private void BuildTypeMap(IConfigurationProvider configurationProvider, ITypeMapConfiguration config)
+using System;
```
lbargaoanu: I know GitHub sometimes messes the diff, but if you could fix it somehow, that would be nice :)

src/AutoMapper/TypeDetails.cs
```diff
@@ -69,8 +72,10 @@ private static IEnumerable<string> PostFixes(IEnumerable<string> postfixes, stri
                         return !property.IsStatic() && shouldMapProperty(property);
                     case FieldInfo field:
                         return !field.IsStatic && shouldMapField(field);
+                    case ConstructorInfo constructor:
```
lbargaoanu: This doesn't belong here, constructors are not members to map.

src/AutoMapper/TypeMap.cs
```diff
@@ -184,7 +184,7 @@ public bool PassesCtorValidation()
 
             var constructors = DestinationTypeToUse
                 .GetDeclaredConstructors()
-                .Where(ci => !ci.IsStatic);
+                .Where(ci => !ci.IsStatic && Profile.ShouldMapCtor(ci));
```
lbargaoanu: This should just check Constructors in TypeDetails.

src/UnitTests/Bug/ShouldMapCtor.cs
```diff
@@ -0,0 +1,179 @@
+﻿using System.Linq;
+using Shouldly;
+using Xunit;
+
+namespace AutoMapper.UnitTests.Bug
+{
+    public abstract class ShouldMapCtor
```
lbargaoanu: Or you can simply rename the namespace from Bug to ShouldMapCtor and then this class is not useful anymore. The file should be one level up. The tests should work against the API. Either assert that some constructor is used when calling Map or assert that configuration validation fails.

src/AutoMapper/Configuration/IProfileConfiguration.cs
lxalln: Agreed, I used Ctor because other places in AM use Ctor in place of Constructor.

docs/Configuration.md
lxalln: I'll move it, I agree.

src/AutoMapper/ProfileMap.cs
lxalln: I'll see what I can do. I am assuming it's line endings, but no idea why it would affect only this file.

src/AutoMapper/TypeMap.cs
lxalln: It should have done that in the first place too 😄 I'll change it.

src/UnitTests/Bug/ShouldMapCtor.cs
lxalln: I'll move it.

I wasn't comfortable with these tests, but couldn't get past the way I'd tried to work around outside of AM. Asserting the configuration validation fails, should be fine. I'll do that.

src/AutoMapper/Configuration/IProfileConfiguration.cs
lbargaoanu: Yes, an oversight :)

src/AutoMapper/ProfileMap.cs
lxalln: I have force saved the file with CRLF line endings, which has produced a much smaller diff in GitHub.

I am guessing that this file has messed up line endings. Not sure what I can do beyond this. It's at least reviewable now, even if 90% of the changes are not really changes.

lbargaoanu: 👍

lxalln:
Right, we're stepping on each other's toes now. I'll leave this alone. I noticed the cosmetic changes too :)

Happy to squash it down if needs be.

lbargaoanu:
@jbogard I think ShouldUseConstructor makes more sense than ShouldMapConstructor.

lbargaoanu:
Sorry about the commits :) I thought it would easier to do it myself.

lxalln:
No problem.

I agree with `ShouldUseConstructor`, unfortunately there is already `DisableConstrutorMapping`, so does that mean we should stick with `Map` instead of `Use`?

lbargaoanu:
@lxalln Can you please do the rename, as I don't have VS handy?

lxalln:
Will do shortly

lbargaoanu:
I was wrong, this will still allow a constructor without parameters to be used, regardless of the condition. I'd say we want it to apply to those too, but I guess it's debatable.

lxalln:
@lbargaoanu The tests I put in, prove that an empty constructor cannot be used if it is being excluded by the filter. Am I missing something?

lbargaoanu:
If there is only the default constructor, it will be used regardless of the condition, right?

lxalln:
The default constructor is returned from `GetDeclaredConstructors` and will be filtered like any other constructor. If that means there are no constructors, then it will fail validation.

PR #2710, opened 2018-07-02
Restoring the array copy mapper; fixes #2706

### LINKED ISSUE #2706: Extremely bad Performance with .NET Native
We have a really big problem with Automapper and when the application is compiled in release mode with .NET Native and optimize for UWP Apps, Xamarin Forms 3.1.

We have a big application with three Profiles. Each Profile has a relatively big profile description.
In UWP with Debug mode the mapping takes about 21ms. 
In .NET Native the mapping takes about **9625ms**! That's 458 times worse! That's crazy.
No lines of code were changed!

The funny thing about it is, that this is always the same for some elements which have a Manual (the only thing i recognize in the moment). But for elements which doesn't have a Manual everything works okay.

But also in General. Automapper is much slower in .NET Native.

### Source/destination types

Source
```csharp
 internal class DeviceEntity : IAppEntity, IDisabledEntity {

        static DeviceEntity() {
            Triggers<DeviceEntity>.Inserting += entry => { entry.Entity.Modified = true; };
            Triggers<DeviceEntity>.Updating += entry => { entry.Entity.Modified = true; };
        }

        public string GuidDevice { get; set; }
        public int? IdDevice { get; set; }
        public int IdDeviceType { get; set; }
        public string GuidFunctionalArea { get; set; }
        public int? IdFunctionalArea { get; set; }
        public string Description { get; set; }
        public string Manufacturer { get; set; }
        public bool Disabled { get; set; }
        public DateTimeOffset? DisabledSince { get; set; }
        public byte[] Image { get; set; }
        public byte[] Manual { get; set; }
        public string ManualName { get; set; }
        public string InventoryNumber { get; set; }
        public double? DefinedSetpoint { get; set; }
        public bool TemperatureIndicator { get; set; }
        public bool Modified { get; set; }
        public DateTimeOffset? ModificationDate { get; set; }

        public DeviceTypeEntity DeviceType { get; set; }
        public FunctionalAreaEntity FunctionalArea { get; set; }
        public HashSet<QuestionAssignmentEntity> QuestionAssignments { get; set; } = new HashSet<QuestionAssignmentEntity>();
        public HashSet<ProductForDeviceEntity> ProductForDevices { get; set; } = new HashSet<ProductForDeviceEntity>();
        public HashSet<TaskEntity> Tasks { get; set; } = new HashSet<TaskEntity>();
        public HashSet<TaskMeasurementEntity> TaskMeasurements { get; set; } = new HashSet<TaskMeasurementEntity>();

        protected bool Equals(DeviceEntity other) {
            return string.Equals(GuidDevice, other.GuidDevice) && IdDevice == other.IdDevice;
        }

        public override bool Equals(object obj) {
            if (ReferenceEquals(null, obj)) {
                return false;
            }

            if (ReferenceEquals(this, obj)) {
                return true;
            }

            if (obj.GetType() != GetType()) {
                return false;
            }

            return Equals((DeviceEntity) obj);
        }

        public override int GetHashCode() {
            unchecked {
                return ((GuidDevice != null ? GuidDevice.GetHashCode() : 0) * 397) ^ IdDevice.GetHashCode();
            }
        }

    }
```
Destination
```csharp
  internal class DeviceModel : IDeviceModel {

        public string GuidDevice { get; set; }
        public int? IdDevice { get; set; }
        public int IdDeviceType { get; set; }
        public string GuidFunctionalArea { get; set; }
        public int? IdFunctionalArea { get; set; }
        public string Description { get; set; }
        public string Manufacturer { get; set; }
        public byte[] Image { get; set; }
        public byte[] Manual { get; set; }
        public string ManualName { get; set; }
        public string InventoryNumber { get; set; }
        public double? DefinedSetpoint { get; set; }
        public bool TemperatureIndicator { get; set; }
        public bool Disabled { get; set; }
        public DateTimeOffset? DisabledSince { get; set; }
        public bool Modified { get; set; }
        public DateTimeOffset? ModificationDate { get; set; }

        public int AnsweredQuestionCount { get; set; }
        public int TotalQuestionCount { get; set; }

        public IDeviceTypeModel DeviceType { get; set; }
        public IFunctionalAreaModel FunctionalArea { get; set; }
        public HashSet<IProductForDeviceModel> ProductForDevices { get; set; } = new HashSet<IProductForDeviceModel>();
        public HashSet<ITaskMeasurementModel> TaskMeasurements { get; set; } = new HashSet<ITaskMeasurementModel>();

    }
```
### Mapping configuration

```csharp
CreateMap<DeviceEntity, DeviceModel>().ForMember(x => x.AnsweredQuestionCount, opt => opt.Ignore())
                .ForMember(x => x.FunctionalArea, opt => opt.Ignore()).ForMember(x => x.TotalQuestionCount, opt => opt.Ignore())
                .ForMember(x => x.TaskMeasurements, opt => opt.Ignore()).ForMember(x => x.DeviceType, opt => opt.Ignore()).ReverseMap()
                .ForMember(x => x.DeviceType, opt => opt.Ignore()).ForMember(x => x.FunctionalArea, opt => opt.Ignore())
                .ForMember(x => x.ProductForDevices, opt => opt.Ignore()).ForMember(x => x.Tasks, opt => opt.Ignore())
                .ForMember(x => x.TaskMeasurements, opt => opt.Ignore()).ForMember(x => x.QuestionAssignments, opt => opt.Ignore());
            CreateMap<DeviceEntity, IDeviceModel>().As<DeviceModel>();
```
Initilization
```csharp

            Mapper.Initialize(cfg => {
                cfg.AllowNullCollections = true;
                cfg.CreateMissingTypeMaps = false;
                cfg.AddProfile<BusinessLogicMapperProfile>();
                cfg.AddProfile<DataMapperProfile>();
                cfg.AddProfile<UiMapperProfile>();
            });

            Mapper.Configuration.CompileMappings();
```
### Expected behavior

I would excpect that performance is nearly the same like in debug for release and .NET Native.

### Actual behavior

Really bad performance for release and .NET Native

jbogard:


lbargaoanu:
But why would this be the default behavior? It doesn't really map and other than .Net Native, nobody complained about performance.

jbogard:
No, this is strictly for the case where we have primitive array -> primitive array. It's just way more efficient than the default behavior, which is a for loop. I don't know if I can think of a reason why this *shouldn't* be the default behavior for primitive arrays, since it's a readily available method to copy arrays.

lbargaoanu:
https://github.com/AutoMapper/AutoMapper/pull/1714#issuecomment-255756462 It's not like you copy 1 million items arrays on a regular basis. We're adding a special case for a corner scenario.

jbogard:
We could detect that you've actually mapped the items. So Array.Copy wouldn't work in cases where you've defined some custom mapping function for two primitives with assignable types. When `Array.Copy` could work by itself, but you've defined some custom matching thing for it.

lbargaoanu:
A mapper might also apply, built in or not.

jbogard:
An `IObjectMapper`? Yeah I guess there’s not an easy way to detect that.

lbargaoanu:
https://github.com/AutoMapper/AutoMapper/blob/28c920fbc7f82ef21bca59e8741ba2d719ecebce/src/AutoMapper/IConfigurationProvider.cs#L108

lbargaoanu:
But chances are, a bunch of mappers would apply, so maybe it gets hairy :)

jbogard:
Digging a little more, it's not easy to tell 1) if we definitely don't have a mapper without going through the entire evaluation process, excluding *some* mappers (like the `ConvertMapper`). And it's also not trivial to see if the two elements types "can" be copied over - it's an [internal call](https://github.com/dotnet/coreclr/blob/master/src/System.Private.CoreLib/src/System/Array.cs#L238).

I wonder, is this something we could explicitly document somehow?

lbargaoanu:
Maybe it would make more sense to only use it when the types are exactly the same.

PR #2720, opened 2018-07-17
Mapper.Initialize now can be called multiple times

jbogard:
With additive type maps.

### DIFF
--- a/src/AutoMapper/IConfigurationProvider.cs
+++ b/src/AutoMapper/IConfigurationProvider.cs
@@ -172,5 +172,11 @@ namespace AutoMapper
         /// <param name="mapRequest">The source/destination map request</param>
         /// <returns>the execution plan</returns>
         LambdaExpression BuildExecutionPlan(MapRequest mapRequest);
+
+        /// <summary>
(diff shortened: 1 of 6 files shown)


jbogard:
The configuration was never read-only. Between open generics, convention type maps, and dynamically created maps, we've needed to have our type map list dynamic. So the capability was always there at least.

This also doesn't make Initialize calls commutative. Once the initial...initialization happens, subsequent calls only create new type maps.

Overall, I think we need to enable `services.AddAutoMapper` to be called multiple times, even with static registration. Maybe that's putting a lock/Lazy inside of that code?

lbargaoanu:
Yes, but that's different. All that code does is CreateTypeMap and it takes one global lock in the end, that's why we don't see concurrency issues anymore. A static AddAutoMapper could replace the static instance.

jbogard:
I’ll do that then. Make AddAutoMapper idempotent, for the static route at
least.

jbogard:
Most of the problems seem to be in ASP.NET Core, so I just removed static registration altogether:

https://github.com/AutoMapper/AutoMapper.Extensions.Microsoft.DependencyInjection/pull/58

PR #2741, opened 2018-08-03
fix PropertiesWithAnInaccessibleSetter

### LINKED ISSUE #2740: IgnoreAllPropertiesWithAnInaccessibleSetter not working with base classes
### Source/destination types

```csharp
    public class Source
    {
        public int IgnoreMe { get; set; }
    }

    public class Destination : DestinationBase
    {
    }

    public class DestinationBase
    {
        public int IgnoreMe { get; private set; }
    }
```

### Mapping configuration

```csharp
            {
                cfg.CreateMap<Source, Destination>()
                    .IgnoreAllPropertiesWithAnInaccessibleSetter();
            });
```

### Expected behavior

The `IgnoreMe` member of the `Destination` object should be ignored when mapping.

### Actual behavior

The `IgnoreMe` member of the `Destination` object is *not* ignored and mapped without issues.

### Steps to reproduce

https://gist.github.com/JoesGab/48f136622234c9ef0ab6f26eb164d449

Note the `DestinationBase`! The test only fails in this case, passes if you put the `IgnoreMe` in the derived type.

JoesGab:
Fix PropertiesWithAnInaccessibleSetter so that it also respects inherited properties, not only declared ones.
Fixes #2740.

### DIFF
--- a/src/AutoMapper/TypeExtensions.cs
+++ b/src/AutoMapper/TypeExtensions.cs
@@ -81,7 +81,7 @@ namespace AutoMapper
 
         public static IEnumerable<PropertyInfo> PropertiesWithAnInaccessibleSetter(this Type type)
         {
-            return type.GetDeclaredProperties().Where(pm => pm.HasAnInaccessibleSetter());
+            return type.GetRuntimeProperties().Where(pm => pm.HasAnInaccessibleSetter());
         }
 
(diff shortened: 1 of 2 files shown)


lbargaoanu:
@jbogard Merge?

PR #2751, opened 2018-08-08
System.ValueTuple is included from 4.7 on

lbargaoanu:
Actually I'm not sure what should be done here, if anything.


jbogard:
I think this is right - though I don't know if we'd want to create the additional .NET 4.7 target?

The more I look at this, the more it seems that going just with `netstandard2.0` would be best.

lbargaoanu:
I'd say we don't need another target :) I think what we have now is a reasonable compromise. Ditching a full framework target would mean losing Reflection.Emit. I'm not sure what else. But I think the full framework is important enough to keep a dedicated target.

PR #2759, opened 2018-08-13
Fix #2753

### LINKED ISSUE #2753: Substitute doesn't work as suggested
For a repro (a failing unit test), see #2752.

From the docs on the `Substitute` method, when mapping a `Parent` to a sequence of `ChildDto`s, I would expect the original source (a `Parent`) to be substituted by it's children (`IEnumerable<Child>`) which would then be mapped to an `IEnumerable<ChildDto>` (since there is a mapping for `Child` to `ChildDto`). However, the test fails.

It seems that `TSubstitution` is expected to inherit from `TSource` which the type constraints on `Substitute` don't enforce.

If this is not a bug, what is the real purpose of `Substitute`? There is little documentation for it.

danielcweber:
This fixes #2753. The docs on `Substitute` say: "Replace the original runtime instance with a new source instance. [...] The returned source object will be mapped instead of what was supplied in the original source object."

This would only work if `TSubstitute` would cast to `TDestination`. This fix enables the feature for arbitrary `TDestination`s.

### DIFF
--- a/src/UnitTests/Bug/ObjectSubstitutionFailure.cs
+++ b/src/UnitTests/Bug/ObjectSubstitutionFailure.cs
@@ -1,0 +1,46 @@
+﻿using System.Collections.Generic;
+
+namespace AutoMapper.UnitTests.Bug
+{
+    namespace ObjectSubstitutionFailure
+    {
+        using Shouldly;
+        using Xunit;
+
+        public class Parent
+        {
+            public IEnumerable<Child> Children => new[] { new Child() };
+        }
+
+        public class Child
+        {
+
+        }
+
+        public class ChildDto
+        {
+
+        }
+
+        public class OverrideExample : AutoMapperSpecBase
+        {
+            protected override MapperConfiguration Configuration { get; } = new MapperConfiguration(cfg =>
+            {
+                cfg
+                    .CreateMap<Parent, IEnumerable<ChildDto>>()
+                    .Substitute(x => x.Children);
+
+                cfg
+                    .CreateMap<Child, ChildDto>();
+            });
+
+            [Fact]
+            public void Should_substitute_correct_object()
+            {
+                Mapper
+                    .Map<IEnumerable<ChildDto>>(new Parent())
+                    .ShouldNotBeNull();
+            }
+        }
+    }
+}
--- a/src/AutoMapper/Execution/TypeMapPlanBuilder.cs
+++ b/src/AutoMapper/Execution/TypeMapPlanBuilder.cs
@@ -28,6 +28,9 @@ namespace AutoMapper.Execution
         private static readonly Expression<Func<ResolutionContext, int>> GetTypeDepthInfo =
             ctxt => ctxt.GetTypeDepth(default);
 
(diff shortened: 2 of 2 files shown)


src/UnitTests/Bug/ObjectSubstitutionFailure.cs
```diff
@@ -0,0 +1,46 @@
...
+
+        }
+
+        public class OverrideExample : AutoMapperSpecBase
+        {
+            protected override MapperConfiguration Configuration { get; } = new MapperConfiguration(cfg =>
+            {
+                cfg
+                    .CreateMap<Parent, IEnumerable<ChildDto>>()
+                    .Substitute(x => x.Children);
```
lbargaoanu: But how is this useful? Why not simply map the children? Also without a type converter this map doesn't make sense. The docs say "Useful when ORMs return proxy types with no relationships to runtime types." So clearly the original intent is lost. But the new one is not clear at all.

lbargaoanu: You can create a type converter for that, you don't need a separate feature, you could even implement Substitute that way, outside AM.

danielcweber: This is tedious. I'm trying to make a feature work that is already present but seemingly broken. I also provide a fix because I like to contribute.

You. Are. Not. Helpful. At. All.

Close it or let @jbogard weigh in. If it's doable with a type converter, I will try figure out how by myself.

lbargaoanu: Yes, it's broken, but any fix should also make sense, right? :)

danielcweber: Then go ahead and define what's "sense".

What didn't make sense was the error I got when I tried to use `Substitute` with a `TSubstitute` that didn't inherit from `TDestination`, because that was not obvious, the docs didn't say it, the generic type constraints didn't suggest it. Put sense into it, I will follow.

lbargaoanu: My best guess is that Substitute made sense at a point when Include didn't work properly. The existing test works now without Substitute, so that's broken too, as an older issue proved. So this seems to me like a feature in search of an use case. But I for one don't see the use case yet. I just see two tests that don't need the feature. Removing it would lead to one of two things I believe. A real use case, and then we can add it back in and fix it. Or a feature less.

jbogard: For this specific test, we'd want people to use a type converter, that's what it's for. To say "use this other way of mapping instead" and that would also apply globally. The original test: https://github.com/AutoMapper/AutoMapper/blob/859612d672a45c62f3cce5c8b236be796497c916/src/UnitTests/ObjectSubstitution.cs is all about weird proxy objects from ORMs.

Since this test is quite a bit different than the existing test, and the underlying configuration can be solved in other ways, I want to make sure the examples we have in tests adequately represent the design intent of the feature.

Which unfortunately I don't have a great grasp of, it was wonky NHibernate proxy junk.

PR #2778, opened 2018-08-28
AssertConfigurationIsValid(profileName) should throw if name does not exist

### LINKED ISSUE #2777: AssertConfigurationIsValid(profileName) should throw if name does not exist
I am using `AssertConfigurationIsValid(string profileName)` to try and run a specific profile. You can see how this can be problematic below:

```csharp
        [Theory]
        [InlineData("Blah Blah I dont actually exist")]
        [InlineData("CustomerProfile")]
        public void Product_Configuration_Is_Valid(string profileName)
        {
            using (var scope = _factory.Server.Host.Services.CreateScope())
            {
                // Test passes!!!! This function should throw if the profile name does not exist.
                scope.ServiceProvider.GetRequiredService<IConfigurationProvider>()
                    .AssertConfigurationIsValid(profileName);
            }
        }
```

Maybe the key is not to modify code in the title method but to do it in `AssertConfigurationIsValid(string profileName)` when it tries to do `_typeMapRegistry.Values.Where(typeMap => typeMap.Profile.Name == profileName)`. In there we just need a line to check for `!Any` and throw.

VictorioBerra:
Fixes #2777.

### DIFF
--- a/src/AutoMapper/MapperConfiguration.cs
+++ b/src/AutoMapper/MapperConfiguration.cs
@@ -278,7 +278,14 @@ namespace AutoMapper
 
         public void AssertConfigurationIsValid(string profileName)
         {
-            _validator.AssertConfigurationIsValid(_typeMapRegistry.Values.Where(typeMap => typeMap.Profile.Name == profileName));
+            var typeMaps = _typeMapRegistry.Values.Where(typeMap => typeMap.Profile.Name == profileName);
+
+            if (!typeMaps.Any())
+            {
+                throw new ArgumentOutOfRangeException(nameof(profileName), $"Cannot find any types for {profileName}.");
+            }
+
+            _validator.AssertConfigurationIsValid(typeMaps);
         }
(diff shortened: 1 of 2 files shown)


src/AutoMapper/MapperConfiguration.cs
```diff
@@ -277,8 +277,15 @@ public void AssertConfigurationIsValid(TypeMap typeMap)
         }
 
         public void AssertConfigurationIsValid(string profileName)
-        {
-            _validator.AssertConfigurationIsValid(_typeMapRegistry.Values.Where(typeMap => typeMap.Profile.Name == profileName));
+        {
+            var typeMaps = _typeMapRegistry.Values.Where(typeMap => typeMap.Profile.Name == profileName);
+
+            if (!typeMaps.Any())
```
lbargaoanu: It seems better to simply check if the profile exists. The Profiles property.

PR #2804, opened 2018-09-23
Drop netstandard13

jbogard:
And rev up to `net461` based on MS recommendation. Basically, `net461` has an issue with `netstandard2.0`, so we should target it explicitly.

This should simplify our dependencies, and fix folks that have to do weird binding redirects with `ValueTuple`.


lbargaoanu:
We don't support 4.5 anymore.
4.7 projects  will use the 4.6.1 target and they will still have a reference to ValueTuple.
Platforms that don't support Reflection.Emit will have a reference to it.
Problematic  I'd say.

jbogard:
Here's the recommendation: https://twitter.com/terrajobst/status/1040410118053982208

lbargaoanu:
I guess 4.5 is old and the value tuple thing is a framework problem. So replacing 4.5 with 4.6.1 might not be a big deal. But I don't know how mobile platforms wil handle Reflection.Emit.

jbogard:
@lbargaoanu They won't. They'll just get a `PlatformNotSupportedException`. Not much to do about it, basically the .NET team restored the reflection package back into `netstandard2.0` with the expectation that you just don't use those features in platforms that don't support it. I don't really want to create targets for all the mobile platforms since all we'd do is throw our own exception in those cases.

lbargaoanu:
Reflection.Emit is part of [.Net Standard 2.1](https://blogs.msdn.microsoft.com/dotnet/2018/11/05/announcing-net-standard-2-1/).

PR #2806, opened 2018-09-25
Value converters

### LINKED ISSUE #2773: Proposal: Member-Scoped Type Converters
I'm in the middle of upgrading an old application that uses 2.2.1, and the old value formatters pretty heavily.

Those were a bad idea, for a number of reasons, but I do find myself wanting the ability to apply a type converter to an individual member.

We can apply type converters to a TypeMap level, but *not* to an individual member. The reason this is useful is I would want to apply some sort of conversion, but only on specific members. It would look something like:

```c#
cfg.CreateMap<Foo, FooDto>()
    .ForMember(dest => dest.Bar, opt => opt.ConvertUsing<BarValueConverter>())
    .ForMember(dest => dest.Baz, opt => opt.MapFrom(src => src.Buzz).ConvertUsing<BazConverter>());
```

The mapping operation would be similar to other type converters, except it's localized to a single member.

jbogard:
Fixes #2773 

This adds a new interface:

```c#
public interface IValueConverter<in TSourceMember, out TDestinationMember>
{
    TDestinationMember Convert(TSourceMember sourceMember, ResolutionContext context);
}
```

Value converters are a cross between type converters and value resolvers. Because value converters are only scoped to members, this makes it much, much easier to reuse them across maps. These are the effective replacement for the old value formatters, minus the weird stacking of value formatters.

Unlike type converters, value converters are just another kind of value resolver, rather than something that completely overrides every other option. The precedence is:

 - Value converter
 - Value resolver
 - Resolving func
 - MapFrom expression
 - Member matching

I thought this simplified things rather than having a one-off for value converters.

Open question - should the value converters also receive the destination member, like the type converter, value resolver, and member value resolver? Right now it doesn't, but it'd be trivial to add it. If I did, it'd lose the variance for that parameter (not sure that's an issue).

Once this gets released, we'll need to update the DI project to scan for these types for registration.

### DIFF
--- a/src/AutoMapper/Configuration/MappingExpression.cs
+++ b/src/AutoMapper/Configuration/MappingExpression.cs
@@ -128,28 +128,11 @@ namespace AutoMapper.Configuration
                 PropertyMapActions.Add(pm => pm.ValueResolverConfig = config);
             }
 
-            public void ConvertUsing(Type valueConverterType)
-            {
-                PropertyMapActions.Add(pm =>
-                {
-                    var config = new ValueConverterConfiguration(valueConverterType, valueConverterType.GetGenericInterface(typeof(IValueConverter<,>)));
+            public void ConvertUsing(Type valueConverterType) 
+                => PropertyMapActions.Add(pm => ConvertUsing(pm, valueConverterType));
 
-                    pm.ValueConverterConfig = config;
-                });
-            }
-
-            public void ConvertUsing(Type valueConverterType, string sourceMemberName)
-            {
-                PropertyMapActions.Add(pm =>
-                {
-                    var config = new ValueConverterConfiguration(valueConverterType, valueConverterType.GetGenericInterface(typeof(IValueConverter<,>)))
-                    {
-                        SourceMemberName = sourceMemberName
-                    };
-
-                    pm.ValueConverterConfig = config;
-                });
-            }
+            public void ConvertUsing(Type valueConverterType, string sourceMemberName) 
+                => PropertyMapActions.Add(pm => ConvertUsing(pm, valueConverterType, sourceMemberName));
 
             public void ConvertUsing<TSourceMember, TDestinationMember>(IValueConverter<TSourceMember, TDestinationMember> valueConverter, string sourceMemberName)
             {
@@ -162,6 +145,16 @@ namespace AutoMapper.Configuration
 
                     pm.ValueConverterConfig = config;
                 });
+            }
+
+            private static void ConvertUsing(PropertyMap propertyMap, Type valueConverterType, string sourceMemberName = null)
+            {
+                var config = new ValueConverterConfiguration(valueConverterType, valueConverterType.GetGenericInterface(typeof(IValueConverter<,>)))
+                {
+                    SourceMemberName = sourceMemberName
+                };
+
+                propertyMap.ValueConverterConfig = config;
             }
         }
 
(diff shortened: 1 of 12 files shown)


jbogard:
Actually the more I think about it the more I would expect the destination value to be passed in, for consistency if nothing else.

src/AutoMapper/PropertyMap.cs
```diff
@@ -79,20 +80,11 @@ public MemberInfo SourceMember
...
-                    return CustomExpression.ReturnType;
-                if (CustomResolver != null)
-                    return CustomResolver.ReturnType;
-                if(ValueResolverConfig != null)
-                    return typeof(object);
-                return SourceMember?.GetMemberType();
-            }
-        }
-
+        public Type SourceType => CustomExpression?.ReturnType
```
lbargaoanu: These should be ordered by priority I guess.

src/UnitTests/ValueConverters.cs
```diff
@@ -0,0 +1,588 @@
+﻿using Shouldly;
+using Xunit;
+
+namespace AutoMapper.UnitTests
+{
+    public class ValueConverters
```
lbargaoanu: Maybe a generic test with ForAllMembers or ForAllPropertyMaps?

src/AutoMapper/Configuration/MappingExpression.cs
```diff
@@ -127,6 +127,42 @@ public void ResolveUsing(Type valueResolverType, string memberName)
 
                 PropertyMapActions.Add(pm => pm.ValueResolverConfig = config);
             }
+
+            public void ConvertUsing(Type valueConverterType)
+            {
+                PropertyMapActions.Add(pm =>
+                {
+                    var config = new ValueConverterConfiguration(valueConverterType, valueConverterType.GetGenericInterface(typeof(IValueConverter<,>)));
```
lbargaoanu: Not much reuse, but you can add another constructor for these two cases.

src/AutoMapper/Configuration/MemberConfigurationExpression.cs
```diff
@@ -284,9 +283,88 @@ public void SetMappingOrder(int mappingOrder)
             PropertyMapActions.Add(pm => pm.MappingOrder = mappingOrder);
         }
 
+        public void ConvertUsing<TValueConverter, TSourceMember>()
+            where TValueConverter : IValueConverter<TSourceMember, TMember>
+        {
+            PropertyMapActions.Add(pm =>
```
lbargaoanu: You could have a common function that calls PropertyMapActions.Add and sets the config. Small, but useful :)

lbargaoanu:
Reusability is the selling point, so variance might matter. Maybe a little messy, but what would differentiate them even more from other resolvers would be some way to skip the final mapping step, as you can already assign the resolved value to the destination. That would fit with how type converters work.

src/AutoMapper/PropertyMap.cs
jbogard: Modified to match `TypeMapPlanBuilder`

src/UnitTests/ValueConverters.cs
jbogard: Added test

src/AutoMapper/Configuration/MappingExpression.cs
jbogard: I figured that the source member/names are optional so we'd need like 6 ctors to get every permutation

lbargaoanu: I was thinking one extra constructor that takes only the converter type. Nitpicking.

src/AutoMapper/Configuration/MemberConfigurationExpression.cs
jbogard: Good call. Looks like there are many overloads in these classes that could use this refactoring.

Issue #2795, opened 2018-09-15
Remove ConstructProjectionUsing 


lbargaoanu:
Replace it with expression based ConstructUsing. It seems like a good time, as we've already bumped the version. But I guess it's debatable. Why wouldn't the same apply to ProjectUsing?

jbogard:
The general argument against is that not every ConstructUsing can be represented as an expression.

Another option would be to remove the ConstructProjectionUsing (really any expression-specific method) and instead have an overload:

```c#
        IMappingExpression<TSource, TDestination> ConstructUsing(Expression<Func<TSource, TDestination>> ctor);
        IMappingExpression<TSource, TDestination> ConstructUsing(Func<TSource, ResolutionContext, TDestination> ctor);
```

If you want to use a complex `Func`, you need to use the more complex overload. I'd like to go one step further and remove `ResolveUsing` but those have an additional complication in that their results are mapped.

lbargaoanu:
Yes, but I was trying to say, ProjectUsing vs ConvertUsing, isn't it the same?

jbogard:
Ah, yes. Those two as well.

jbogard:
So would we remove the "ResolveUsing" overloads that just map via `Func` to collapse to `MapFrom`? Save `ResolveUsing` for actual resolvers?

lbargaoanu:
You mean MapFrom(Func), I suppose (not MapFrom(Expression)). That makes sense, but I think it's too big of a change. I guess we could provide some built in extension methods, but that would defeat the purpose as most people would probably use those.

lbargaoanu:
Updated with ResolveUsing.

lbargaoanu:
But the problem here is that we might break a lot of  projections when going from Func to Expression. What used to apply only for Map will apply for ProjectTo too now.

jbogard:
I'm not suggesting breaking `ProjectTo`, but this will break some `Func` based versions of the overloads. Mainly, if you're using a `Func`-based version that's getting replaced with `Expression`, you *must* use expressions now. Whether or not that expression can be used by `ProjectTo` is different.

What this will explicitly prevent though is people wanting to use different configurations for `ProjectTo` and `Map`, which is fine by me.

If you have a complex lambda statement that can't be converted to an expression, you'll just need to use the overloads that allow `Func`.

jbogard:
Also, I'm fine with the `ResolveUsing` changes. That makes sense - those overloads use an object rather than func.

[...]

PR #2809, opened 2018-09-26
Consolidate ResolveUsing to MapFrom

jbogard:
Per #2795 discussion


lbargaoanu:
A small thing, but I wouldn't mind seeing UseValue gone :)

jbogard:
Ha, oh right! Because we used to force people to only use `MapFrom` from properties.

I'd be OK with that,

lbargaoanu:
OK, it should be simple to do. I'll open a PR whenever :)

jbogard:
Too late lol

lbargaoanu:
Nice :)

yhnavein:
Ok, fantastic. I have replaced `ResolveUsing` with the `MapFrom`, but how should I deal nicely with the `An expression tree lambda may not contain a null propagating operator` error?
That was the whole purpose of using ResolveUsing for me, as it was short and dandy. And I have plenty of null checking in my mappings.

```diff
 -.ForMember(dest => dest.UserId, opts => opts.ResolveUsing(src => src.User?.Id))
 +.ForMember(dest => dest.UserId, opts => opts.MapFrom(src => src.User?.Id))
```

I find the following solution unacceptable:
`.ForMember(dest => dest.UserId, opts => opts.MapFrom(src => src.User == null ? (long?) null : src.User.Id))`

Is there any other way to deal with this breaking change?

lbargaoanu:
Check [the upgrade guide](http://docs.automapper.org/en/latest/8.0-Upgrade-Guide.html).

PR #2810, opened 2018-09-27
Consider ForPath when checking for cycles

### LINKED ISSUE #2813: Mapping with ForPath causes StackOverflowException
https://stackoverflow.com/questions/52503102/why-is-forpath-causing-a-stackoverflowexception-and-formember-is-not?s=19|0.0000

lbargaoanu:
Fixes #2813.

### DIFF
--- a/src/AutoMapper/Execution/TypeMapPlanBuilder.cs
+++ b/src/AutoMapper/Execution/TypeMapPlanBuilder.cs
@@ -93,7 +93,8 @@ namespace AutoMapper.Execution
             typeMapsPath.Add(_typeMap);
             var defaultPropertyMap = new PropertyMap(default(MemberInfo), null);
             var properties = 
-                _typeMap.GetPropertyMaps().Where(pm=>pm.CanResolveValue()).Select(pm=>new { PropertyTypeMap = ResolvePropertyTypeMap(pm), PropertyMap = pm })
+                _typeMap.GetPropertyMaps().Concat(_typeMap.PathMaps.Select(pm => new PropertyMap(pm))).Where(pm=>pm.CanResolveValue())
+                .Select(pm=>new { PropertyTypeMap = ResolvePropertyTypeMap(pm), PropertyMap = pm })
                 .Concat(_typeMap.ConstructorMappingTypes.Select(tp => new { PropertyTypeMap = ResolveTypeMap(tp), PropertyMap = defaultPropertyMap }))
                 .Where(p => p.PropertyTypeMap != null && !p.PropertyTypeMap.PreserveReferences && p.PropertyTypeMap.MapExpression == null);
             foreach(var property in properties)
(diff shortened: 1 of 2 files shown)


jbogard:
Is there an issue this fixes? Or just found this separately?

lbargaoanu:
No issue, from SO. The test should crash on master. With SO :)

jbogard:
Ha! Can you create a GH issue with the SO link? Makes the release notes a bit easier.

src/AutoMapper/Execution/TypeMapPlanBuilder.cs
```diff
@@ -93,8 +93,9 @@ private void CheckForCycles(HashSet<TypeMap> typeMapsPath)
             typeMapsPath.Add(_typeMap);
             var defaultPropertyMap = new PropertyMap(default(MemberInfo), null);
             var properties = 
-                _typeMap.GetPropertyMaps().Where(pm=>pm.CanResolveValue()).Select(pm=>new { PropertyTypeMap = ResolvePropertyTypeMap(pm), PropertyMap = pm })
-                .Concat(_typeMap.ConstructorMappingTypes.Select(tp => new { PropertyTypeMap = ResolveTypeMap(tp), PropertyMap = defaultPropertyMap }))
+                _typeMap.GetPropertyMaps().Concat(_typeMap.PathMaps.Select(pm => new PropertyMap(pm))).Where(pm=>pm.CanResolveValue())
```
jbogard: This is the second place we're wrapping `PathMap` with `PropertyMap`, I wonder if there's maybe a common interface or base class between the two? Then those could be exposed up at `TypeMap`?

lbargaoanu: Yes, but I don't think that would really help. Because when building the execution plan (the other place) we really want the whole PropertyMap interface (and PathMap implements just a few things). Here probably that interface could be useful, but it's just one place at the moment.

lbargaoanu: Probably null would work here as well :)

lbargaoanu: No, I need the type map.

jbogard: Do you think there's anything to do with the two places in that same class where we're creating PropertyMaps from PathMaps? I guess it seems weird to me we're doing it twice.

lbargaoanu: Do we want to rewrite CreatePropertyMapFunc so it works both with the common interface and the full property map interface? Or have a more complete interface of which PathMap implements a few methods? I think that's weird too. Doable, but...

jbogard: Meh. Just leave it for now.

PR #2814, opened 2018-10-01
Set ignored in the property map constructed from a path map

lbargaoanu:
If you think the right fix is to introduce a complete interface implemented by both, I can do that. I wasn't convinced, so I didn't go that way.


jbogard:
Nah I'm not convinced either

jbogard:
Is this for another SO question?

lbargaoanu:
No, I figured this one out by myself :)

PR #2820, opened 2018-10-05
Contextual exceptions for path maps

jbogard:
No longer restricted to property maps. Changes the exception object to include `IMemberMap` instead of just `PropertyMap`


lbargaoanu:
I guess the same applies to constructor maps.

jbogard:
We might have to tweak the parameter mapping then

PR #2843, opened 2018-10-22
Check for null reference

aeslinger0:
I kept getting a null reference exception in unit tests when the configuration wasn't supplied to the mapper. (I'm migrating from the static methods to the instance-based ones). This should give a more meaningful message.


jbogard:
@aeslinger0 CLA check :(

PR #2850, opened 2018-10-26
Added example for Castle Windsor

Gadensgaard:
Added example from Castle Windsor. 
Removed "Gotchas" and added a section describing how to use QueryableExtensions with DI.

### DIFF
--- a/docs/Dependency-injection.md
+++ b/docs/Dependency-injection.md
@@ -21,11 +21,11 @@ var dest = mapper.Map<Source, Destination>(new Source { Value = 15 });
 
 ### QueryableExtensions
 
-When using DI with [QueryableExtensions](Queryable-Extensions.html), remember to install the IConfigurationProvider in the the IoC container.
+When using DI with [QueryableExtensions](Queryable-Extensions.html), remember to install the IConfigurationProvider in the IoC container.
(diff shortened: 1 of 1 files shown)


docs/Dependency-injection.md
```diff
@@ -19,9 +19,11 @@ var mapper = new Mapper(Mapper.Configuration, childContainer.GetInstance);
 var dest = mapper.Map<Source, Destination>(new Source { Value = 15 });
 ```
 
-#### Gotchas
+#### QueryableExtensions
 
-Using DI is effectively mutually exclusive with using the IQueryable.ProjectTo extension method.  Use ``` IEnumerable.Select(_mapper.Map<DestinationType>).ToList() ``` instead.
```
lbargaoanu: I think what this sentence is trying to say is that ProjectTo doesn't support DI, value resolvers and type converters, because it's just an expression to be interpreted by the EF provider. So you might want to make that more clear, not replace it.

Gadensgaard: But that is already documented under QueryableExtensions, and has nothing to do with DI?

lbargaoanu: That's not relevant. People make mistakes and this sentence is trying to prevent that.

Gadensgaard: Added addional comment, with link to documentation.

PR #2866, opened 2018-11-09
Added IMapper.ProjectTo

### LINKED ISSUE #2857: IMapper.ProjectTo
It would certainly make sense for the people using DI. It fits with the move away from static. It would blur a little more the difference between Map and ProjectTo, but I guess we can live with that. We would probably have to keep the existing extensions, we do have a static Mapper class after all. So we would have two ways of doing the same thing. Not great, but maybe it's worth it.

lbargaoanu:
Fixes #2857. I omitted the overloads with two parameters. I think we can live without them. People can easily add them, but I guess we could have them built in as extension methods.


lbargaoanu:
The upgrade guide?

jbogard:
Nah the page on queryable extensions

Issue #2882, opened 2018-11-18
Bug after switching to V8


jogibear9988:
Following Code wored in Automapper 7:

       IMapper mapper;
            mappers.TryGetValue(key, out mapper);
            if (mapper == null)
            {
                var config = new MapperConfiguration(cfg => {
                    cfg.CreateMap(source.GetType(), destination.GetType());
                    cfg.AllowNullCollections = true;
                    cfg.AllowNullDestinationValues = true;
                });
                mapper = config.CreateMapper();
                mappers.Add(key, mapper);
            }

now I get the following Exception:

![automapper](https://user-images.githubusercontent.com/364896/48679628-f0a68e80-eb92-11e8-970f-099a874f1b04.png)

stacktrace:

![automapper2](https://user-images.githubusercontent.com/364896/48679630-f7cd9c80-eb92-11e8-85be-08793db1ec2c.png)

calling code:

![automapper3](https://user-images.githubusercontent.com/364896/48679631-fdc37d80-eb92-11e8-8604-8166645bdf03.png)

lbargaoanu:
A repro would help. Make a [gist](https://gist.github.com/lbargaoanu/9c7233441c3a3413cc2b9b9ebb5964a9) that we can execute and see fail.

jogibear9988:
I've created a Test. It's a issue using Automapper with Castle Dyanmic Proxy  Classes

PR #2884, opened 2018-11-19
Test for issue #2882

jogibear9988:


lbargaoanu:
I see, but that doesn't help much, because we're not going to reference Castle.Core. We need a repro without it. What is it about that proxy that causes the issue?

jogibear9988:
I don't know what is the issue. 
Caste Core is only referenced by the tests, not by the main package. 

But if we find out why it does not work, maybe I can create a test without castle core. Can you help with that? Code worked in v7 without problems. And I think it's a common scenario to use automapper with proxy objects

lbargaoanu:
I have no idea why that proxy is special. Once we have a repro without Castle, I'll take a look.

jogibear9988:
Can't you fetch my fork and run the code? I've no Idea where to look!

jogibear9988:
image

Parameter.IsOptional is false,
Parameter.ParameterType is of IInterceptor[]
SourceMebers List is empty

jogibear9988:
Proxy Type, has a Field of Name "__interceptors"

image

jogibear9988:
@lbargaoanu I think I found the issue.

Automapper creates a ConstructorParameter without a Name. And your Code in 
    
     if (string.IsNullOrEmpty(nameToSearch))
                return true;

Returns true, so a it says a matching member is found (but don't know why it worked in V7)

lbargaoanu:
I'm not sure I understand. But if you do, a PR is welcome. Of course that needs a failing test without Castle.

jogibear9988:
Failing test without castle...

lbargaoanu:
OK, so you mean Castle creates a ConstructorParameter without a name. You wrote AM and that doesn't make much sense.

jogibear9988:
Yes, castle creates a Constructor Parameter without a Name.

Look at my updated Pull Req. It now does not use Castle.Core Any more and fails with the same error!

lbargaoanu:
But you do know that AM can create proxies for you, including implementing INotifyPropertyChanged.

jogibear9988:
I've seen it. But this is existing Code, and I cannot change it atm!
And it worked with the previous Version, so I think we should look that the new one does not break existing Code.
Is there anything more I could test/help?

lbargaoanu:
This is not the first time this happens. And I'm pretty sure it will happen again. I mean, parameters without names? Really? :) I've opened a PR, but this is not a good place to be. I cringe when I look at that test.

jogibear9988:
Why should a Name in generated Code be needed?
I only map to existing Instances created by Castle Dynmaic Proxy, so Automapper don't need to map anything of the Constructor.
Is there no way, so I could tell that it should only Map Proiperties? (I thought this is/was the default)

jogibear9988:
And shouldn't in DefaultMember the lines

            if (string.IsNullOrEmpty(nameToSearch))
                return true;

also be removed?

lbargaoanu:
You can disable constructor mapping (by default it's enabled). There is a method on the configuration. I've tried removing that code in your branch and it didn't go well :) That's why I started again from scratch.

PR #2887, opened 2018-11-19
A constructor parameter with no name won't match anything

### LINKED ISSUE #2882: Bug after switching to V8

lbargaoanu:
Fixes #2882. Closes #2884. Maybe I should drop that test and make some things public so I can write a white box test.

### DIFF
--- a/src/AutoMapper/ReflectionExtensions.cs
+++ b/src/AutoMapper/ReflectionExtensions.cs
@@ -53,9 +53,6 @@ namespace AutoMapper
         public static IEnumerable<TypeInfo> GetDefinedTypes(this Assembly assembly) =>
             assembly.DefinedTypes;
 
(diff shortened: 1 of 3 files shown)


jbogard:
@lbargaoanu it might not hurt to make some more of the internal type stuff public for testing purposes. Or we can do the internals visible to trick

PR #2892, opened 2018-11-22
Added methods BeforeMap and AfterMap for IMappingExpression with ResolutionContext

robin-thoni:
This allows to use `ResolutionContext` when callbacks are added in `MapperConfigurationExpression.ForAllMaps()`, when only the non-templated `IMappingExpression` is available.

I guess the line encoding is buggy, so that's why the change is so big for `MappingExpression.cs`, but only lines 58-60 and 67-69 were added.

### DIFF
--- a/src/UnitTests/BeforeAfterMapping.cs
+++ b/src/UnitTests/BeforeAfterMapping.cs
@@ -33,6 +33,30 @@ namespace AutoMapper.UnitTests.BeforeAfterMapping
             afterMapCalled.ShouldBeTrue();
         }
 
+        [Fact]
+        public void Before_and_After_overrides_should_be_called()
+        {
+            var beforeMapCalled = false;
+            var afterMapCalled = false;
+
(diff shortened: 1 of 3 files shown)


src/UnitTests/BeforeAfterMapping.cs
```diff
@@ -33,6 +33,24 @@ public void Before_and_After_should_be_called()
             afterMapCalled.ShouldBeTrue();
         }
 
+        [Fact]
+        public void Before_and_After_overrides_should_be_called()
```
lbargaoanu: These tests seem to call the generic versions, so they are not relevant. Right? :)

robin-thoni: Note to myself: stop making PR at 1AM. I'll fix that tonight if I have time

robin-thoni: Fixed it

PR #2904, opened 2018-11-29
Remove CreateMissingTypeMaps from Profile

### LINKED ISSUE #2897: CreateMissingTypeMaps works only globally
I think we should remove it from Profile. We can make it throw initially.

lbargaoanu:
Fixes #2897.

### DIFF
--- a/src/AutoMapper/Mapper.cs
+++ b/src/AutoMapper/Mapper.cs
@@ -180,7 +180,7 @@ namespace AutoMapper
 
         #endregion
 
-        private readonly IConfigurationProvider _configurationProvider;
+        private readonly MapperConfiguration _configurationProvider;
         private readonly Func<Type, object> _serviceCtor;
 
         public Mapper(IConfigurationProvider configurationProvider)
@@ -190,7 +190,7 @@ namespace AutoMapper
 
         public Mapper(IConfigurationProvider configurationProvider, Func<Type, object> serviceCtor)
         {
-            _configurationProvider = configurationProvider ?? throw new ArgumentNullException(nameof(configurationProvider));
+            _configurationProvider = (MapperConfiguration) configurationProvider ?? throw new ArgumentNullException(nameof(configurationProvider));
             _serviceCtor = serviceCtor ?? throw new ArgumentNullException(nameof(serviceCtor));
             DefaultContext = new ResolutionContext(new ObjectMappingOperationOptions(serviceCtor), this);
         }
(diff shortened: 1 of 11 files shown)


src/AutoMapper/ProfileMap.cs
```diff
@@ -35,7 +35,7 @@ public ProfileMap(IProfileConfiguration profile, IConfiguration configuration)
             ShouldMapField = profile.ShouldMapField ?? configuration?.ShouldMapField ?? (p => p.IsPublic());
             ShouldMapProperty = profile.ShouldMapProperty ?? configuration?.ShouldMapProperty ?? (p => p.IsPublic());
             ShouldUseConstructor = profile.ShouldUseConstructor ?? configuration?.ShouldUseConstructor ?? (c => true);
-            CreateMissingTypeMaps = profile.CreateMissingTypeMaps ?? configuration?.CreateMissingTypeMaps ?? true;
+            CreateMissingTypeMaps = (profile as IConfiguration ?? configuration)?.CreateMissingTypeMaps ?? true;
```
lbargaoanu: No, I don't have a better idea :)

TylerCarlson1: Wouldn't `CreateMissingTypeMaps = configuration?.CreateMissingTypeMaps ?? true;` be better since profile is no longer being used for CreateMissingTypeMaps?

src/AutoMapper/Mapper.cs
```diff
@@ -25,7 +25,7 @@ public class Mapper : IRuntimeMapper
         public static IConfigurationProvider Configuration
```
lbargaoanu: Somewhat trivial, but I think here and for static mapper instance, we can use the concrete types, there is little value in using the interfaces.

lbargaoanu:
@jbogard If all is well, we should merge this so we get the version updated.

src/AutoMapper/ProfileMap.cs
jbogard: Does ProfileMap even need `CreateMissingTypeMaps`? If it's removed from `Profile`? This casting is bothering me - there's a deeper design issue here, I think.

lbargaoanu: Yes, those configuration interfaces are problematic. Part of the problem is that you can create a profile that doesn't have a reference to the global configuration. But without refactoring those again, I don't see a better way.

lbargaoanu: Anyway, this PR aside, the version neeeds to be updated for the MyGet build.

lbargaoanu: I see your point. But I don't think it's really propagating. It's more of a quick solution. And I think it's better then what we have now. A setting that's ignored. It's the minimal change required to remove CreateMissingTypeMaps from Profile.

PR #2915, opened 2018-12-10
Constructor validation fails for arrays

ushenkodmitry:
Hi,
after updating to version 8.0 all my tests started failing. The reason is in this PR.
All tests passed with the 7.1 version.

### DIFF
--- a/src/AutoMapper/TypeMap.cs
+++ b/src/AutoMapper/TypeMap.cs
@@ -110,6 +110,7 @@ namespace AutoMapper
             || CustomCtorFunction != null
             || ConstructDestinationUsingServiceLocator
             || ConstructorMap?.CanResolve == true
+            || DestinationTypeToUse.IsArray
             || DestinationTypeToUse.IsInterface()
             || DestinationTypeToUse.IsAbstract()
             || DestinationTypeToUse.IsGenericTypeDefinition()
--- a/src/UnitTests/Regression.cs
+++ b/src/UnitTests/Regression.cs
@@ -1,10 +1,8 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
-using System.Collections.ObjectModel;
 using Shouldly;
 using Xunit;
-using System.Linq;
 
 namespace AutoMapper.UnitTests
 {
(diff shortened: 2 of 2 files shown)


lbargaoanu:
Use ConvertUsing instead or set DisableCtorValidation for that map.

src/AutoMapper/TypeMap.cs
```diff
@@ -110,6 +110,7 @@ public PathMap FindOrCreatePathMapFor(LambdaExpression destinationExpression, Me
             || CustomCtorFunction != null
             || ConstructDestinationUsingServiceLocator
             || ConstructorMap?.CanResolve == true
+            || DestinationTypeToUse.IsArray
```
jbogard: This seems very specific. Why doesn't it fail for the no-arg ctor check further down?

lbargaoanu: > This used to work because ConstructorMap.CanResolve was broken. An array doesn't have a default constructor.

I don't have a better explanation and I'm not sure what you mean :)

ushenkodmitry:
Great, thank you.

пн, 10 дек. 2018 г., 20:03 Lucian Bargaoanu <notifications@github.com>:

src/AutoMapper/TypeMap.cs
jbogard: What is special about arrays?

lbargaoanu: They don't have a default constructor, but there is a special function to create them. And we do use that at runtime, just the validation fails.

lbargaoanu: Actually I'm not sure we would use that for a custom map. The test works because a destination is passed in. So just fail?

ushenkodmitry:
I've added DisableCtorValidation to the mapping configuration. And now i have InvalidoperationException:  Sequence contains no elements.

lbargaoanu:
Try the [MyGet](http://docs.automapper.org/en/stable/The-MyGet-build.html) build. Or ConvertUsing.

ushenkodmitry:
Exception raises in ConstructorParameterMap.cs file, line 24. While accessing SourceType property.

ushenkodmitry:
I can not use ConvertUsing, because my array already exists.

lbargaoanu:
So, why would that make any difference?

ushenkodmitry:
I've found the right overload! Thank you for help.

Issue #2930, opened 2019-01-02
Support creating maps via attributes


jbogard:
Folks in the community have created marker interfaces like "`IAutoMapFrom<TSource>`" to facilitate conventional creation of maps with scanning. We've got conventions but nothing that builds in via attributes.

For simple mappings, rather than relying on dynamic mapping, this would let you create a map instead of using a profile. Something like:

```c#
[AutoMap(SourceType = typeof(Foo)]
[AutoMap(typeof(Foo))]
[AutoMap(typeof(Foo), ReverseMap = true)]
public class FooDto {
}
```

jbogard:
Originally a feature idea from @JonPSmith using marker interfaces

TylerCarlson1:
This test would get the desired effect from current conventions.

The question is would this be settable convention that we can make, or is it an always on kind of feature?
Also same convention could be made with `IAutoMapFrom<TSource>` in this manner.

I would want to make it configurable, for any reason people want to disable it, and possibly on by default.

```
public class AttributeUsage : AutoMapperSpecBase
    {
        protected override MapperConfiguration Configuration { get; } = new MapperConfiguration(cfg =>
        {
            // Not sure why has to be false in order to pass but does
            cfg.CreateMissingTypeMaps = false;
            cfg.AddConditionalObjectMapper().Where((s, d) =>
            {
                var fromReverseMap = d.GetCustomAttribute<MapToAttribute>();
                return d.IsAssignableFrom(s.GetCustomAttribute<MapToAttribute>()?.ToType) ||
                       (fromReverseMap?.ReverseMap == true && s.IsAssignableFrom(fromReverseMap.ToType));
            });
        });

        public class To
        {
            public int I { get; set; }
        }

        [MapTo(typeof(To))]
        public class From
        {
            public int I { get; set; }
        }

        [MapTo(typeof(From), ReverseMap = true)]
        public class ToBack
        {
            public int I { get; set; }
        }

        [Fact]
        public void Main()
        {
            Mapper.Map<To>(new From{I = 5}).I.ShouldBe(5);
            Mapper.Map<ToBack>(new From { I = 5 }).I.ShouldBe(5);

            Should.Throw<AutoMapperMappingException>(() => Mapper.Map<From>(new To()));
        }

        [AttributeUsage(AttributeTargets.Class)]
        public class MapToAttribute : Attribute
        {
            public Type ToType { get; }
            public bool ReverseMap { get; set; }

[...]

PR #2933, opened 2019-01-03
Attributes to define a type map configuration

jbogard:


jbogard:
This also includes a refactoring of moving common functionality of typed/untyped mapper expressions into a common base class, separating the two interfaces into two base classes.

jbogard:
Fixes #2930

src/AutoMapper/Configuration/MapperConfigurationExpression.cs
```diff
@@ -44,41 +48,62 @@ public void AddProfile(Profile profile)
...
-
-            foreach (var profile in profiles)
+            foreach (var type in allTypes)
             {
-                AddProfile(profile);
+                if (typeof(Profile).IsAssignableFrom(type) && !type.IsAbstract)
+                {
+                    AddProfile(type.AsType());
+                }
+                foreach (var autoMapAttribute in type.GetCustomAttributes<AutoMapAttribute>())
```
lbargaoanu: There should be only one attribute I guess, because otherwise some kind of validation is needed.

lbargaoanu: OK, so why not assert that by using GetCustomAttribute instead?

lbargaoanu: OK, not that way :)  But check that the same pair is not duplicated.

jbogard: I think I'd rather leave that off for now - I'm not sure that this would really happen. What could happen is someone calling `CreateMap` and using this attribute. Today that'll be a configuration exception.

jbogard: Looks like we can't define multiple attributes anyway, unless I mark the attribute as `AllowMultiple`, which I have not. And I don't think I will - it's weird to map multiple sources to the same destination.

jbogard:
Everything else look good @lbargaoanu ?

src/AutoMapper/Configuration/MapperConfigurationExpression.cs
lbargaoanu: That was my initial thought too, AllowMultiple is false by default so why not GetAttribute.

jbogard:
Ah, gotcha. I can fix that.

PR #2944, opened 2019-01-10
WIP Attribute map configuration

### LINKED ISSUE #2931: Support a MapFrom attribute
There is a built-in "MapToAttribute" but that requires you decorate the source type. `MapFrom` is far more prevalent, so create a `MapFromAttribute` that does exactly what it says:

```c#
public class FooDto {
    [MapFrom("SomeOtherValue")]
}
```

An added bonus would be able to use this with `nameof`:

```c#
public class FooDto {
    [MapFrom(nameof(Foo.SomeOtherValue)]
}
```

jbogard:
Fixes #2931

### DIFF
--- a/src/AutoMapper/Configuration/Annotations/SourceMemberAttribute.cs
+++ b/src/AutoMapper/Configuration/Annotations/SourceMemberAttribute.cs
@@ -1,0 +1,16 @@
+﻿using System;
+
+namespace AutoMapper.Configuration.Annotations
+{
+    public sealed class SourceMemberAttribute : Attribute, IMemberConfigurationProvider
+    {
+        public string Name { get; }
+
+        public SourceMemberAttribute(string name) => Name = name;
+
+        public void ApplyConfiguration(IMemberConfigurationExpression memberConfigurationExpression)
+        {
+            memberConfigurationExpression.MapFrom(Name);
+        }
+    }
+}
--- a/src/AutoMapper/Configuration/MapperConfigurationExpression.cs
+++ b/src/AutoMapper/Configuration/MapperConfigurationExpression.cs
@@ -98,8 +98,16 @@ namespace AutoMapper.Configuration
                 var autoMapAttribute = type.GetCustomAttribute<AutoMapAttribute>();
                 if (autoMapAttribute != null)
                 {
-                    var mappingExpression = autoMapAttributeProfile.CreateMap(autoMapAttribute.SourceType, type);
+                    var mappingExpression = (MappingExpression) autoMapAttributeProfile.CreateMap(autoMapAttribute.SourceType, type);
                     autoMapAttribute.ApplyConfiguration(mappingExpression);
+
+                    foreach (var memberInfo in type.GetMembers(BindingFlags.Public | BindingFlags.Instance))
+                    {
+                        foreach (var memberConfigurationProvider in memberInfo.GetCustomAttributes().OfType<IMemberConfigurationProvider>())
+                        {
+                            mappingExpression.ForMember(memberInfo, cfg => memberConfigurationProvider.ApplyConfiguration(cfg));
+                        }
+                    }
                 }
             }
 
--- a/src/AutoMapper/Configuration/Annotations/NullSubstituteAttribute.cs
+++ b/src/AutoMapper/Configuration/Annotations/NullSubstituteAttribute.cs
@@ -1,0 +1,22 @@
+﻿using System;
+
+namespace AutoMapper.Configuration.Annotations
+{
+    public sealed class NullSubstituteAttribute : Attribute, IMemberConfigurationProvider
+    {
+        /// <summary>
+        /// Value to use if source value is null
+        /// </summary>
+        public object Value { get; }
+
+        public NullSubstituteAttribute(object value)
+        {
+            Value = value;
+        }
+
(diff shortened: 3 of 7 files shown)


jbogard:
Ready to review now. @JonPSmith this is the PR for attributes on source members. Unfortunately we don't allow string-based path configuration (MapFrom("Foo.Bar.Baz")) so if you need to redirect flattening, you can't use this.

JonPSmith:
Hi @jbogard,

I'll try and look at it this weekend or Monday. 

PS. I seems a bit light on unit tests - I normally look for those as they a great documentation on how this work.

src/AutoMapper/Configuration/Annotations/IgnoreAttribute.cs
```diff
@@ -0,0 +1,18 @@
...
+
+namespace AutoMapper.Configuration.Annotations
+{
+    /// <summary>
+    /// Ignore this member for configuration validation and skip during mapping.
+    /// </summary>
+    /// <remarks>
+    /// Must be used in combination with <see cref="AutoMapAttribute" />
+    /// </remarks>
+    public sealed class IgnoreAttribute : Attribute, IMemberConfigurationProvider
```
lbargaoanu: `[AttributeUsage(AttributeTargets.Field | AttributeTargets.Property)]` ? Wherever is applies.

src/AutoMapper/Configuration/Annotations/SourceMemberAttribute.cs
```diff
@@ -0,0 +1,22 @@
...
+
+namespace AutoMapper.Configuration.Annotations
+{
+    /// <summary>
+    /// Specify the source member to map from. Can only reference a member on the <see cref="AutoMapAttribute.SourceType" /> type
+    /// </summary>
+    /// <remarks>
+    /// Must be used in combination with <see cref="AutoMapAttribute" />
+    /// </remarks>
+    public sealed class SourceMemberAttribute : Attribute, IMemberConfigurationProvider
```
lbargaoanu: Maybe this could be applies to constructor parameters too.

lbargaoanu:
Maybe some validations? That existing attributes cannot be used the old ones (I know of two, IgnoreMap and SourceToDestinationMapperAttribute). That the class needs AutoMap for the other attributes to apply. These could be added later, it's true.

lbargaoanu:
AutoMapAttribute could be applied to structs too (it's missing in AttributeUsage).

JonPSmith:
I downloaded the AutoMapper and selected the attribute-map-configuration branch. The unit tests in namespace AttributeBasedMaps was very helpful indeed.

This all looks very good and my comments below are just tweaks to make it clearer.

I have two question on the implementation. 
1. How would some unit test their code that uses the `AutoMap` attribute? Your unit tests use `NonValidatingSpecBase`, is that what a developer would use? I supply code for [unit testing/NonDi usage](https://github.com/JonPSmith/EfCore.GenericServices/wiki/Unit-testing-your-code) of my libraries.
2. Does the `ApplyConfiguration` methods need to be public? If they could be internal then the developer won't see them, which is always a good thing.

Here are some thoughts to make the code clearer:  
It would be good to have `AttributeUsage` on all the attributes, as that is a great bit of documentation and save people doing the wrong thing. For instance its not clear whether the `NullSubstituteAttribute` can only be used on a property, or is it also applicable on the whole class. My reading of the code says property only, but the `AttributeUsage` would enforce that.

src/AutoMapper/Configuration/MapperConfigurationExpression.cs
```diff
@@ -98,8 +98,16 @@ private void AddMapsCore(IEnumerable<Assembly> assembliesToScan)
                 var autoMapAttribute = type.GetCustomAttribute<AutoMapAttribute>();
                 if (autoMapAttribute != null)
                 {
-                    var mappingExpression = autoMapAttributeProfile.CreateMap(autoMapAttribute.SourceType, type);
+                    var mappingExpression = (MappingExpression) autoMapAttributeProfile.CreateMap(autoMapAttribute.SourceType, type);
                     autoMapAttribute.ApplyConfiguration(mappingExpression);
+
+                    foreach (var memberInfo in type.GetMembers(BindingFlags.Public | BindingFlags.Instance))
```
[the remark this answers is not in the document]
jbogard: Don't have a `TypeDetails` yet. I also don't bother with the `ShouldMapXyz` methods.

jbogard:
@JonPSmith 

1. I had to unit test in the shown manner because I needed to isolate the scanned maps for validation. Scanning picks up other configuration in the test assembly and it was interfering with my tests. A user would test normally.
2. They don't need to be, but I don't see how a user would see them. You really only see the public constructors.

The missing `AttributeTargets` was an oversight, I've added them all back in.

Docs are still to be written, too.

jbogard:
@lbargaoanu I'm on the fence about what to do with the existing attributes. The `IgnoreMap` attribute has always been an outlier - I might just want to deprecate it in favor of this. I don't think you'd want to mix attributes and API configuration, but `IgnoreMap` lets you do that.

As for `SourceToDestinationMapperAttribute` and `MapToAttribute`, that one is a bit odd to me anyway, since you specify the attribute on the *source* type. It's confined to `Conventions`, so I don't think there will be much overlap.

PR #2952, opened 2019-01-17
Added AddProfiles method that takes an IEnumerable<Profile>

### LINKED ISSUE #2942: AddProfiles() overload that takes an IEnumerable<Profile> ?
Is there a reason why there is no overload of AddProfiles that just takes a collection of Profiles?

I recognize this is a minor quibble, but it seems like a silly omission. I can either add a single Profile at a time, or I can scan any number of assemblies to auto-add multiple profiles. I was surprised to find there's no built-in way to add multiples without going all the way into assembly scanning.

I feel like this should have come up years ago, so I'm surprised that I'm having to make this request!

To put it specifically, I'd simply like to be able to do this:
```
IEnumerable<Profile> profiles = GetProfilesFromSomewhere();
foreach (Profile profile in profiles)
{
  cfg.AddProfile(profile);
}
```

Like this:
```
IEnumerable<Profile> profiles = GetProfilesFromSomewhere();
cfg.AddProfiles(profiles);
```

jpridx:
- Simple foreach that calls AddProfile for each element of the sequence
 - Added unit test AddProfiles.cs

Fixes #2942.

### DIFF
--- a/src/UnitTests/AddProfiles.cs
+++ b/src/UnitTests/AddProfiles.cs
@@ -1,0 +1,40 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using Shouldly;
+using Xunit;
+
+namespace AutoMapper.UnitTests
+{
+    public class AddProfiles : SpecBase
+    {
+        public class Source { }
+        public class Dest { }
+
+        public class ForwardProfile : Profile
+        {
+            public ForwardProfile()
+            {
+                CreateMap<Source, Dest>();
+            }
+        }
+
+        public class ReverseProfile : Profile
+        {
+            public ReverseProfile()
+            {
+                CreateMap<Dest, Source>();
+            }
+        }
+
+        [Fact]
+        public void Should_not_throw_when_loading_multiple_profiles()
+        {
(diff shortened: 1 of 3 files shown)


src/UnitTests/AddProfiles.cs
```diff
@@ -0,0 +1,40 @@
...
+            }
+        }
+
+        [Fact]
+        public void Should_not_throw_when_loading_multiple_profiles()
+        {
+            IEnumerable<Profile> profiles = new Profile[] { new ForwardProfile(), new ReverseProfile() };
+            var config = new MapperConfiguration(cfg => cfg.AddProfiles(profiles));
+
+            config.AssertConfigurationIsValid();
```
lbargaoanu: This tests only asserts the method exists, it doesn't really check it adds both profiles.

jbogard: @jpridx the `MapperConfiguration` object has a `Profile` property that you can assert against. Configuration validation can succeed if there are zero maps.

jpridx: I added a test that verifies that the profiles were added. @jbogard thank you for pointing me to the Profiles property.

PR #2960, opened 2019-01-22
Add support to specify ShouldMapMethod 

### LINKED ISSUE #2959: Add support to skip scanning some methods when building TypeDetails
From a .net core lib I am consuming a legacy lib written in .net 3.5. I am configuring some mappings with classes that live in that lib. Some of these classes are exposing some non-static public methods that returns or receives as arguments non-compatible types with .net core (types residing in System.Web or System.Web.Services). Even though, I am sure that these methods won't be used from the code I am building, I am getting an error in startup because of the following method that runs when creating internally the type details

```csharp
        private MethodInfo[] BuildPublicNoArgMethods()
        {
            return Type.GetAllMethods()
                .Where(mi => mi.IsPublic && !mi.IsStatic && mi.DeclaringType != typeof(object))
                .Where(m => (m.ReturnType != typeof(void)) && (m.GetParameters().Length == 0))
                .ToArray();
        }
```

Since the return or parameter type cannot be loaded, the initialization process stops with a TypeLoadException.

Changing the legacy assembly is not an option because is out of my control. And I am trying to avoid to add a reference to System.Web or System.Web.Services just for the purpose to fool the initialization mapping.

To prevent this issue, it would be great if we could limit what is being scanned during type details creation. I can think of several solutions
1. Add new config property `SkipScanPublicNoArgMethods` - In this case non-static public methods won't be scanned at all
2. Add new config property `IgnoreTypeLoadErrorsInPublicMethods`
3. Add a new config property `ShouldScanPublicMethod` (`Func<MethodInfo, bool>`)

I think 1) is very simple. Do you think any of these solutions is reasonable?

Please let me know yout thoughts.

bfcamara:
Add support to specify ShouldMapMethod to determine which methods, of those that are eligible to be mapped (parameterless, public, and non-static or extension methods) should be mapped.

Fixed #2959

### DIFF
--- a/src/AutoMapper/TypeDetails.cs
+++ b/src/AutoMapper/TypeDetails.cs
@@ -24,7 +24,7 @@ namespace AutoMapper
             PublicWriteAccessors = BuildPublicAccessors(publicWritableMembers);
             PublicNoArgMethods = BuildPublicNoArgMethods(config.ShouldMapMethod);
             Constructors = GetAllConstructors(config.ShouldUseConstructor);
-            PublicNoArgExtensionMethods = BuildPublicNoArgExtensionMethods(config.SourceExtensionMethods, config.ShouldMapMethod);
+            PublicNoArgExtensionMethods = BuildPublicNoArgExtensionMethods(config.SourceExtensionMethods.Where(config.ShouldMapMethod));
             AllMembers = PublicReadAccessors.Concat(PublicNoArgMethods).Concat(PublicNoArgExtensionMethods).ToList();
             DestinationMemberNames = AllMembers.Select(mi => new DestinationMemberName { Member = mi, Possibles = PossibleNames(mi.Name, config.Prefixes, config.Postfixes).ToArray() });
         }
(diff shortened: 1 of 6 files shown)


src/AutoMapper/TypeDetails.cs
```diff
@@ -99,9 +99,9 @@ public struct DestinationMemberName
 
         public IEnumerable<DestinationMemberName> DestinationMemberNames { get; set; }
 
-        private IEnumerable<MethodInfo> BuildPublicNoArgExtensionMethods(IEnumerable<MethodInfo> sourceExtensionMethodSearch)
+        private IEnumerable<MethodInfo> BuildPublicNoArgExtensionMethods(IEnumerable<MethodInfo> sourceExtensionMethodSearch, Func<MethodInfo, bool> shouldMapMethod)
         {
-            var explicitExtensionMethods = sourceExtensionMethodSearch.Where(method => method.GetParameters()[0].ParameterType == Type);
+            var explicitExtensionMethods = sourceExtensionMethodSearch.Where(shouldMapMethod).Where(method => method.GetParameters()[0].ParameterType == Type);
```
lbargaoanu: I think we can remove all the filters inside the method and simply filter the input parameter when calling the method.

bfcamara: Make sense!

src/UnitTests/ShouldMapMethod.cs
```diff
@@ -0,0 +1,73 @@
+﻿using Xunit;
+using Shouldly;
+using System;
+
+namespace AutoMapper.UnitTests
+{
+    public class ShouldMapMethod : NonValidatingSpecBase
+    {
```
lbargaoanu: This looks good, but the extension method case needs a test too.

bfcamara: Ok. I'm gonna add it.

bfcamara:
Hi @jbogard, did you have a chance to look to this PR? Please let me know if anything is missing.

Thanks in advance.

bfcamara:
Thank you very much @jbogard. Any chance to launch a pre-release version on nuget that include this change.

lbargaoanu:
@bfcamara Try the [MyGet](https://docs.automapper.org/en/stable/The-MyGet-build.html) build.

bfcamara:
@lbargaoanu Thank you. MyGet build worked.

PR #2966, opened 2019-01-29
Fix for #2961

### LINKED ISSUE #2961: Mapping parameters in VB not working
Related to #1748

I've just updated Automapper to 8.0 (and .Net to 4.6.1, which i think might be the cause of the issue) and paramterised mappings now fail with the error:

```
Field 'Taploma.EventTemplateProfile+_Closure$__1-0.$VB$Local_userID' is not defined for type 'Taploma.EventInstance'
```

Looking at the previous issue i raised it seems like the newer version of .NET may have changed the name of closure variables.

### Mapping configuration

```VB.NET

  Dim userID = 0

        CreateMap(Of EventTemplate, MyEventsDto).
            ForMember(Function(dest As MyEventsDto) dest.Template,
                      Sub(opts) opts.MapFrom(Function(src As EventTemplate) src)).
            ForMember(Function(dest As MyEventsDto) dest.Instances,
                      Sub(opts) opts.MapFrom(Function(src As EventTemplate) src.Instances.
                                                 Where(Function(x) x.Invites.Any(Function(i) i.SentToID = userID))))

```

### Version: x.y.z
Automapper: 8.0.0
EF: 6.2.0

bodrick:
Seems that for VB.Net dictionary mapping of parameters works but wasn't working for parameterized queries, this fixes the problem.

Fixes #2961

### DIFF
--- a/src/AutoMapper/QueryableExtensions/ExpressionBuilder.cs
+++ b/src/AutoMapper/QueryableExtensions/ExpressionBuilder.cs
@@ -361,6 +361,12 @@ namespace AutoMapper.QueryableExtensions
                 }
                 var parameterName = node.Member.Name;
 
+                const string vbPrefix = "$VB$Local_";
+                if (parameterName.StartsWith(vbPrefix, StringComparison.Ordinal))
+                {
+                    parameterName = parameterName.Substring(vbPrefix.Length);
+                }
+
                 var matchingMember = _parameters.GetType().GetDeclaredProperty(parameterName);
 
                 return matchingMember != null 


src/AutoMapper/QueryableExtensions/ExpressionBuilder.cs
```diff
@@ -361,6 +361,12 @@ protected override Expression VisitMember(MemberExpression node)
                 }
                 var parameterName = node.Member.Name;
 
+                const string vbPrefix = "$VB$Local_";
```
lbargaoanu: This code needs to be reused somehow. Maybe a base class for the two cases.

PR #2972, opened 2019-02-04
Include members 

### LINKED ISSUE #2949: Include member mappings
Like in #2948. We can do flattening, but this one would allow you to reuse the configuration.
```csharp
cfg.CreateMap<Inner, Dto>();
cfg.CreateMap<Outer, Dto>().Include(src => src.Inner);
``` 
We could have things map to the same destination members, maybe just throw then.
We could start by not supporting cyclical dependencies.
It doesn't seem simple.

lbargaoanu:
Closes #2949.

### DIFF
--- a/src/AutoMapper/TypeMap.cs
+++ b/src/AutoMapper/TypeMap.cs
@@ -1,4 +1,3 @@
-
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
--- a/src/AutoMapper/TypeMap.cs
+++ b/src/AutoMapper/TypeMap.cs
@@ -271,6 +271,7 @@ namespace AutoMapper
             }
 
             _orderedPropertyMaps = PropertyMaps.OrderBy(map => map.MappingOrder).ToArray();
+            _propertyMaps.Clear();
 
             MapExpression = CreateMapperLambda(configurationProvider, null);
         }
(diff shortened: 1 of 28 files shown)


src/AutoMapper/TypeMap.cs
```diff
@@ -288,39 +286,11 @@ public void Seal(IConfigurationProvider configurationProvider)
                 null :
                 new TypeMapPlanBuilder(configurationProvider, this).CreateMapperLambda(typeMapsPath);
 
-        public PropertyMap GetExistingPropertyMapFor(MemberInfo destinationProperty)
```
lbargaoanu: I'm not sure what this does, but the tests pass :)

src/AutoMapper/TypeMap.cs
```diff
@@ -18,17 +17,17 @@ namespace AutoMapper
...
+        private readonly HashSet<LambdaExpression> _beforeMapActions = new HashSet<LambdaExpression>();
         private readonly HashSet<TypePair> _includedDerivedTypes = new HashSet<TypePair>();
         private readonly HashSet<TypePair> _includedBaseTypes = new HashSet<TypePair>();
-        private readonly List<PropertyMap> _propertyMaps = new List<PropertyMap>();
-        private readonly List<PathMap> _pathMaps = new List<PathMap>();
-        private readonly List<SourceMemberConfig> _sourceMemberConfigs = new List<SourceMemberConfig>();
-        private readonly IList<PropertyMap> _inheritedMaps = new List<PropertyMap>();
+        private readonly Dictionary<string, PropertyMap> _propertyMaps = new Dictionary<string, PropertyMap>();
+        private readonly Dictionary<MemberPath, PathMap> _pathMaps = new Dictionary<MemberPath, PathMap>();
+        private readonly Dictionary<MemberInfo, SourceMemberConfig> _sourceMemberConfigs = new Dictionary<MemberInfo, SourceMemberConfig>();
```
lbargaoanu: Probably this should be keyed by member name instead.

Issue #2936, opened 2019-01-07
Support configuring attribute maps with additional properties


jbogard:
Follow up to #2930, the initial implementation only does reverse map configuration.

There are a number of configuration options that we could include as part of the `AutoMapAttribute` configuration.

Easy ones:

 - ConstructUsingServiceLocator (bool)
 - MaxDepth (int)
 - PreserveReferences (bool)
 - DisableCtorValidation (bool)
 - MemberList (enum)
 - IncludeAllDerived (bool)
 - TypeConverter (Type)

Not-so-easy-ones:

 - Value transformers
 - Expression/func-based type converters
 - Before/after map
 - Include/IncludeBase

I'd lean towards if you're trying to do the "hard ones", then you should use the API because any attribute-based configuration would be ugly.

gmilani:
Hello @jbogard @lbargaoanu 
I want to start contributing to AutoMapper project. Can I start with the easy configuration options?

lbargaoanu:
Sure, but do look at https://github.com/AutoMapper/AutoMapper/pull/2944 first.

gmilani:
We are configuring the mapping directly in the class, the properties are configured from annotations. So I can not imagine a situation where it would be necessary to use `IncludeAllDerived`, since the configuration is already explicitly in the property of the base class. Can you show me an example if I'm mistaken?

lbargaoanu:
A MapFrom will be inherited by default, which is nice, but [runtime polymorphism](http://docs.automapper.org/en/latest/Mapping-inheritance.html#runtime-polymorphism) does't work by default.

PR #2981, opened 2019-02-10
AutoMap attribute configuration options

gmilani:
Adding the following configuration options by the AutoMap attribute

- ConstructUsingServiceLocator
- MaxDepth
- PreserveReferences
- DisableCtorValidation
- IncludeAllDerived
- TypeConverter

#2936

### DIFF
--- a/src/AutoMapper/AutoMapAttribute.cs
+++ b/src/AutoMapper/AutoMapAttribute.cs
@@ -37,6 +37,11 @@ namespace AutoMapper
         public bool DisableCtorValidation { get; set; }
 
         /// <summary>
+        /// If set to true, include this configuration in all derived types' maps.
+        /// </summary>
+        public bool IncludeAllDerived { get; set; }
+
+        /// <summary>
         /// Skip normal member mapping and convert using a <see cref="ITypeConverter{TSource,TDestination}"/> instantiated during mapping.
         /// </summary>
         public Type TypeConverter { get; set; }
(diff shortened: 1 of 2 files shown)


src/AutoMapper/AutoMapAttribute.cs
```diff
@@ -15,12 +15,73 @@ public AutoMapAttribute(Type sourceType)
...
         public bool ReverseMap { get; set; }
 
+        /// <summary>
+        /// If set to true, construct the destination object using the service locator.
+        /// </summary>
+        public bool ConstructUsingServiceLocator { get; set; }
+
+        /// <summary>
+        /// For self-referential types, limit recurse depth.
+        /// Enables PreserveReferences.
```
lbargaoanu: That's not the case. Maybe in previous versions.

src/UnitTests/AttributeBasedMaps.cs
```diff
@@ -606,5 +607,221 @@ public void Should_validate_successfully()
...
+                var dest = Mapper.Map<Dest>(source);
+                dest.Level.ShouldBe(1);
+                dest.Children[0].Level.ShouldBe(2);
+                dest.Children[0].Children.ShouldAllBe(d => d == null);
+                dest.Children[1].Level.ShouldBe(3);
+                dest.Children[1].Children.ShouldAllBe(d => d == null);
+            }
+        }
+
+        public class When_specifying_to_preserve_references_via_attribute : NonValidatingSpecBase
```
lbargaoanu: This test passes regardless of the attribute setting. Because of the built in logic to prevent SO. Try a non recursive mapping.

gmilani:
Hi, @lbargaoanu 
I ended up getting my schedule tight. As soon as I free myself, I will pay attention to this case again.

src/UnitTests/AttributeBasedMaps.cs
gmilani: I changed the test!

Issue #2991, opened 2019-02-18
Extensions API


Tasteful:
# Problem description

When building extensions based on AutoMapper (AM) it does not exists any place where the configuration data can be stored together with the AM configuration. In AutoMapper.Collection (AMC) we are using static parameters to store the configured comparer that are configured together with the AM mapping. When running AM with AMC configuration in parallel; example for testing, the configuration will often be corrupt and mapping will not work correctly that will result in errors in the application. 

# Suggested solution

To be able to build stable extension on AM without all the time extend AM with new features I suggest to implement `FeatureCollection` that can handle all different possibilities extensions. The concept of `FeatureCollection` will be the same in AM as ASP.NET Core Http Abstraction is using to extend the Http Request with new features without need to change the core implementation.

A `FeatureCollection` is a typed dictionary where each item can be sealed when the configuration executes the `Seal`. The item in the `FeatureCollection` can be accessed during runtime and will act as the configuration store for the extensions.

In AM two places is identified that need the `FeatureCollection` that contains `IFeature`.

1. `IConfigurationProvider` (breaking change).
2. `TypeMap`.

To be able to populate the above `FeatureCollection`s the following members have been modified. The below `I*Feature` is used to configure the `IFeature` and place them in either the `IConfigurationProvider` or in the `TypeMap` before they are sealed.

1. `AdvancedConfiguration` to hold a collection of `IMapperConfigurationExpressionFeature` (this may be moved into `IMapperConfigurationExpression` for consistency).
2. `IMappingExpression` to hold a collection of `IMappingExpressionFeature` (breaking change).

See the linked PRs for proof of concept.

lbargaoanu:
Sounds good. If @jbogard agrees, we can discuss details about the implementation.

jbogard:
It's an interesting concept! I had assumed just extension methods to the configuration expression would suffice - but it doesn't appear to be the case.

Looks like it cleans up the Collections setup quite a bit.

Tasteful:
I was thinking that it should be something simple like that from the beginning so stared with only use reflection to add items into `MappingExpressionBase.TypeMapActions`, but when I look little deeper the objects `IMapperConfigurationExpression`/`IMappingExpression` does not have any relationship to the created object that exists in the `IConfigurationProvider`/`TypeMap`/`TypePair` so it is hard to connect them to each other.

lnaie:
[...]

PR #2992, opened 2019-02-18
POC: FeatureCollection

Tasteful:
See https://github.com/AutoMapper/AutoMapper/issues/2991


src/AutoMapper/Configuration/MappingExpression.cs
```diff
@@ -18,8 +18,14 @@ public void ConvertUsing<TTypeConverter>()
         public IMappingExpression ReverseMap()
         {
             var reverseMap = new MappingExpression(new TypePair(Types.DestinationType, Types.SourceType), MemberList.Source);
-            reverseMap.MemberConfigurations.AddRange(MemberConfigurations.Select(m => m.Reverse()).Where(m => m != null));
+            reverseMap.MemberConfigurations.AddRange(MemberConfigurations.Select(m => m.Reverse()).Where(m => m != null));
             ReverseMapExpression = reverseMap;
+
+            foreach (var feature in Features)
```
lbargaoanu: I think the collection should have a method ReverseConfiguration or smth.

src/AutoMapper/Configuration/MappingExpressionBase.cs
```diff
@@ -74,6 +75,10 @@ public void Configure(TypeMap typeMap)
             {
                 typeMap.AddValueTransformation(valueTransformer);
             }
+            foreach(var feature in Features)
```
lbargaoanu: The same, a Configure method on the collection.

src/AutoMapper/IFeatureCollectionBase.cs
```diff
@@ -0,0 +1,29 @@
...
+        /// <typeparam name="TFeature">The type of the feature.</typeparam>
+        /// <returns>The feature or null if feature not exists.</returns>
+        TFeature Get<TFeature>() where TFeature : TValue;
+
+        /// <summary>
+        /// Sets the feature for type <typeparamref name="TFeature"/>.
+        /// </summary>
+        /// <typeparam name="TFeature">The type of the t feature.</typeparam>
+        /// <param name="feature">The feature. To remove a feature, pass null value.</param>
+        void Set<TFeature>(TFeature feature) where TFeature : TValue;
```
lbargaoanu: A collection with Set seems weird. Maybe an Add that disallows duplicates, like Dictionary?

src/AutoMapper/MapperConfiguration.cs
```diff
@@ -46,6 +46,11 @@ public MapperConfiguration(MapperConfigurationExpression configurationExpression
 
             Configuration = new ProfileMap(configurationExpression);
             Profiles = new[] { Configuration }.Concat(configurationExpression.Profiles.Select(p => new ProfileMap(p, configurationExpression))).ToArray();
+
+            foreach(var feature in configurationExpression.Advanced.Features)
```
lbargaoanu: Configure method on Features.

src/AutoMapper/MapperConfiguration.cs
```diff
@@ -371,6 +378,11 @@ private void Seal()
             foreach (var typeMap in _typeMapRegistry.Values)
             {
                 typeMap.Seal(this);
+            }
+
+            foreach (var feature in Features)
```
lbargaoanu: Seal method on Features.

src/AutoMapper/MappingExpressionFeatureCollectionExtensions.cs
```diff
@@ -0,0 +1,29 @@
+﻿namespace AutoMapper
+{
+    public static class MappingExpressionFeatureCollectionExtensions
```
lbargaoanu: I don't this this is needed.

src/AutoMapper/TypeMap.cs
```diff
@@ -281,6 +283,8 @@ public void Seal(IConfigurationProvider configurationProvider)
                     .OrderBy(map => map.MappingOrder).ToArray();
 
             MapExpression = CreateMapperLambda(configurationProvider, null);
+
+            ((FeatureCollection)Features).Seal(configurationProvider);
```
lbargaoanu: Maybe just drop all the interfaces and use the classes instead?

src/AutoMapper/Internal/FeatureCollectionBase.cs
```diff
@@ -0,0 +1,84 @@
+﻿using System;
+using System.Collections;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+
+namespace AutoMapper.Internal
+{
+    public class FeatureCollectionBase<TValue> : IEnumerable<KeyValuePair<Type, TValue>>
+    {
+        private readonly Lazy<ConcurrentDictionary<Type, TValue>> _features = new Lazy<ConcurrentDictionary<Type, TValue>>(() => new ConcurrentDictionary<Type, TValue>());
```
lbargaoanu: Configuration doesn't need to be thread safe. Also Lazy just makes things harder to read. Maybe drop the readonly flag and replace the dictionary with a readonly one after Seal. If it isn't too complicated.

Tasteful: The `Lazy` is added to not allocate extra memory for the `Features`-collection that will be created for each `IMappingExpression`/`TypeMap` if that not is used. Should I still remove the lazy?

the `_features` field is protected by the set on `this[]` to not allow changes if the collection is sealed so no reason to replace the inner dictionary.

src/AutoMapper/IFeatureCollectionBase.cs
Tasteful: If we use `Add` and throw we also need to have an `Remove`-method to be able to replace an already added item. Naming is from the asp.net core FeatureCollection.

Otherwise the `AddOrUpdate` is a common pattern for collections.

src/AutoMapper/MappingExpressionFeatureCollectionExtensions.cs
Tasteful: No, correct. It was some code that I need before I added the `Get`/`Set` on the collection itself. I will remove.

src/AutoMapper/Internal/FeatureCollectionBase.cs
lbargaoanu: I understand the intent. It just doesn't seem useful. I think that replacing the dictionary with a readonly one and dropping the flag is a more elegant way to achieve the same purpose. But maybe it's just me :)

src/AutoMapper/TypeMap.cs
Tasteful: I don't see the need for the interfaces for a lot of places for this feature, but AM is using interfaces for most of the classes I added functionality into so tried to only follow the pattern. Should I clean the interfaces that not is needed away?

lbargaoanu: That's to better separate public API from implementation. But given the purpose of this feature, I think classes are just fine. cc @jbogard

src/AutoMapper/Configuration/MappingExpression.cs
Tasteful: The AM itself use the `Reverse()` on ex `IPropertyMapConfiguration` (line 21 above), but can add a method on the collection to return an reversed collection that will be used. I think we should keep the name `Reverse()` (to follow the above) or add it as `CreateReverseCollection()`.

lbargaoanu: There is already a Reverse on IEnumerable, so I thought we should avoid the confusion.

src/AutoMapper/IFeatureCollectionBase.cs
jbogard: So this is emulating https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.http.features.featurecollection?view=aspnetcore-2.2 ?

lbargaoanu: I've never used that, but why would you need to call Set twice for the same extension? It seems simpler to just throw.

Tasteful: Yes, it is the item I checked during build, I ignored the `Revision` property due to the usage in .net core where the collection have a "base collection" and then you can replace/add items on that one that not will reflect in the "base collection".

src/AutoMapper/Internal/FeatureCollectionBase.cs
Tasteful: I will change it but will not initiate the collection before first set/add-usage. then we don't allocate not needed memory.

src/AutoMapper/IFeatureCollectionBase.cs
Tasteful: In [xunit](https://github.com/xunit/abstractions.xunit/blob/master/src/xunit.abstractions/Frameworks/ITestFrameworkOptions.cs) we have similar collection but it using a string-key instead, liking this model better.

lbargaoanu: Why?

Tasteful: I don't need to remember the magic strings for the key. I also have a feeling thst it use less allocations but nothing that I have run profiler on.

lbargaoanu: No, that one makes sense :) The Set thing I don't understand. It seems like an extension shouldn't need to call Set twice, so Add seems better to me.

src/AutoMapper/Internal/FeatureCollectionBase.cs
lbargaoanu: Allocating an empty collection doesn't really matter. Having extra code to avoid that is just premature optimization.

src/AutoMapper/IFeatureCollectionBase.cs
Tasteful: I think they prefere to be able to set an item multiple time (updating values) without first checking if the key already exists and in that case remove before the add can occur. If using Add and we should simulate the dictionary we should also throw exception on Get if key not exists.

The xunit also use GetValue/SetValue.

@jbogard any specific taste for Get/Add/Remove vs. Get/Set?

jbogard: Is there a data structure that already matches the behavior we're looking for?

lbargaoanu: It's a dictionary as implementation, but I think the idea is to expose a restricted interface. I, for one, don't see a need for anything besides GetOrAdd. This would ensure a single feature per extension and it would be easy to use.

Tasteful: I should say that sometimes it need to replace the data for a specific extension, example
```
cfg.CreateMap<OrderDomain, OrderEf>()
    .EqualityComparison((oo, dto) => oo.Id == dto.Id)
    .ReverseMap()
    .EqualityComparison((dto, oo) => dto.AnotherId == oo.AnotherId);
```
where the `ReverseMap()` will add an reverse version of the `.EqualityComparison((oo, dto) => oo.Id == dto.Id)` and then it will be overwritten by `.EqualityComparison((dto, oo) => dto.AnotherId == oo.AnotherId)`. It also depend of if the item that is stored in the feature collection is an immutable object or not. If the object is immutable we need to replace the object, with a new object, otherwise it can probably be updated but we in that case always need to use logic to check if the object exists as a feature. If we have a `Set` method only then we always overwrite the item with the new item and don't care if that was existing before adding the new object. That means that only Get/Add not is enough and we need to have Get/Add/Remove in that case.

lbargaoanu: You could replace the object as you say, or you can only have GetOrAdd and update an existing object.

src/AutoMapper/Internal/FeatureCollectionBase.cs
```diff
@@ -0,0 +1,73 @@
...
+{
+    public class FeatureCollectionBase<TValue> : IEnumerable<KeyValuePair<Type, TValue>>
+    {
+        private IDictionary<Type, TValue> _features = new Dictionary<Type, TValue>();
+
+        public TValue this[Type key]
+        {
+            get
+            {
+                return _features.TryGetValue(key, out var feature)
```
lbargaoanu: https://github.com/AutoMapper/AutoMapper/blob/8f8794e40d3fff8255542edf4e63af35f3410857/src/AutoMapper/Configuration/PrimitiveExtensions.cs#L13

src/AutoMapper/FeatureCollection.cs
```diff
@@ -0,0 +1,16 @@
+﻿using AutoMapper.Internal;
+
+namespace AutoMapper
+{
+    public class FeatureCollection : FeatureCollectionBase<IFeature>, IFeature
+    {
+        void IFeature.Seal(IConfigurationProvider configurationProvider)
```
lbargaoanu: This can be public and then the casts below are not needed anymore.

src/AutoMapper/Internal/FeatureCollectionBase.cs
```diff
@@ -0,0 +1,73 @@
...
+        {
+            return (TFeature)this[typeof(TFeature)];
+        }
+
+        /// <summary>
+        /// Sets the feature for type <typeparamref name="TFeature"/>.
+        /// </summary>
+        /// <typeparam name="TFeature">The type of the t feature.</typeparam>
+        /// <param name="feature">The feature. To remove a feature, pass null value.</param>
+        public void Set<TFeature>(TFeature feature)
```
lbargaoanu: This doesn't need to be generic. We can rely on GetType and this would simplify things, as the key/value thing is encapsulated and all calling code can just use the values.

src/AutoMapper/Internal/FeatureCollectionBase.cs
```diff
@@ -0,0 +1,73 @@
...
+            where TFeature : TValue
+        {
+            this[typeof(TFeature)] = feature;
+        }
+
+        public IEnumerator<KeyValuePair<Type, TValue>> GetEnumerator()
```
lbargaoanu: => _features.GetEnumerator();

src/AutoMapper/Internal/FeatureCollectionBase.cs
jbogard: Is there any case we'd prefer `TFeature` over `feature.GetType()`? I can't think of any.

Tasteful: I think `typeof(TFeature)` will be treated more as compile time instead of running `feature.GetType()` but I can be wrong on that. The usage will be the same with `Fetures.Set(new MyFeatureClass())`.

src/AutoMapper/FeatureCollection.cs
Tasteful: I was making it explicit implementation to not show the method to the outside, it is only the internal api that should call this method. 

Making the method `internal` will work for the use-case also, but then it will be harder to write unit-tests for it (if we not allowing internal members for the unit tests).

Should it be hidden for the normal user with either the explicit implementation or internal?

jbogard: Our convention right now is to internalize it.

src/AutoMapper/IFeatureCollectionBase.cs
Tasteful: I changed the name to `Add` instead and it will have the same logic as the indexed property `this[type] = value` so it replaces the value if it is already existing.

src/AutoMapper/Internal/FeatureCollectionBase.cs
Tasteful: The `typeof(TFeature)` is using the declared type and the `feature.GetType()` is using the runtime type (https://stackoverflow.com/a/17495443). So if we using `TFeature` we can allow sub-classing of the feature-classes and they still will be treated as the same. Using `.GetType()` they will be different individual features.

From that I prefer the typed version.

lbargaoanu: Yes, but we don't have to support that here. Extensions developers inherit away internally.

src/AutoMapper/IFeatureCollectionBase.cs
lbargaoanu: That seems misleading.

src/AutoMapper/Internal/FeatureCollectionBase.cs
Tasteful: If we see the typed version with `TFeature` we can restrict the feature to be of a specific type. If we use `feature.GetType()` we also need an `if (!(feature is TValue)) throw new ArgumentException("Feature is of wrong type.");` to ensure the type of the feature. I prefer to have the typed version.

src/AutoMapper/IFeatureCollectionBase.cs
Tasteful: Depend on how the feature class is build up I in that case prefer `AddOrUpdate` instead of `GetOrAdd` to not always need to put in an delegate for add even if I only want to check if the feature exists.

When you write code that is aware or the feature and will check if that exists before adding need to add that if-logic in the calling code.

PR #2997, opened 2019-02-20
Dropping value tuples.

### LINKED ISSUE #2750: System.ValueTuple package should not be referenced for .NET 4.7+ projects
### Expected behavior

System.ValueTuple dependency should not be included in projects that target .NET 4.7 and up.

### Actual behavior

System.ValueTuple dependency is added and causes our web projects, which target .net 4.7.2, to require the following web.config binding redirect:
```
<dependentAssembly>
	<assemblyIdentity name="System.ValueTuple" publicKeyToken="CC7B13FFCD2DDD51" culture="neutral"/>
	<bindingRedirect oldVersion="0.0.0.0-4.0.3.0" newVersion="4.0.3.0"/>
</dependentAssembly>
```
Which is in turn leading to a `System.ExecutionEngine` exception for us in w3wp.exe that I've been unable to track down.  Our projects are a mix of .NET 4.7.2 and .NET Standard 2.0 projects and the dependency is likely causing a binary compatibility issue between the different implementations.  We've experienced similar issues with System.Net.Http in the past.

With AutoMapper 7.0.0, the dependency is not there, and all works well.

jbogard:
Fixes #2750

### DIFF
--- a/src/AutoMapper/AutoMapper.csproj
+++ b/src/AutoMapper/AutoMapper.csproj
@@ -26,7 +26,6 @@
 
   <ItemGroup Condition=" '$(TargetFramework)' == 'net461' ">
     <Reference Include="Microsoft.CSharp" />
-    <PackageReference Include="System.ValueTuple" Version="4.5.0" />
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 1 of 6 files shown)


screamingworld:
Thank you guys! This was realy fast and saw that the tuple was not extensive used. So I think that is a good and cleaner solution than using the value tuples...
I am waiting for the release...

PR #3001, opened 2019-02-25
Respect PathMappings in unmapped property names list

### LINKED ISSUE #3000: MemberList.Source and ForPath
### Source/destination types

```csharp
class TargetOuter
{
    public TargetInner Inner { get; set; }

    // The properties below should be ignored, they are not relevant
    public int Unrelated { get; set; }
    public string AlsoUnrelated { get; set; }
}

class TargetInner
{
    public string MyProp { get; set; }
}

class Input
{
    public string Source { get; set; }
}
```

### Mapping configuration

```csharp
Mapper.Initialize(opt => opt.CreateMap<Input, TargetOuter>(MemberList.Source)
                .ForPath(x => x.Inner.MyProp, y => y.MapFrom(x => x.Source)));
```

### Expected behavior
The call of `Mapper.AssertConfigurationIsValid()` should not throw an exception, because all properties from the source (note the `MemberList.Source` configuration value) have been mapped.

### Actual behavior
The call throws the following exception:

```
AutoMapper.AutoMapperConfigurationException: 
Unmapped members were found. Review the types and members below.
Add a custom mapping expression, ignore, add a custom resolver, or modify the source/destination type
For no matching constructor, add a no-arg ctor, add optional arguments, or map all of the constructor parameters
===============================================================
Input -> TargetOuter (Source member list)
AutoMapper_Assertion.Input -> AutoMapper_Assertion.TargetOuter (Source member list)

Unmapped properties:
Source

   at AutoMapper.ConfigurationValidator.AssertConfigurationIsValid(IEnumerable`1 typeMaps) in C:\projects\automapper\src\AutoMapper\ConfigurationValidator.cs:line 28
   at AutoMapper.MapperConfiguration.AssertConfigurationIsValid() in C:\projects\automapper\src\AutoMapper\MapperConfiguration.cs:line 305
   at AutoMapper.Mapper.AssertConfigurationIsValid() in C:\projects\automapper\src\AutoMapper\Mapper.cs:line 179
```

The mapping actually works, which can be seen on the `equal` variable in the reproduce steps below.

### Steps to reproduce

```csharp
var input = new Input() {Source = "Hello World!"};

// The following line throws an exception, see above
// Mapper.AssertConfigurationIsValid();

var output = Mapper.Map<TargetOuter>(input)
var equal = output.Inner.MyProp == input.Source; // equal is true
```

jlnostr:
Instead of only checking expression that have been created using `ForMember`, this also respects the `ForPath` configuration as well
Fixes #3000.

### DIFF
--- a/src/AutoMapper/PropertyMap.cs
+++ b/src/AutoMapper/PropertyMap.cs
@@ -143,17 +143,6 @@ namespace AutoMapper
         public void AddValueTransformation(ValueTransformerConfiguration valueTransformerConfiguration) =>
             _valueTransformerConfigs.Add(valueTransformerConfiguration);
 
-        private class MemberFinderVisitor : ExpressionVisitor
-        {
-            public MemberExpression Member { get; private set; }
-
-            protected override Expression VisitMember(MemberExpression node)
-            {
-                Member = node;
-                return base.VisitMember(node);
-            }
-        }
-
         internal void CheckMappedReadonly()
         {
             if(IsResolveConfigured && !ReflectionHelper.CanBeSet(DestinationMember))
--- a/src/AutoMapper/PropertyMap.cs
+++ b/src/AutoMapper/PropertyMap.cs
@@ -75,8 +75,6 @@ namespace AutoMapper
                                   ?? CustomMapExpression?.ReturnType
                                   ?? SourceMember?.GetMemberType();
 
-        public override TypePair Types => IsMapped ? new TypePair(SourceType, DestinationType) : default;
-
         public void ChainMembers(IEnumerable<MemberInfo> members) =>
             _memberChain.AddRange(members as IList<MemberInfo> ?? members.ToList());
 
--- a/src/AutoMapper/PropertyMap.cs
+++ b/src/AutoMapper/PropertyMap.cs
@@ -12,7 +12,7 @@ namespace AutoMapper
     using static Expression;
 
     [DebuggerDisplay("{DestinationMember.Name}")]
-    public class PropertyMap : IMemberMap
+    public class PropertyMap : DefaultMemberMap
     {
         private readonly List<MemberInfo> _memberChain = new List<MemberInfo>();
         private readonly List<ValueTransformerConfiguration> _valueTransformerConfigs = new List<ValueTransformerConfiguration>();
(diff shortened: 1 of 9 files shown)


src/AutoMapper/TypeMap.cs
```diff
@@ -180,11 +180,18 @@ public string[] GetUnmappedPropertyNames()
                     .Where(smc => smc.IsIgnored())
                     .Select(pm => pm.SourceMember.Name).ToList();
 
+                var redirectedPathMembers = PathMaps
```
lbargaoanu: Yes, that's the gist of it. But it's kind of ad-hoc. We have this common interface IMemberMap, so I think implementing smth through that interface would be better. So the end result would be smth like `PropertyMaps.Concat(PathMaps).Where`. Doable I think.

jlnostr: Yeah, the `TypeMap.MemberMaps` property does this very well. However, I don't think that there's a smart solution (or I didn't understand the codebase enough) using the interface.

Mainly because the information about the source types is saved in totally different places. For `PropertyMap`s, I would use `PropertyMap.SourceMember` while for the `PathMap`, I evaluate an expression. Of course, the `TypeMap` property is common to both types, however I personally don't like the idea of using Reflection again, when it was already done at a previous place (in the `PropertyMap._memberChain` list) for one of the types.

lbargaoanu: You can add SourceMember to IMemberMap and implement it in a similar way to PropertyMap.

jlnostr:
I've made some changes to the interface as @lbargaoanu suggested. However, I think that `PropertyMap` can derive from `DefaultMemberMap` to avoid some of the duplicated code.

What do you think about that?

src/AutoMapper/PropertyMap.cs
```diff
@@ -1,165 +1,154 @@
...
-            DestinationMember = destinationMember;
-        }
-
-        public PropertyMap(PropertyMap inheritedMappedProperty, TypeMap typeMap)
-            : this(inheritedMappedProperty.DestinationMember, typeMap) => ApplyInheritedPropertyMap(inheritedMappedProperty);
-
+    using static Expression;
+
+    [DebuggerDisplay("{DestinationMember.Name}")]
+    public class PropertyMap : IMemberMap
```
lbargaoanu: Can you please try fixing the diff somehow? It's probably GitHub's fault, but it does make things difficult.

jlnostr: Seems that there have been a lot of whitespace changes, which here hidden in the IDE. Fix is coming soon, in the meantime, you can disable them on GitHub by setting the diff settings next to the review button.

src/AutoMapper/TypeMap.cs
```diff
@@ -172,13 +172,13 @@ public string[] GetUnmappedPropertyNames()
             }
             else
             {
-                var redirectedSourceMembers = PropertyMaps
-                    .Where(pm => pm.IsMapped && pm.SourceMember != null && pm.SourceMember.Name != pm.DestinationName)
+               var redirectedSourceMembers = MemberMaps
+                    .Where(pm => !pm.Ignored && pm.SourceMember != null && pm.SourceMember.Name != pm.DestinationName)
```
lbargaoanu: This seems wrong. Try adding IsMapped in the interface.

lbargaoanu:
Yes, try deriving PropeertMap from DefaultMap.

PR #3020, opened 2019-04-03
fixed enum values caseinsensitivity bug

PatrickvEk:
Fixed enum value case insensitivity bug

Enum.GetName(sourceEnumType, source), true) the true here means parse case insensitive but the condition above was not case insensitive.


PatrickvEk:
Glad to contribute. :)

PR #3030, opened 2019-04-17
Features cleanup

### LINKED ISSUE #2991: Extensions API

lbargaoanu:
Fixes #2991. Closes #2992.


src/AutoMapper/Features/Features.cs
```diff
@@ -0,0 +1,40 @@
...
+        /// </summary>
+        /// <typeparam name="TFeatureToFind">The type of the feature.</typeparam>
+        /// <returns>The feature or null if feature not exists.</returns>
+        public TFeatureToFind Get<TFeatureToFind>() where TFeatureToFind : TFeature => (TFeatureToFind)_features.GetOrDefault(typeof(TFeatureToFind));
+
+        /// <summary>
+        /// Add or update the feature. Existing feature of the same type will be replaced.
+        /// </summary>
+        /// <param name="feature">The feature.</param>
+        public void AddOrUpdate(TFeature feature) => _features[feature.GetType()] = feature;
```
jbogard: You're going to hate me - but I think I prefer `Set` instead of `AddOrUpdate`. Then it's `Get`/`Set`.

lbargaoanu: But look at the calling code. Weird :)

Issue #3036, opened 2019-04-19
Custom Readonly Generic Typed Collection Property not mapped correctly


THeivers:
### Source/destination types

```csharp
        public class MyCustomTypeCollection<T> : IList<T>
        {
            private List<T> _collection = new List<T>();

            public T this[int index] { get => ((IList<T>)_collection)[index]; set => ((IList<T>)_collection)[index] = value; }

            public int Count => ((IList<T>)_collection).Count;

            public bool IsReadOnly => ((IList<T>)_collection).IsReadOnly;

            public void Add(T item)
            {
                ((IList<T>)_collection).Add(item);
            }

            public void Clear()
            {
                ((IList<T>)_collection).Clear();
            }

            public bool Contains(T item)
            {
                return ((IList<T>)_collection).Contains(item);
            }

            public void CopyTo(T[] array, int arrayIndex)
            {
                ((IList<T>)_collection).CopyTo(array, arrayIndex);
            }

            public IEnumerator<T> GetEnumerator()
            {
                return ((IList<T>)_collection).GetEnumerator();
            }

            public int IndexOf(T item)
            {
                return ((IList<T>)_collection).IndexOf(item);
            }

            public void Insert(int index, T item)
            {
                ((IList<T>)_collection).Insert(index, item);
            }

            public bool Remove(T item)
            {
                return ((IList<T>)_collection).Remove(item);
            }

            public void RemoveAt(int index)
            {
                ((IList<T>)_collection).RemoveAt(index);
            }

            IEnumerator IEnumerable.GetEnumerator()
            {
                return ((IList<T>)_collection).GetEnumerator();
            }
        }

        public class SourceItem
        {
            public string Name { get; set; }
            public List<string> ShipsTo { get; set; }
        }

        public class DestItem
        {
            public string Name { get; set; }
            public MyCustomTypeCollection<string> ShipsTo { get; } = new MyCustomTypeCollection<string>();
        }
```

### Mapping configuration

```csharp
new MapperConfiguration(cfg => cfg.CreateMap<SourceItem, DestItem>()
            .ForMember(d => d.ShipsTo, m => m.UseDestinationValue()));
```

### Expected behavior
ShipsTo on DestItem should have converted items from the source item.

### Actual behavior
No items are added to the destination collection. Configuration.AssertConfigurationIsValid() also does not throw any exceptions.

Adding MapFrom and manually pointing it to the source property does work. But in that case I would expect an exception during validation.

[...]

PR #3037, opened 2019-04-19
Potential fix for Custom Readonly Generic Typed Collections

THeivers:
Summary of the changes:
- Expanded PrimitiveHelper.IsListType to include the generic list interface
- Added extra unit test to check IsListType on a custom collection type that just implements `IList<T>`(but not `IList`)
- Added unit test to confirm fix for #3036

#3036


lbargaoanu:
Fixed by https://github.com/AutoMapper/AutoMapper/pull/2928.

lbargaoanu:
I think the original intent here might have been to special case read-only collection properties so they work by default (without MapFrom or UseDestinationValue). Now the MapFrom is required except for the special cases. Maybe it would make sense to remove the special case or to implement the original intent. @jbogard might want a PR in the later direction.

PR #3041, opened 2019-04-22
Refactor feature extensions

### LINKED ISSUE #3039: AddOrUpdateFeature extensions needed?
I don't think we need the FeatureExtensions class - instead, like with ASP.NET Core, those feature implementors will call `Features.Set` themselves.

lbargaoanu:
Fixes #3039. I say, let's keep them, but whatever works.


lbargaoanu:
Those derived classes seemed kind of anemic, so I went with extension methods.

PR #3044, opened 2019-04-25
AllowNullDestinationValues function strengthen

hyless:
When the AllowNullDestinationValues is false, the property type of string, and the resource property type is not included , set the default value as empty.


hyless:
When the AllowNullDestinationValues is false, the property type of string, and the resource property type is not included , set the default value as empty.

PR #3050, opened 2019-04-26
Enabled multiple usage of AutoMapAttribute

### LINKED ISSUE #3049: Allow multiple uses of AutoMapAttribute
We frequently define maps from multiple sources to a single destination. It'd be great, and clean up our code substantially, to be able to do so using the new attribute mapping, but the `AutoMapAttribute` class doesn't allow multiple usage as the relevant flag in `AttributeUsage` is not set.

As an example, we may wish to instantiate a Command class from either an API call model, or a UI View Model:

```csharp
public class Command()
{
    // members

    public class AutoMapperProfile : Profile
    {
        public AutoMapperProfile()
        {
            CreateMap<ApiModel, Command>();
            CreateMap<ViewModel, Command>();
        }
    }
}
```

Ideally, we'd like to be able to do

```csharp
[AutoMap(typeof(ApiModel))]
[AutoMap(typeof(ViewModel))]
public class Command()
{
    // members
}
```

I've had a very quick look, and it appears that setting the appropriate flag on the `AutoMapAttribute.cs` class and changing how those attributes are loaded in `MapperConfigurationExpression.cs` is what's required to allow this.

Sharkwald:
Fixes #3049.
This is a rough first pass to enable this. All existing tests are green, but it needs some specifically covering the multiple attribute use scenario.

### DIFF
--- a/src/AutoMapper/Configuration/MapperConfigurationExpression.cs
+++ b/src/AutoMapper/Configuration/MapperConfigurationExpression.cs
@@ -109,20 +109,17 @@ namespace AutoMapper.Configuration
                 var autoMapAttributes = type.GetCustomAttributes<AutoMapAttribute>();
                 foreach (var autoMapAttribute in autoMapAttributes)
                 {
-                    if (autoMapAttribute != null)
-                    {
-                        var mappingExpression =
-                            (MappingExpression) autoMapAttributeProfile.CreateMap(autoMapAttribute.SourceType, type);
-                        autoMapAttribute.ApplyConfiguration(mappingExpression);
+                    var mappingExpression =
+                        (MappingExpression) autoMapAttributeProfile.CreateMap(autoMapAttribute.SourceType, type);
+                    autoMapAttribute.ApplyConfiguration(mappingExpression);
 
-                        foreach (var memberInfo in type.GetMembers(BindingFlags.Public | BindingFlags.Instance))
+                    foreach (var memberInfo in type.GetMembers(BindingFlags.Public | BindingFlags.Instance))
+                    {
+                        foreach (var memberConfigurationProvider in memberInfo.GetCustomAttributes()
+                            .OfType<IMemberConfigurationProvider>())
                         {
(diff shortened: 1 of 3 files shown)


Sharkwald:
Added a simple test to describe the scenario I have in mind. Are any more required?

src/AutoMapper/Configuration/MapperConfigurationExpression.cs
```diff
@@ -106,17 +106,23 @@ private void AddMapsCore(IEnumerable<Assembly> assembliesToScan)
...
-                var autoMapAttribute = type.GetCustomAttribute<AutoMapAttribute>();
-                if (autoMapAttribute != null)
+                var autoMapAttributes = type.GetCustomAttributes<AutoMapAttribute>();
+                foreach (var autoMapAttribute in autoMapAttributes)
                 {
-                    var mappingExpression = (MappingExpression) autoMapAttributeProfile.CreateMap(autoMapAttribute.SourceType, type);
-                    autoMapAttribute.ApplyConfiguration(mappingExpression);
-
-                    foreach (var memberInfo in type.GetMembers(BindingFlags.Public | BindingFlags.Instance))
+                    if (autoMapAttribute != null)
```
lbargaoanu: I think the null check can be removed.

Sharkwald: So it can.

Sharkwald:
The test failures are being caused by the change to the test base class; because the name is not being specified (as `nameof(AutoMapAttribute)`) it seems to be validating globally and tripping on some failure conditions set elsewhere in the test code.

lbargaoanu:
I'll take a look.

lbargaoanu:
I just reverted :) I guess those scanning tests should live in their own assembly, because now it just scans everything.

Sharkwald:
Brilliant :D.

lbargaoanu:
We need your signature :)

Sharkwald:
Apologies, that should be it now!

PR #3075, opened 2019-05-10
Fix small bug in constructor

flash2048:
In the constructor of the class confused variables.
In the version with an error, the `When_scanning_by_name` class `Should_load_profiles` test fails.


lbargaoanu:
What happens here is that the scanning tests scan all the types in the assembly. What would help here is isolating the scanning tests in their own assembly so this way at least they only interfere with each other. A PR is welcome in that direction.

PR #3080, opened 2019-05-16
Get ToString with zero parameters in FlagsEnumMapper

### LINKED ISSUE #3079: FlagsEnumMapper sometimes throws an exception
### Source/destination types

```csharp
public class A
{
	public Flag F { get; set; }
}

public class B
{
	public Flag F { get; set; }
}

[Flags]
public enum Flag
{
	O = 0,
	A = 1,
	B = 2
}
```

### Mapping configuration

```csharp
var mapper = new Mapper(new MapperConfiguration(config =>
{
	config.CreateMap<A, B>();
}));
```

### Version: 8.1.0
 
.net 4.6.2

### Expected behavior

Works as expected.

### Actual behavior

**Sometimes** exception is thrown
```csharp
System.ArgumentException: Incorrect number of arguments supplied for call to method 'System.String ToString(System.String, System.IFormatProvider)' at
System.Linq.Expressions.Expression.ValidateArgumentCount(MethodBase method, ExpressionType nodeKind, Int32 count, ParameterInfo[] pis) at
System.Linq.Expressions.Expression.ValidateArgumentTypes(MethodBase method, ExpressionType nodeKind, ReadOnlyCollection`1& arguments) at
System.Linq.Expressions.Expression.Call(Expression instance, MethodInfo method, IEnumerable`1 arguments) at
AutoMapper.Mappers.FlagsEnumMapper.MapExpression(IConfigurationProvider configurationProvider, ProfileMap profileMap, IMemberMap memberMap, Expression sourceExpression, Expression destExpression, Expression contextExpression) at
AutoMapper.Execution.ExpressionBuilder.MapExpression(IConfigurationProvider configurationProvider, ProfileMap profileMap, TypePair typePair, Expression sourceParameter, Expression contextParameter, IMemberMap propertyMap, Expression destinationParameter) at
...
```
### Steps to reproduce

The problem is in **FlagsEnumMapper**, namelly, in following call
```csharp
Call(sourceExpression, sourceExpression.Type.GetDeclaredMethod("ToString"))
```
GetDeclaredMethod calls extension method GetRuntimeMethods to get list of methods
```csharp
public static IEnumerable<MethodInfo> GetRuntimeMethods(this Type type)
{
   RuntimeReflectionExtensions.CheckAndThrow(type);
   return (IEnumerable<MethodInfo>) type.GetMethods(BindingFlags.Instance | BindingFlags.Static | BindingFlags.Public | BindingFlags.NonPublic);
}
```
GetMethods documentation https://docs.microsoft.com/en-us/dotnet/api/system.type.getmethods?view=netframework-4.6.2
> The GetMethods method does not return methods in a particular order, such as alphabetical or declaration order. Your code must not depend on the order in which methods are returned, because that order varies.

So sometimes GetDeclaredMethod("ToString") returns other method than expected(e.g. ToString(string format, IFormatProvider provider) instead of ToString()) and 
 **Incorrect number of arguments supplied for call to method** exception is thrown.

pdudnikov:
Fixes #3079.
 - modify FlagsEnumMapper to get ToString method with zero parameters


lbargaoanu:
The test should be your use case (check the other tests). And here we need the ToString without parameters, so you should explicitly request that one (there is another GetDeclaredMethod overload that does that).

lbargaoanu:
Or, even better I think, remove the overload without type parameters and add `params` to the other overload, so not specifying parameters means look for an overload without parameters.

pdudnikov:
> Or, even better I think, remove the overload without type parameters and add `params` to the other overload, so not specifying parameters means look for an overload without parameters.

GetDeclaredMethod overload without type parameters is used in other places to get method by name regardless of parameters count.

lbargaoanu:
Yes, and they either mean zero parameters, or they should explicitly say which overload they want. But this way works too. A test is missing.

PR #3102, opened 2019-05-28
Deprecate CreateMissingTypeMaps

lbargaoanu:


src/AutoMapper/IMapperConfigurationExpression.cs
```diff
@@ -11,7 +11,8 @@ public interface IMapperConfigurationExpression : IProfileExpression
         /// <summary>
         /// Create missing type maps during mapping, if necessary
         /// </summary>
-        bool? CreateMissingTypeMaps { get; set; }
+        [Obsolete("Support for automatically created maps will be removed.")]
```
jbogard: Let's include the version where it will be removed, 9.0, and include the fix, "you will need to explicitly configure maps".

emperador-ming:
Oh my god! What are we supposed to do with anonymous classes projection? O_O


```
return
(
	from person in People
	from scheduleDayBasis in person.Schedule.ScheduleDayBasis 
	from begining in this.GetSchedule(scheduleDayBasis .id_schedule)
	where person.initials== initials && schedule.DaysOfTheWeek.day == dayOfTheWeek
	select new
	{
		initials,		
		beginig.beginingDate,		
		minutes = begining.minutesOfInterruption ?? 0
	}
).ProjectTo<Schedule>(mapper.ConfigurationProvider);
````

**Comment by lbargaoanu** [comment]
You should project directly to some DTO.

**Comment by emperador-ming** [comment]
> You should project directly to some DTO.

OK. I thought it was a bad practice or prone to error (N+1 problem and so on).

**Comment by mlaily** [comment]
What's the motivation behind this change? Isn't AutoMapper supposed to remove tedious explicit configuration like that?

**Comment by jbogard** [comment]
@mlaily no, it was *never* about removing explicit configuration. It was about enforcing conventions:

https://jimmybogard.com/automappers-design-philosophy/

Implicit configuration means we can't enforce design conventions. I never wanted that, it's very dangerous to do (you should ALWAYS have `AssertConfigurationIsValid`) and regret ever putting it in.

If you want lower config, check out attribute maps. But I'm reinforcing explicit opt-in config as it's the only safe way to map.

**Comment by mlaily** [comment]
Thanks, I was actually reading the post you linked when I saw your reply :)

It (and many other things) makes a lot more sense now.

BTW, attribute based mapping doesn't work well for us because we are using AutoMapper in a desktop application, so everything is reversed, and I don't want to put AutoMap attributes in my business objects.

Anyway, I respect your decision. The only regret I have is that you are only now being more restrictive about this, and deprecating a lot of stuff people came to rely on.

**Comment by jbogard** [comment]
There was an alternative - assembly scanning - that would create the maps via conventions. So AutoMapper scans and looks for types with attributes to create maps, but we could support other means as well. The previous implementation used dynamic maps, but we're still open to scanning via conventions (like EF Core does).

**Comment by fuine6** [comment]
I only use AutoMapper when the source and destination have the same properties.
I don't want explicitly create the maps.
If it is unmapped, AutoMapperConfigurationException is enough for me.

**Comment by jbogard** [comment]
@fuine6 What does "unmapped" mean? A single property? You'd rather have a runtime exception on a map than a single unit test to check all maps?

**Comment by fuine6** [comment]
> What does "unmapped" mean? A single property? You'd rather have a runtime exception on a map than a single unit test to check all maps?

Sorry, My English is poor.
I show a example here:

```
public class CommamdDto
{
}

public class Commamd
{

	public string Email { get; set; }
}

var commamdDto = new CommamdDto { UserName = "test" };
var command = _mapper.Map<Commamd>(commamdDto);// <---here will throw AutoMapperConfigurationException.
```
I understand that this is unsafe.
It happens at runtime.
The best way is as you said, explicitly create the maps.

But we have some legacy code, if I update 8.1.0 to 8.1.1, it will break.
And then, I need to configure CreateMissingTypeMaps = true.

It show the message below:
```
Support for automatically created maps will be removed in version 9.0. You will need to explicitly configure maps, manually or using reflection. Also consider attribute mapping
```

If it is really removed in version 9.0.
We have the following options:
1. Modify all code, it has a very large number, If there is something missing, it is also a runtime exception, this will also happen when developer forget to CreateMap.
2. Use version 8.0 and do not update, This does not seem to be a good way.

I wish to keep the settings instead of removing them.
In our new project, we will follow the explicitly configuration.
But we need the past project to run correctly.

**Comment by henkmollema** [comment]
Why was this setting changed in a minor version? It causes a lot of runtime exceptions for missing mappings which used to be created implicitly in 8.1.0.

**Comment by lbargaoanu** [comment]
@henkmollema You're not paying attention.

**Comment by henkmollema** [comment]
Please enlighten me then 😄

**Comment by henkmollema** [comment]
This sample works with 8.1.0 and throws an exception for 8.1.1:

```cs
class Program
{
    {
        var configuration = new MapperConfiguration(_ => { });
        var mapper = configuration.CreateMapper();

        var model = new FooDomainModel { Foo = "Hello World" };
        var dto = mapper.Map<FooDto>(model);
        Console.WriteLine(dto.Foo);
    }
}

public class FooDomainModel
{
}

public class FooDto
{
}
```

jbogard:
Yeah, that was a miss, we shouldn't have made a behavior change in a minor release.

@lbargaoanu is it worth reverting the behavior change *only* for an 8.1.2 release?

lbargaoanu:
It seems to me that most of the people complaining don't even know that `CreateMissingTypeMaps` exists. Reverting would only create confusion. And besides, why would you upgrade? There is nothing in 8.1.1 except the first step to removing stuff in 9.0.0.

henkmollema:
"Complaining" might be a bit exaggerating, I'm just pointing out that a patch version introduced a breaking change during runtime and I wondered if it was intentional or not. Whether it's caused by `CreateMissingTypeMaps` seems irrelevant to me.

lbargaoanu:
@jbogard Maybe just release 9.0? :)

jbogard:
@henkmollema No, it wasn't intentional when I reviewed it - I'd like to not to introduce breaking behavioral changes in bugfix releases (even if the API remains unchanged).

And yeah, perhaps just release 9.0!

Issue #3105, opened 2019-05-30
ResolutionContext unavailable in IMappingAction


hallidev:
Consider the following .NET Core 2.2 Console App:

```
    class Program
    {
        static void Main(string[] args)
        {
            var config = new MapperConfiguration(cfg => {
                cfg.CreateMap<SomePersonObject, SomeOtherPersonObject>()
                    .AfterMap<NameMeJohnAction>();
            });

            var bob = new SomePersonObject {Name = "Bob"};

            var mapper = config.CreateMapper();

            // I want the name "Pete" available in AfterMap
            var john = mapper.Map<SomeOtherPersonObject>(bob, opt => opt.Items["OverrideName"] = "Pete");
            
        }
    }

    public class SomePersonObject
    {
        public string Name { get; set; }
    }

    public class SomeOtherPersonObject
    {
        public string Name { get; set; }
    }

    public class NameMeJohnAction : IMappingAction<SomePersonObject, SomeOtherPersonObject>
    {
        public void Process(SomePersonObject source, SomeOtherPersonObject destination)
        {
            // I want to check "OverrideName" and change the name to "Pete", but I can't!
            destination.Name = "John";
        }
    }
```

### Version: x.y.z
AutoMapper 8.0+

### Expected behavior
Some way to access ResolutionContext in AfterMap

### Actual behavior
No way to access ResolutionContext in AfterMap

### Steps to reproduce
Paste above class into blank .NET Core Console App and run

lbargaoanu:
A PR is welcome. Otherwise, you'll have to use a version of AfterMap that does receive the context.

hallidev:
Is there currently a version of AfterMap that allows for DI as well as
ResolutionContext?

In my actual use case, I'm using DI but also need access to
ResolutionContext.

If not, I can try to put a PR together, but my initial thought is to add a
third parameter to IMappingAction.Process(), which would be a breaking
change to the interface. Any idea on how you'd want it to work?

lbargaoanu:
Not that I'm aware of, but you can always workaround it. I don't think a breaking change is a good idea, but @jbogard might disagree. I would add a new interface that takes a structure as parameter. That structure would have as members all the parameters. The way the ForPath->Condition works. Then adding another parameter would not be a breaking change.

jbogard:
Ugh, this was just an oversight really. I'm happy to take a PR, but it'd have to wait for the next major release.

hallidev:
[...]

PR #3106, opened 2019-05-31
Adding ResolutionContext to IMappingAction<TSource, TDestination>.Process()

hallidev:
```
Adding ResolutionContext to IMappingAction<TSource, TDestination>.Process()

#3105
```


lbargaoanu:
@jbogard If we go this route, next time we add a parameter, it will be another breaking change.

hallidev:
It seems like ResolutionContext is meant to be extended to avoid breaking changes. As @jbogard mentioned in #3105, this looks like it may have just been an oversight.

I'm also struggling with what a different interface would be called to achieve the same functionality at the same spot.

hallidev:
Why was this closed? Are there plans to address this in a different way?

jbogard:
Yeah all the interfaces like this are supposed to have the ResolutionContext, I just missed this one.

hallidev:
Test added

PR #3130, opened 2019-06-18
Automate updating last version binary

lbargaoanu:
@jbogard I guess what's missing here is authentication, like [this](https://www.appveyor.com/docs/how-to/git-push/). So it makes sense only on merge. Thoughts?

### DIFF
--- a/default.ps1
+++ b/default.ps1
@@ -14,7 +14,7 @@ properties {
 
 
 task default -depends local
-task local -depends compile, test
+task local -depends compile, copy_last_major_version, test
 task ci -depends clean, release, local, pack, benchmark
 
 task clean {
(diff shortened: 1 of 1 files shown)


jbogard:
I think we can do all this via command line? The dotnet cli lets us create a project and add a Nuget package reference and build.

jbogard:
Then we wouldn’t need to commit the binary at all, it could just grab the package from NuGet.

lbargaoanu:
But then we still need some version of the binary in the repo for things to work locally. And the results would be different than on CI. A little weird. I'll give it a shot.

lbargaoanu:
I think the best way would be using nuget.exe in a prebuild action. It works the same locally and on CI and it doesn't require the binary in source control. But it needs some logic. 9.0 would be compared with 8.0 and 8.1 also with 8.0. I started with some bat files and that didn't work out :) I guess is takes some PowerShell skills.

Tasteful:
You can query the git for the tags and from that figure our the last version to check with.
```
$version_tags = (git tag -l) | Where-Object { $_ -match "^v\d+\.\d+\.\d+$" } | Sort-Object { [Version]$_ };
```

lbargaoanu:
The logic is simple enough, but try writing that in a batch file :) That's why I think PowerShell is needed.

	if(version.Minor == 0 && version.Build == 0){
		var newVersion = new Version(version.Major - 1, 0, 0).Dump();
	}else{
		var newVersion = new Version(version.Major, 0, 0).Dump();
	}

PR #3134, opened 2019-06-19
Remove dynamic maps

### LINKED ISSUE #3063: Remove dynamic maps
After having these out for a bit, is there any real value to them? They seem to cause more problems than they solve.

I don't use them, none of our teams do, and they go against the [original design philosophy](https://jimmybogard.com/automappers-design-philosophy/).

This would mean convention-based stuff would go too, to be replaced with scanning-based conventions. The only late-created maps I'd want are closing generic types.

- **Deprecate CreateMissingTypeMaps and change the default to false #3102**

jbogard:
Removes all means of creating dynamic/inline maps. Fixes #3063


lbargaoanu:
I don't think deleting tests is a good idea. Because we still have dynamic mapping with generics. And most of the issues apply. Especially concurrency issues and recursive mapping. I'm not saying it's trivial to rewrite those tests with generics, but I think it's time well spent.

jbogard:
We've got a test for the generics thing already I thought? The same file, `ResolveWithGenericMap`.

lbargaoanu:
That's true. But for example ResolveTypeMapThreadingIssues has no generic counterpart, right? I think we should check every case.

lbargaoanu:
ConventionCreateMapsWithCircularReference is another case.

jbogard:
I dunno about the convention one - that would resolve to `Mapper.Map`

lbargaoanu:
So we're talking about ConventionCreateMapsWithCircularReference. Yes, it would just call Map, but it would have to create those closed generics maps that reference each other. Or I'm completely missing your point :)

jbogard:
@lbargaoanu added back the tests with generic flavor.

lbargaoanu:
Dynamic_constructor_mapping. Maybe this one is too much, I don't know :)

jbogard:
OK added one test for that last one.

LAST ONE

lbargaoanu:
All right :) So, maybe skip the failing tests at this time? :)

jbogard:
Sigh

jbogard:
Ah these look like proper failing tests - it's generic types containing generic members.

lbargaoanu:
So, valuable tests, I guess :)

jbogard:
There's still a failing test but I commented it out

lbargaoanu:
I think skipping it is better, as it's less likely that we'll ignore it.

jbogard:
Can’t, it’s the config validation

PR #3141, opened 2019-06-24
Moving enums after assignable to simplify assignable enums

jbogard:
And adding a little code comment on the order of mappers so I don't forget how these are ordered and grouped.


lbargaoanu:
Delete the branch?

PR #3144, opened 2019-06-25
Automate updating last version binary

lbargaoanu:


lbargaoanu:
Can we always have a link to the compat file of a release in the release notes? If not, I can simply add a docs page so people know where to look.

jbogard:
Yeah that works for me.

PR #3156, opened 2019-07-05
ForEach - IEnumerator<T>.Dispose is not called

### LINKED ISSUE #3154: ForEach - IEnumerator<T>.Dispose is not called
Hi,

we have implemented a custom collection which heavily relies on `IEnumerator<T>.Dispose()` being called.

The `AutoMapper.Internal.ExpressionFactory.ForEach` method does not call Dispose even if the enumeratorVar is of type IDisposable (maybe because it is a [`IEnumerator<T>`](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.ienumerator-1)).

According to Mark Michaelis (author of the [Essential C# books](https://www.amazon.co.uk/Mark-Michaelis/e/B001IOFCFA/)), the enumerator should be checked if it implements `IDisposable` and if, `Dispose` should be called. See [Understanding C# foreach Internals and Custom Iterators with yield](https://msdn.microsoft.com/en-us/magazine/mt797654.aspx), figure 3

```csharp
System.Collections.Generic.Stack<int> stack = new System.Collections.Generic.Stack<int>();
System.Collections.Generic.Stack<int>.Enumerator enumerator;
IDisposable disposable;
enumerator = stack.GetEnumerator();
try
{
  int number;
  while (enumerator.MoveNext())
  {
    number = enumerator.Current;
    Console.WriteLine(number);
  }
}
finally
{
  // Explicit cast used for IEnumerator<T>.
  disposable = (IDisposable) enumerator;
  disposable.Dispose();
  // IEnumerator will use the as operator unless IDisposable
  // support is known at compile time.
  // disposable = (enumerator as IDisposable);
  // if (disposable != null)
  // {
  //   disposable.Dispose();
  // }
}
```

### Expected behavior

Please call `Dispose` after each completed foreach loop if the enumerator implements it.

### Actual behavior

`Dispose` is not called, even if it is implemented.

### Steps to reproduce

```csharp
using System.Collections;
using System.Collections.Generic;
using System.Diagnostics;
using System.Linq;
using AutoMapper;

namespace AutoMapperForEach
{
    public static class Program
    {
        #region Methods

        private static void Main(string[] args)
        {
            MapperConfiguration configuration = new MapperConfiguration(cfg => cfg.CreateMap<MyClass, MyClass>());
            MyClass my1 = new MyClass
            {
                Numbers = new MyCollection<int> { 1, 2, 3 }
            };
            IMapper mapper = configuration.CreateMapper();
            MyClass my2 = mapper.Map<MyClass>(my1);
            Debug.Assert(my1.Numbers.SequenceEqual(my2.Numbers));
        }

        #endregion
    }

    public class MyClass
    {
        #region Properties

        public MyCollection<int> Numbers { get; set; }

        #endregion
    }

    public class MyCollection<T> : IList<T>
    {
        #region Fields

        private readonly List<T> _collection = new List<T>();

        #endregion

        #region Finalizers / Dispose

        private class MyEnumerator : IEnumerator<T>
        {
            #region Fields

            private readonly IList<T> _collection;

            private int i = -1;

            #endregion

            #region Constructors

            public MyEnumerator(IList<T> collection)
            {
                _collection = collection;
            }

            #endregion

            #region Finalizers / Dispose

            public void Dispose()
            {
                // called by SequenceEqual
                // never called my AutoMapper
            }

            #endregion

            #region IEnumerator<T> Members

            public bool MoveNext() => ++i < _collection.Count;

            public void Reset()
            {
            }

            object IEnumerator.Current => Current;

            public T Current => _collection[i];

            #endregion
        }

        #endregion

        #region IList<T> Members

        public void Add(T item)
        {
            _collection.Add(item);
        }

        public void Clear()
        {
            _collection.Clear();
        }

        public bool Contains(T item)
        {
            return _collection.Contains(item);
        }

        public void CopyTo(T[] array, int arrayIndex)
        {
        }

        public bool Remove(T item)
        {
            return _collection.Remove(item);
        }

        public int Count
        {
            get { return _collection.Count; }
        }

        public bool IsReadOnly
        {
            get { return false; }
        }

        IEnumerator IEnumerable.GetEnumerator()
        {
            return GetEnumerator();
        }

        public IEnumerator<T> GetEnumerator()
        {
            return new MyEnumerator(this);
        }

        public int IndexOf(T item)
        {
            return _collection.IndexOf(item);
        }

        public void Insert(int index, T item)
        {
            _collection.Insert(index, item);
        }

        public void RemoveAt(int index)
        {
            _collection.RemoveAt(index);
        }

        public T this[int index]
        {
            get { return _collection[index]; }
            set { _collection[index] = value; }
        }

        #endregion
    }
}
```

buddde:
ForEach calls Dispose on enumerator. Fixes #3154.

### DIFF
--- a/src/AutoMapper/Internal/ExpressionFactory.cs
+++ b/src/AutoMapper/Internal/ExpressionFactory.cs
@@ -57,7 +57,7 @@ namespace AutoMapper.Internal
             var breakLabel = Label("LoopBreak");
 
             Expression disposeCall;
-            if (enumeratorType.GetInterface("System.IDisposable", true) == null)
+            if (!typeof(IDisposable).IsAssignableFrom(enumeratorType))
             {
                 // Try to dispose at runtime
                 var disposableVar = Variable(typeof(IDisposable), "disposable");
--- a/src/AutoMapper/Internal/ExpressionFactory.cs
+++ b/src/AutoMapper/Internal/ExpressionFactory.cs
@@ -56,29 +56,54 @@ namespace AutoMapper.Internal
 
             var breakLabel = Label("LoopBreak");
 
-            var disposableVar = Variable(typeof(IDisposable), "disposable");
-            Expression tryDisposeCall = Block(new[] { disposableVar },
-                Assign(disposableVar, TypeAs(enumeratorVar, typeof(IDisposable))),
-                IfThen(NotEqual(disposableVar, Constant(null, typeof(object))),
-                    Call(disposableVar, disposableVar.Type.GetInheritedMethod(nameof(IDisposable.Dispose)))));
-
-            var loop = Block(new[] { enumeratorVar },
-                enumeratorAssign,
-                TryFinally(
-                    Loop(
-                        IfThenElse(
-                            Equal(moveNextCall, Constant(true)),
-                            Block(new[] { loopVar },
-                                Assign(loopVar, ToType(Property(enumeratorVar, "Current"), loopVar.Type)),
-                                loopContent
+            if (enumeratorType.GetInterface("System.IDisposable", true) != null)
+            {
+                // always dispose
+                var loop = Block(new[] { enumeratorVar },
+                    enumeratorAssign,
+                    TryFinally(
+                        Loop(
+                            IfThenElse(
+                                Equal(moveNextCall, Constant(true)),
+                                Block(new[] { loopVar },
+                                    Assign(loopVar, ToType(Property(enumeratorVar, "Current"), loopVar.Type)),
+                                    loopContent
+                                ),
+                                Break(breakLabel)
                             ),
-                            Break(breakLabel)
-                        ),
-                        breakLabel),
-                    tryDisposeCall)
-            );
+                            breakLabel),
+                        Call(enumeratorVar, enumeratorVar.Type.GetInheritedMethod(nameof(IDisposable.Dispose))))
+                );
 
-            return loop;
+                return loop;
+            }
+            else
+            {
+                // Dispose at runtime
+                var disposableVar = Variable(typeof(IDisposable), "disposable");
+                Expression tryDisposeCall = Block(new[] { disposableVar },
+                    Assign(disposableVar, TypeAs(enumeratorVar, typeof(IDisposable))),
+                    IfThen(NotEqual(disposableVar, Constant(null, typeof(object))),
+                        Call(disposableVar, disposableVar.Type.GetInheritedMethod(nameof(IDisposable.Dispose)))));
+
+                var loop = Block(new[] { enumeratorVar },
+                    enumeratorAssign,
+                    TryFinally(
+                        Loop(
+                            IfThenElse(
+                                Equal(moveNextCall, Constant(true)),
+                                Block(new[] { loopVar },
+                                    Assign(loopVar, ToType(Property(enumeratorVar, "Current"), loopVar.Type)),
+                                    loopContent
+                                ),
+                                Break(breakLabel)
+                            ),
+                            breakLabel),
+                        tryDisposeCall)
+                );
+
+                return loop;
+            }
         }
 
         public static Expression ForEachArrayItem(Expression array, Func<Expression, Expression> body)
--- a/src/AutoMapper/Internal/ExpressionFactory.cs
+++ b/src/AutoMapper/Internal/ExpressionFactory.cs
@@ -56,18 +56,26 @@ namespace AutoMapper.Internal
 
             var breakLabel = Label("LoopBreak");
 
+            var disposableVar = Variable(typeof(IDisposable), "disposable");
+            Expression tryDisposeCall = Block(new[] { disposableVar },
+                Assign(disposableVar, TypeAs(enumeratorVar, typeof(IDisposable))),
+                IfThen(NotEqual(disposableVar, Constant(null, typeof(object))),
+                    Call(disposableVar, disposableVar.Type.GetInheritedMethod(nameof(IDisposable.Dispose)))));
+
             var loop = Block(new[] { enumeratorVar },
                 enumeratorAssign,
-                Loop(
-                    IfThenElse(
-                        Equal(moveNextCall, Constant(true)),
-                        Block(new[] { loopVar },
-                            Assign(loopVar, ToType(Property(enumeratorVar, "Current"), loopVar.Type)),
-                            loopContent
+                TryFinally(
+                    Loop(
+                        IfThenElse(
+                            Equal(moveNextCall, Constant(true)),
+                            Block(new[] { loopVar },
+                                Assign(loopVar, ToType(Property(enumeratorVar, "Current"), loopVar.Type)),
+                                loopContent
+                            ),
+                            Break(breakLabel)
                         ),
-                        Break(breakLabel)
-                    ),
-                breakLabel)
+                        breakLabel),
+                    tryDisposeCall)
             );
 
             return loop;
--- a/src/AutoMapper/Internal/ExpressionFactory.cs
+++ b/src/AutoMapper/Internal/ExpressionFactory.cs
@@ -1,5 +1,6 @@
 ﻿using System;
-using System.Diagnostics;
+using System.Collections;
+using System.Collections.Generic;
 using System.Linq;
 using System.Linq.Expressions;
 using System.Reflection;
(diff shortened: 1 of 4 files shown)


lbargaoanu:
There need to be at least two tests. One for the success case and one for the exception case.

src/AutoMapper/Internal/ExpressionFactory.cs
```diff
@@ -56,18 +56,26 @@ public static Expression ForEach(Expression collection, ParameterExpression loop
 
             var breakLabel = Label("LoopBreak");
 
+            var disposableVar = Variable(typeof(IDisposable), "disposable");
+            Expression tryDisposeCall = Block(new[] { disposableVar },
+                Assign(disposableVar, TypeAs(enumeratorVar, typeof(IDisposable))),
```
lbargaoanu: This is a boxing conversion. Also an information known beforehand because we can check `enumeratorType` and check for a `Dispose` method. If it exists, add the call.

buddde: How can you check the `enumeratorType` for a `Dispose` method?

I tried it, but if the `GetEnumerator` method returns `IEnumerator` (instead of `IEnumerator<T>`) but the runtime enumerator is of type `IEnumerator<T>` and hance `IDisposable` you better check the concrete type at runtime.

I do not know of any way to do the runtime check without boxing.

lbargaoanu: Don't check the runtime type. Just use `enumeratorType`. In a proper implementation the return type of `GetEnumerator` should be a more specific type.

buddde:
```
public class Demo : IEnumerable
{
    public IEnumerator GetEnumerator()
    {
        return new EnumeratorDisposable();
        return new EnumeratorOfT();
    }

    private class EnumeratorDisposable : IEnumerator, IDisposable
    {
        public bool MoveNext() => true;

        public void Reset() { }

        public object Current { get; }

        public void Dispose() { }
    }

    private class EnumeratorOfT : IEnumerator<object>, IDisposable
    {
        public bool MoveNext() => true;

        public void Reset() { }

        public object Current { get; }

        public void Dispose() { }
    }
}

```
In these two cases, the `Dispose` method would not be called.
That is not what I want.

lbargaoanu: Having a struct as enumerator is a well known optimization, used throughout the BCL. Your code effectively throws that away. Check the IL generated by the compiler and you'll see there's no boxing there.

buddde: Oh, i did not think of `struct`s.

But, as far as I can see, the IL generated only optimizes the boxing away in case there is an `IEnumerator<T>`. In case if an `IEnumerator`, it casts and boxes (like me).

Maybe I can make two different loops:

- a loop optimized for `IEnumerator<T>`
- a loop with runtime check for `IEnumerator`

What do you think?

buddde: I updated the PR. Can you please check it again. Thank you.

The missing tests will be added as soon as the changes in the ForEach method are ok for you.

src/AutoMapper/Internal/ExpressionFactory.cs
```diff
@@ -56,18 +56,37 @@ public static Expression ForEach(Expression collection, ParameterExpression loop
 
             var breakLabel = Label("LoopBreak");
 
+            Expression disposeCall;
+            if (enumeratorType.GetInterface("System.IDisposable", true) == null)
```
lbargaoanu: There is no need for strings here.

src/AutoMapper/Internal/ExpressionFactory.cs
```diff
@@ -56,18 +56,37 @@ public static Expression ForEach(Expression collection, ParameterExpression loop
...
+            {
+                // Try to dispose at runtime
+                var disposableVar = Variable(typeof(IDisposable), "disposable");
+                disposeCall = Block(new[] { disposableVar },
+                    Assign(disposableVar, TypeAs(enumeratorVar, typeof(IDisposable))),
+                    IfThen(NotEqual(disposableVar, Constant(null, typeof(object))),
+                        Call(disposableVar, disposableVar.Type.GetInheritedMethod(nameof(IDisposable.Dispose)))));
+
```
lbargaoanu: There is no need for strings here.

src/AutoMapper/Internal/ExpressionFactory.cs
```diff
@@ -56,18 +56,37 @@ public static Expression ForEach(Expression collection, ParameterExpression loop
...
+            }
+            else
+            {
+                // Always dispose directly
+                disposeCall = Call(enumeratorVar, enumeratorVar.Type.GetInheritedMethod(nameof(IDisposable.Dispose)));
+            }
```
lbargaoanu: There is no need for strings here.

src/AutoMapper/Internal/ExpressionFactory.cs
buddde: Ok, so how do I call the `Dispose` method?

src/AutoMapper/Internal/ExpressionFactory.cs
buddde: Ok, so how do I call the `Dispose` method?

src/AutoMapper/Internal/ExpressionFactory.cs
buddde: Ok, I change that.

src/AutoMapper/Internal/ExpressionFactory.cs
lbargaoanu: It should be the same on both branches, smth similar to [this](https://github.com/AutoMapper/AutoMapper/blob/f11fa0477e8698e1659e4dad993b041d8488c8cd/src/AutoMapper/Execution/TypeMapPlanBuilder.cs#L214).

buddde: Thank you for your input, but I do not know how to apply this to my problem.

lbargaoanu: OK, just carry on, I'll do that when you're done.

buddde: Ok, thank you very much. I added some unittests, too.

src/UnitTests/ForEachEnumeratorDispose.cs
```diff
@@ -0,0 +1,343 @@
+﻿namespace AutoMapper.UnitTests
+{
+    using System;
+    using System.Collections;
+    using System.Collections.Generic;
+    using Shouldly;
+    using Xunit;
+
+    public class ForEachEnumeratorDispose : NonValidatingSpecBase
```
lbargaoanu: The same tests need to work for struct enumerators. Check the `List` source code for an example.

src/UnitTests/ForEachEnumeratorDispose.cs
```diff
@@ -0,0 +1,343 @@
...
+            };
+            Mapper.Map<SourceDest>(source).ShouldNotBeNull();
+
+            source.DisposableNumbers.IsDisposed.ShouldBe(true);
+        }
+
+        [Fact]
+        public void ForEach_for_disposable_type_should_call_Dispose_directly()
+        {
+            Configuration.CompileMappings();
```
lbargaoanu: Remove all these calls.

src/UnitTests/ForEachEnumeratorDispose.cs
lbargaoanu: Why NonValidatingSpecBase? The configuration should be valid.

lbargaoanu: Also there are no tests for the error case.

src/UnitTests/ForEachEnumeratorDispose.cs
```diff
@@ -0,0 +1,434 @@
...
+            GenericCollection.ShouldThrowException = false;
+            SourceDest source = new SourceDest { GenericNumbers = new GenericCollection() };
+            Mapper.Map<SourceDest>(source).ShouldNotBeNull();
+            source.GenericNumbers.IsDisposed.ShouldBe(true);
+        }
+
+        [Fact]
+        public void ForEach_for_known_disposable_struct_should_call_Dispose_despite_of_exception()
+        {
+            GenericCollection.UseStruct = true;
```
lbargaoanu: This won't work. Because the struct would be boxed. See [this](https://referencesource.microsoft.com/#mscorlib/system/collections/generic/list.cs,569). So you need a proper implementation for struct. But I guess only two tests for struct should do (one IDisposable and one not). I'm not sure we need more.

buddde: Why would the struct be boxed? Isn't there a direct call of the `Dispose` method?
Do you know any solution for it?

lbargaoanu: Follow that link. The public GetEnumerator returns the struct, that's the optimization I was talking about.

buddde: I'm sorry, but I still do not understand neither the problem nor the solution.

I can see that the `GetEnumerator` method returns a struct, but where does the boxing happen? Do you want me to check for the non-generic `IEnumerator<T>.GetEnumerator()` method in order to omit the runtime check?

lbargaoanu: The boxing happens in your GetEnumerator because you return an interface, not the concrete type. Again, check how List does it and do the same.

buddde: I hope, my PR fulfills your requirements.

lbargaoanu: At this point it felt cleaner to start from scratch, which I did in #3158.

PR #3160, opened 2019-07-08
WIP ForCtorParam MapFrom any expression

### LINKED ISSUE #3159: MapFrom doesn't resolve constructor parameters that have no matching source property
Re-opening issue #2745 with additional details.

### Source/destination types

```csharp
public class Source
{
    public string Result { get; }

    public Source(string result)
    {
        Result = result;
    }
}

public class Dest
{
    public string Result{ get; }
    public dynamic Details { get; }

    public Dest(string result, DestInner1 inner1)
    {
        Result = result;
        Details = inner1;
    }
    public Dest(string result, DestInner2 inner2)
    {
        Result = result;
        Details = inner2;
    }

    public class DestInner1
    {
        public int Value { get; }

        public DestInner1(int value)
        {
            Value = value;
        }
    }

    public class DestInner2
    {
        public int Value { get; }

        public DestInner2(int value)
        {
            Value = value;
        }
    }
}
```

### Mapping configuration

```csharp
protected override MapperConfiguration Configuration { get; } = new MapperConfiguration(config =>
{
    config.CreateMap<Source, Dest>()
        .ForCtorParam("inner2", cfg => cfg.MapFrom(_ => new Dest.DestInner2(100)));
});
```

### Expected behavior

Destination.Details should contain an instance of DestInner2.

### Actual behavior

AutoMapper.AutoMapperConfigurationException: 'Dest does not have a constructor with a parameter named 'inner2'.

### Steps to reproduce

```csharp
public void Should_redirect_value()
{
    var dest = Mapper.Map<Source, Dest>(new Source("Success"));

    dest.ShouldNotBeNull();
}
```

I have been looking into this issue, and following the comments in #2745, I added a dummy "inner2" parameter to the `Source` object, which causes `MapDestinationPropertyToSource` in `TypeMapFactory` to pass.

However, because `Dest` has two constructors, one accepting an instance of `inner1` and the other `inner2`, I would need to have both dummy parameters in `Source`.

Unfortunately, due to the way that constructor mapping is achieved, the first constructor always passes, so it always hits the "cannot find a constructor with parameter 'inner2'" error.

I note that the previous person to report this was asked to submit a PR last August and hasn't done so, so I have looked into it myself.  I have a working prototype, but I don't know the codebase anywhere near well enough to know if I'm missing things.

All existing tests still pass, as does the one I have added to Constructor.cs.

Any assistance would be appreciated.

spydacarnage:
- Added code to TypeMapFactory+ConstructorMap to check for valid ForCtorParam entries to match against a Constructor

Fixes #3159.

### DIFF
--- a/src/AutoMapper/ApiCompatBaseline.txt
+++ b/src/AutoMapper/ApiCompatBaseline.txt
@@ -1,6 +1,7 @@
 Compat issues with assembly AutoMapper:
 MembersMustExist : Member 'AutoMapper.AutoMapperConfigurationException.PropertyMap.get()' does not exist in the implementation but it does exist in the contract.
 MembersMustExist : Member 'AutoMapper.AutoMapperConfigurationException.PropertyMap.set(AutoMapper.PropertyMap)' does not exist in the implementation but it does exist in the contract.
+MembersMustExist : Member 'AutoMapper.ConstructorMap..ctor(System.Reflection.ConstructorInfo, AutoMapper.TypeMap)' does not exist in the implementation but it does exist in the contract.
 TypesMustExist : Type 'AutoMapper.CtorParamConfigurationExpression<TSource>' does not exist in the implementation but it does exist in the contract.
 InterfacesShouldHaveSameMembers : Interface member 'AutoMapper.IConfigurationProvider.Features' is present in the implementation but not in the contract.
 InterfacesShouldHaveSameMembers : Interface member 'AutoMapper.IConfigurationProvider.Features.get()' is present in the implementation but not in the contract.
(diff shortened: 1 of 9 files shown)


lbargaoanu:
I'm not sure this is the right approach. It seems better to try and move the MapDestinationCtorToSource code in MappingExpressionBase.Configure where the configuration is available.

spydacarnage:
Thanks for the feedback.

Unfortunately, my first thought was to look at the Configure method, but
the Configure tree doesn't get called until after the TypeMapFactory has
chosen the (sometimes incorrect) constructor to map to.

The ForCtorParam mappings have to be checked during that phase somehow, and
running the whole Configure tree early also doesn't work because that needs
to do things later in the process.

So, while I agree that this isn't the perfect approach, I can't see another
obvious route.

lbargaoanu:
What I'm saying is that the code that chooses the constructor should be in Configure.

spydacarnage:
Okay, I'll have a look at how that sits today and get back to you.

lbargaoanu:
Closing as stale. But feel free to open a new one when it's ready.

spydacarnage:
@lbargaoanu - the update I put up 2 weeks ago was that "I believe it is ready" state - I thought you adding the milestone was an implicit acceptance.

lbargaoanu:
This PR is not even ready for review.

spydacarnage:
Not sure what happened - it went through all the checks and passed...

Have created #3169 as a second attempt.

PR #3162, opened 2019-07-10
Inherit included members

lbargaoanu:


src/AutoMapper/PropertyMap.cs
```diff
@@ -93,6 +93,7 @@ public void ApplyInheritedPropertyMap(PropertyMap inheritedMappedProperty)
             ValueResolverConfig = ValueResolverConfig ?? inheritedMappedProperty.ValueResolverConfig;
             ValueConverterConfig = ValueConverterConfig ?? inheritedMappedProperty.ValueConverterConfig;
             _valueTransformerConfigs.InsertRange(0, inheritedMappedProperty._valueTransformerConfigs);
+            _memberChain = _memberChain.Count == 0 ? inheritedMappedProperty._memberChain : _memberChain;
```
jbogard: This seems weird to replace the reference for a collection. I know it's what we're doing with the others but it still seems odd.

lbargaoanu: Yes, but it does seem like the right thing to do is to replace it (ok, other than clear and insert). I guess we'll see what happens when people use it.

PR #3169, opened 2019-07-22
ForCtorParam MapFrom any expression

### LINKED ISSUE #3159: MapFrom doesn't resolve constructor parameters that have no matching source property

spydacarnage:
Fixes #3159.

### DIFF
--- a/src/AutoMapper/Configuration/CtorParamConfigurationExpression.cs
+++ b/src/AutoMapper/Configuration/CtorParamConfigurationExpression.cs
@@ -7,10 +7,10 @@ namespace AutoMapper.Configuration
 {
     public class CtorParamConfigurationExpression<TSource> : ICtorParamConfigurationExpression<TSource>, ICtorParameterConfiguration
     {
-        private readonly string _ctorParamName;
+        public string CtorParamName { get; }
         private readonly List<Action<ConstructorParameterMap>> _ctorParamActions = new List<Action<ConstructorParameterMap>>();
 
-        public CtorParamConfigurationExpression(string ctorParamName) => _ctorParamName = ctorParamName;
+        public CtorParamConfigurationExpression(string ctorParamName) => CtorParamName = ctorParamName;
 
         public void MapFrom<TMember>(Expression<Func<TSource, TMember>> sourceMember)
         {
@@ -31,10 +31,10 @@ namespace AutoMapper.Configuration
                 throw new AutoMapperConfigurationException($"The type {typeMap.Types.DestinationType.Name} does not have a constructor.\n{typeMap.Types.DestinationType.FullName}");
             }
 
-            var parameter = ctorParams.SingleOrDefault(p => p.Parameter.Name == _ctorParamName);
+            var parameter = ctorParams.SingleOrDefault(p => p.Parameter.Name == CtorParamName);
             if (parameter == null)
             {
-                throw new AutoMapperConfigurationException($"{typeMap.Types.DestinationType.Name} does not have a constructor with a parameter named '{_ctorParamName}'.\n{typeMap.Types.DestinationType.FullName}");
+                throw new AutoMapperConfigurationException($"{typeMap.Types.DestinationType.Name} does not have a constructor with a parameter named '{CtorParamName}'.\n{typeMap.Types.DestinationType.FullName}");
             }
             parameter.CanResolveValue = true;
 
--- a/src/AutoMapper/Configuration/MappingExpressionBase.cs
+++ b/src/AutoMapper/Configuration/MappingExpressionBase.cs
@@ -58,7 +58,20 @@ namespace AutoMapper.Configuration
                 }
             }
 
-            foreach(var action in TypeMapActions)
+            var destTypeInfo = typeMap.DestinationTypeDetails;
+            if(!typeMap.DestinationType.IsAbstract())
+            {
+                foreach(var destCtor in destTypeInfo.Constructors.OrderByDescending(ci => ci.GetParameters().Length))
+                {
+                    if(typeMap.Profile.MapDestinationCtorToSource(typeMap, destCtor, typeMap.SourceTypeDetails, CtorParamConfigurations))
+                    {
+                        break;
+                    }
+                }
+            }
+
+
+            foreach (var action in TypeMapActions)
             {
                 action(typeMap);
             }
--- a/src/AutoMapper/ProfileMap.cs
+++ b/src/AutoMapper/ProfileMap.cs
@@ -12,7 +12,6 @@ namespace AutoMapper
     [DebuggerDisplay("{Name}")]
     public class ProfileMap
     {
-        private readonly TypeMapFactory _typeMapFactory = new TypeMapFactory();
         private readonly IEnumerable<ITypeMapConfiguration> _typeMapConfigs;
         private readonly IEnumerable<ITypeMapConfiguration> _openTypeMapConfigs;
         private readonly LockingConcurrentDictionary<Type, TypeDetails> _typeDetails;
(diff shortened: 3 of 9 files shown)


src/AutoMapper/ConstructorMap.cs
```diff
@@ -4,6 +4,7 @@
 using System.Linq.Expressions;
 using System.Reflection;
 using AutoMapper.Execution;
+using AutoMapper.Configuration;
```
lbargaoanu: Please revert changes to this file.

src/AutoMapper/ProfileMap.cs
```diff
@@ -32,8 +32,8 @@ public ProfileMap(IProfileConfiguration profile, IConfiguration configuration)
             EnableNullPropagationForQueryMapping = profile.EnableNullPropagationForQueryMapping ?? configuration?.EnableNullPropagationForQueryMapping ?? false;
```
lbargaoanu: Please revert changes to this file.

src/AutoMapper/TypeMapFactory.cs
```diff
@@ -24,52 +24,52 @@ public TypeMap CreateTypeMap(Type sourceType, Type destinationType, ProfileMap o
...
-            {
-                foreach (var destCtor in destTypeInfo.Constructors.OrderByDescending(ci => ci.GetParameters().Length))
-                {
-                    if (MapDestinationCtorToSource(typeMap, destCtor, sourceTypeInfo, options))
-                    {
-                        break;
-                    }
-                }
-            }
+            //if (!destinationType.IsAbstract())
```
lbargaoanu: Remove this.

src/AutoMapper/TypeMapFactory.cs
```diff
@@ -24,52 +24,52 @@ public TypeMap CreateTypeMap(Type sourceType, Type destinationType, ProfileMap o
...
             {
                 return false;
             }
             return options.MemberConfigurations.Any(_ => _.MapDestinationPropertyToSource(options, sourceTypeInfo, destType, destMemberType, destMemberInfo, members));
         }
 
-        private bool MapDestinationCtorToSource(TypeMap typeMap, ConstructorInfo destCtor, TypeDetails sourceTypeInfo, ProfileMap options)
-        {
-            var ctorParameters = destCtor.GetParameters();
+        //private bool MapDestinationCtorToSource(TypeMap typeMap, ConstructorInfo destCtor, TypeDetails sourceTypeInfo, ProfileMap options)
```
lbargaoanu: Remove this.

src/AutoMapper/ConstructorMap.cs
spydacarnage: Updates made

src/AutoMapper/ProfileMap.cs
spydacarnage: Updates made

src/AutoMapper/TypeMapFactory.cs
spydacarnage: Updates made

src/AutoMapper/TypeMapFactory.cs
spydacarnage: Updates made

src/UnitTests/General.cs
```diff
@@ -3,11 +3,41 @@
 using Shouldly;
```
lbargaoanu: Revert changes to this file.

spydacarnage: Apologies - that was a hangover from testing another reported issue simultaneously.  Removed now.

src/AutoMapper/Configuration/MappingExpressionBase.cs
```diff
@@ -100,6 +113,41 @@ public void Configure(TypeMap typeMap)
...
+                }
+                ctorMap.AddParameter(parameter, resolvers.ToArray(), canResolve);
+            }
+
+            typeMap.ConstructorMap = ctorMap;
+
+            return ctorMap.CanResolve;
+        }
+
+        private bool MapDestinationPropertyToSource(ProfileMap options, TypeDetails sourceTypeInfo, Type destType, Type destMemberType, string destMemberInfo, LinkedList<MemberInfo> members)
```
lbargaoanu: That seems just copy pasted from TypeMapFactory.

spydacarnage: It is, as it is needed in both locations.  I wasn't sure what your policy on code sharing is.

I could remove it from both locations and add it as an instance method of `ProfileMap`, as that is shared between both locations?  (I would also move `MapDestinationCtorToSource` to `ProfileMap`, too, as they are obviously linked).

Thoughts?

lbargaoanu: OK, move them both.

src/AutoMapper/Configuration/CtorParamConfigurationExpression.cs
```diff
@@ -43,5 +43,10 @@ public void Configure(TypeMap typeMap)
                 action(parameter);
             }
         }
+
+        public bool CheckCtorParamName(string paramName)
```
lbargaoanu: This just hides the meaning. Just expose the parameter name.

src/AutoMapper/Configuration/MappingExpressionBase.cs
spydacarnage: Moved

src/AutoMapper/Configuration/CtorParamConfigurationExpression.cs
spydacarnage: Update made

src/UnitTests/Constructors.cs
```diff
@@ -1609,4 +1609,133 @@ public void Should_redirect_value()
...
+            config.CreateMap<Source, Dest>()
+                .ForCtorParam("inner1", cfg => cfg.MapFrom(_ => new Dest.DestInner1(100)));
+        });
+
+        [Fact]
+        public void Should_redirect_value()
+        {
+            var dest = Mapper.Map<Dest>(new Source("Success"));
+
+            dest.ShouldNotBeNull();
```
lbargaoanu: Assert here what the property values should be. The same for the other test.

spydacarnage: Done

PR #3180, opened 2019-08-06
Update Custom-value-resolvers.md

lbargaoanu:


docs/Dependency-injection.md
```diff
@@ -29,22 +29,20 @@ Note that IQueryable.ProjectTo is [more limited](Queryable-Extensions.html#suppo
 
 ### ASP.NET Core
 
-There is a [NuGet package](https://www.nuget.org/packages/AutoMapper.Extensions.Microsoft.DependencyInjection/) to be used with the default injection mechanism described [here](https://github.com/AutoMapper/AutoMapper.Extensions.Microsoft.DependencyInjection/blob/master/README.md) and used in [this project](https://github.com/jbogard/ContosoUniversityCore/blob/master/src/ContosoUniversityCore/Startup.cs).
+There is a [NuGet package](https://www.nuget.org/packages/AutoMapper.Extensions.Microsoft.DependencyInjection/) to be used with the default injection mechanism described [here](https://github.com/AutoMapper/AutoMapper.Extensions.Microsoft.DependencyInjection) and used in [this project](https://github.com/jbogard/ContosoUniversityCore/blob/master/src/ContosoUniversityCore/Startup.cs).
 
-Once the nuget package is downloaded, simply add AutoMapper to your IServiceCollection in your startup.cs class:
+You define the configuration using [profiles](Configuration.html#profile-instances). And then you let AutoMapper know where those profiles are defined by calling the extension method `AddAutoMapper` (for `IServiceCollection`) in your startup.cs class:
```
lbargaoanu: @Tasteful Better? Smth essential missing?

Tasteful: I think we should add some words here about the configure parameter, example

---

Global configuration for AutoMapper can be configured with the `Action<IMapperConfigurationExpression>` parameter that exists in the `AddAutoMapper` extension method.
```c#
services.AddAutoMapper(cfg => 
{ 
    cfg.SourceMemberNamingConvention = new LowerUnderscoreNamingConvention();
}, 
profileAssembly1, profileAssembly2 /*, ...*/);
```
or marker types:
```c#
services.AddAutoMapper(cfg => 
{ 
    cfg.SourceMemberNamingConvention = new LowerUnderscoreNamingConvention();
}, 
profileType1, profileType2 /*, ...*/);
```

lbargaoanu: Maybe, but that doesn't belong here (it belongs in the DI repo). That's not usually needed. This is just on overview. How you start things.

Tasteful: Even that the code is for the DI package the documentation is all generated from this repo and in the other we only have the readme file and I think most of the developers is starting with documentation from https://docs.automapper.org and when this area is taking up how to use the DI I think the developers is not going into that repo and read the readme there.

lbargaoanu: Some people are going to struggle no matter what. What I really care about is the getting started docs. What I think a regular Joe developer needs. How do I use AM with .Net Core DI? You expect a to the point answer and this is it. If you cannot be bothered to follow the link to get more information, well, good luck :)

PR #3183, opened 2019-08-08
Swapped the ASP.NET Core for a Tutorial

RotterAxel:
Added an easy to follow and understand Tutorial on how to incoroporate Automapper in the ASP.NET Core Dependency Injection System. 
I tried to follow the text that was previously there and did not find it helpfull at all.

### DIFF
--- a/docs/Dependency-injection.md
+++ b/docs/Dependency-injection.md
@@ -29,33 +29,72 @@ Note that IQueryable.ProjectTo is [more limited](Queryable-Extensions.html#suppo
 
 ### ASP.NET Core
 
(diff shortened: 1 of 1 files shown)


lbargaoanu:
In fact the existing documentation describes the proper way to use AM in ASP.NET Core. See also https://github.com/AutoMapper/AutoMapper.Extensions.Microsoft.DependencyInjection/blob/master/README.md.

lbargaoanu:
So what _exactly_ is not clear in the existing documentation?

RotterAxel:
In my Opinion with the current documentation steps 3, 4 and 5 are great resources to see an example implmementation. They might be added to the current Text as examples. 

I don´t know what you mean by "average developer" so I´m just going to ignore the remark, since I´m trying to help you make your documentation better.  Please refrain from such commentaries and keep your professionalism.

PR #3187, opened 2019-08-09
Match destination enumerable types with it's enumerable for LINQ

### LINKED ISSUE #3186: Support IEnumerable<T> for LINQ child member projections
Consider our types:

```c#
public class ParentDto {
    public IEnumerable<ChildDto> Children { get; set; }
}

public class Parent {
    public IEnumerable<Child> Children { get; set; }
}
```

When we call `ProjectTo`, our LINQ query is:

```c#
var parent = querayble.Select(p => new {
    Children = p.Children.Select(c => /* projection */).ToList()
});
```

However, the destination type is only `IEnumerable`, not a list. We should only call `ToList` if the destination type is a list. Right now we do `ToArray` if it's an array, then default back to `ToList`. We should match the destination type - only do `ToList` if the destination type is `IList`.

I ran into this on a projection from a MongoDB driver, which does not support `ToList` because that implies running the list at the server.

jbogard:
Fixes #3186

### DIFF
--- a/src/AutoMapper/QueryableExtensions/Impl/EnumerableExpressionBinder.cs
+++ b/src/AutoMapper/QueryableExtensions/Impl/EnumerableExpressionBinder.cs
@@ -2,7 +2,6 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Linq.Expressions;
 using AutoMapper.Configuration;
-using AutoMapper.Mappers;
 using AutoMapper.Mappers.Internal;
 
 namespace AutoMapper.QueryableExtensions.Impl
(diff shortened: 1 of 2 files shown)


src/AutoMapper/QueryableExtensions/Impl/EnumerableExpressionBinder.cs
```diff
@@ -29,17 +28,24 @@ private static MemberAssignment BindEnumerableExpression(IConfigurationProvider
...
                     return null;
                 }
-                expression = transformedExpressions.Aggregate(expression, (source, lambda) => Select(source, lambda));
+                expression = transformedExpressions.Aggregate(expression, Select);
             }
 
-            expression = Expression.Call(typeof(Enumerable), propertyMap.DestinationType.IsArray ? "ToArray" : "ToList", new[] { destinationListType }, expression);
+            expression =
+                propertyMap.DestinationType.IsArray
+                    ? Expression.Call(typeof(Enumerable), nameof(Enumerable.ToArray), new[] {destinationListType}, expression)
```
lbargaoanu: Some replication here. Only the method name is different. And it's the same for IList and ICollection.

jbogard: Changed to just do `ICollection<T>` since that covers both cases

PR #3211, opened 2019-09-03
Create no-response.yml

lbargaoanu:
https://github.com/probot/no-response

### DIFF
--- a/.github/no-response.yml
+++ b/.github/no-response.yml
@@ -1,0 +1,8 @@
+
+# Number of days of inactivity before an Issue is closed for lack of response
+daysUntilClose: 7
+# Label requiring a response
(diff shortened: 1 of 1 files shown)


.github/no-response.yml
```diff
@@ -0,0 +1,8 @@
+
+# Number of days of inactivity before an Issue is closed for lack of response
+daysUntilClose: 7
```
jbogard: Maybe make it like a month? A week isn't too long for someone to fix it themselves

lbargaoanu: I'm only closing issues so I don't have to look at them. A month is practically forever :) I can simply leave them open. I might forget about some issues this way, but I can live with that.

Issue #3244, opened 2019-10-18
Enum value mapping and assertion


HenkKin:
A few years ago I implemented an extension to AutoMapper to handle enum value mapping. Especially having control about the enum values by AssertConfigurationIsValid would be nice. 

I created a github with a working example: https://github.com/HenkKin/Enums.Extensions.AutoMapper

### Source/destination types

```csharp
public enum Source
{
    Default = 0,
    First = 1,
    Second = 2
}

public enum Destination
{
    Default = 0,
    Second = 2
}
```

### Mapping configuration

```csharp
CreateMap<Source, Destination>()
     .AsEnumMap(EnumMappingType.Value) // OR EnumMappingType.Name
     .MapFromEnumValue(Source.First, Destination.Default); 
```

### Testing

```csharp
[Fact]
public void WhenProfilesAreConfigured_ItShouldNotThrowException()
{
    // Arrange
    var config = new MapperConfiguration(configuration =>
    {
        configuration.AddMaps(typeof(AssemblyInfo).GetTypeInfo().Assembly);
    });

    // default automapper assertions
    config.AssertConfigurationIsValid();

    // Assert
    config.AssertEnumConfigurationIsValid(); // this line asserts all enum value mapping configs
}
```
### Feature request
It would be nice when the mapping functionality will become default functionality of AutoMapper. 
- CreateMap -> should support enum mapping directly without calling AsEnumMap
- AssertConfigurationIsValid -> should include validating enum mapping too.

lbargaoanu:
You can add custom validations through an extension point. See [here](https://github.com/AutoMapper/AutoMapper/blob/7a00700de61cd3234b6a6eb4bf0f6bbe402369b1/src/UnitTests/ConfigurationValidation.cs#L30).

HenkKin:
> You can add custom validations through an extension point. See here.

This is not working for TypeMaps with ConvertUsing (ITypeConverter). These TypeMaps will bypas this validation extions point. I've  already tried it.

lbargaoanu:
Well, you write the code there, so you can do whatever. Other than that, I guess it could. A PR is welcome.

PR #3247, opened 2019-10-22
Added Enum value mapping

HenkKin:
Added enum value mapping configuration to AutoMapper:

```csharp
public enum Source
{
    Default = 0,
    First = 1,
    Second = 2
}

public enum Destination
{
    Default = 0,
    Second = 2
}

internal class YourProfile : Profile
{
    public YourProfile()
    {
	CreateMap<Source, Destination>()
                .ConvertAsEnum(opt => opt
                    .MapByValue() // or .MapByName()
                    .MapValue(Source.First, Destination.Second));
    }
}
```


HenkKin:
Would anyone want to do the review?

lbargaoanu:
Not me :) I don't think this belongs in AM. Perhaps I've mislead you and I'm sorry if that's the case. [Here](https://github.com/AutoMapper/AutoMapper/issues/3244#issuecomment-543715006) I was only talking about the validation extension point.

HenkKin:
> Not me :) I don't think this belongs in AM. Perhaps I've mislead you and I'm sorry if that's the case. Here I was only talking about the validation extension point.

Normally enum types should not be registered in AutoMapper, because automapper will map it as an integer. If you want to map some enum values, you have to create a custom TypeConverter. But TypeConverters are not tested in AssertConfigurationIsValid, so if you want to control your enum mapping logic, you should create some unittests. When you have a lot of enums, it is unnecessarily repeating work to create a lot of TypeConverters and associated unittests.

I think this enum mapping feature should belong to AutoMapper. With AM we are in control about our mappings. It is testable by AssertConfigurationIsValid. The validation extension point is not working for TypeConverters. Also the logic within the TypeConverter is not tested by AssertConfigurationIsValid. With this pull request, the enum types are (optionally) testable too. When an enum value is added to one enum type, the AssertConfigurationIsValid thows an exception when the value cannot mapped to another enum type with the missing enum value. With this change, as a developer, you have the option to be in control about your enum mappings, without creating a TypeConverter for every enum type pair and associated unittests. Also this change is not compelling, but it is an option to the developer.

Tasteful:
I have done some tests with your code to make the implementation based on the `Feature` instead and it's only the validation that I don't get to run automatic and the reason is that when we cant force a validation to execute if we use a custom map function/expression method.

From `TypeMap`
```c#
        public bool ShouldCheckForValid =>
            CustomMapFunction == null
            && CustomMapExpression == null
            && TypeConverterType == null
            && DestinationTypeOverride == null
            && ConfiguredMemberList != MemberList.None
            && !(IsValid ?? false);
```
Maybe it can be added a flag that can be turned on to force the validation to execute on the `TypeMap` to be able to add custom validations also.

If it's ok to call an extension method in your project `Configuration.AssertConfigurationIsValid(true)` instead of default `Configuration.AssertConfigurationIsValid()`  you can if you want to use the attached POC project that don't need any changes inside auto mapper.

[AutoMapper.Extensions.EnumMapping.zip](https://github.com/AutoMapper/AutoMapper/files/3881161/AutoMapper.Extensions.EnumMapping.zip)

HenkKin:
> I have done some tests with your code to make the implementation based on the Feature instead and it's only the validation that I don't get to run automatic and the reason is that when we cant force a validation to execute if we use a custom map function/expression method.
> From TypeMap
>         public bool ShouldCheckForValid =>
>             CustomMapFunction == null
>             && CustomMapExpression == null
>             && TypeConverterType == null
>             && DestinationTypeOverride == null
>             && ConfiguredMemberList != MemberList.None
>             && !(IsValid ?? false);
> Maybe it can be added a flag that can be turned on to force the validation to execute on the TypeMap to be able to add custom validations also.
> If it's ok to call an extension method in your project Configuration.AssertConfigurationIsValid(true) instead of default Configuration.AssertConfigurationIsValid()  you can if you want to use the attached POC project that don't need any changes inside auto mapper.
> AutoMapper.Extensions.EnumMapping.zip

Earlier I created an extension (https://github.com/HenkKin/Enums.Extensions.AutoMapper) to AutoMapper based on extensions methods and an AssertEnumConfigurationIsValid. It stays completely outside AutoMapper. I like your implementation based on feature. Thats a better choice of integration than only extensions methods. But it has the same problem with AssertConfigurationIsValid. I believe that this pull request should be core functionality of AutoMapper.

Tasteful:
@lbargaoanu @jbogard What do you think of adding custom validation that is executed when the `AssertConfigurationIsValid` is invoked?

lbargaoanu:
That already [exists](http://docs.automapper.org/en/latest/Configuration-validation.html#custom-validations), just it's not called for user maps. It's easy to change.

Tasteful:
@lbargaoanu the `TypeMap.ShouldCheckForValid` is executed for all types that returns true, so in this case the custom validator that is added into `cfg.Advanced.Validator` is not execute.

The `TypeMap.ShouldCheckForValid` looks like the following and is not executed if we have a custom function, custom map express, type converter etc that make it not cover all cases. To be able to make it good the validation in the extension the validation method also need to be executed for these cases.
```
        public bool ShouldCheckForValid =>
            CustomMapFunction == null
            && CustomMapExpression == null
            && TypeConverterType == null
            && DestinationTypeOverride == null
            && ConfiguredMemberList != MemberList.None
            && !(IsValid ?? false);
```

lbargaoanu:
https://github.com/AutoMapper/AutoMapper/blob/df1b5fe13d64f11885746291987aca89fe808ee0/src/AutoMapper/ConfigurationValidator.cs#L83

I would guess al that's needed is to move this line above, before the `if`.

HenkKin:
I

> AutoMapper/src/AutoMapper/ConfigurationValidator.cs
>     
>     
>          Line 83
>       in
>       df1b5fe
>     
>     
>     
>     
> 
>         
>           
>            _config.Validate(context); 
>         
>     
>   
> 
> 
> I would guess all that's needed is to move this line above, before the if.

I created a pull request for this change (https://github.com/AutoMapper/AutoMapper/pull/3274).

@lbargaoanu @jbogard Is it possible to add a new extension to AutoMapper for enum-mapping, something like AutoMapper.Extensions.EnumMapping? 

I have a working changeset based on this pull request and the zip file of @Tasteful which is working well with AssertConfigurationIsValid.

HenkKin:
I created the extension library: https://github.com/HenkKin/AutoMapper.Extensions.EnumMapping
@jbogard how can I create AutoMapper.Extensions.EnumMapping library? Can you provide the github library and the nuget name?

At the moment this library is pointing to myget source AutoMapper 9.1.0-ci-*, because it is waiting for version 9.1.0.

@Tasteful: thanks for pointing me in the right direction of using Features. Is there any documentation about using Features of AutoMapper?

Tasteful:
I think documentation is missing for the _Feature_-feature. I created the initial spec (https://github.com/AutoMapper/AutoMapper/issues/2991) to work with AutoMapper.Collection where we have problem with running test in parallell when the settings was stored in static concurrent dictionary.

I think a PR for an docs-article about this topic is welcome to describe for extensions developer how they can integrate own settings into the mapping and configuration.

PR #3250, opened 2019-10-25
Obsolete the context mapper overloads that receive options

lbargaoanu:


Directory.Build.props
```diff
@@ -1,7 +1,7 @@
 <Project>
 
     <PropertyGroup>
-        <VersionPrefix>9.0.1</VersionPrefix>
+        <VersionPrefix>9.1.1</VersionPrefix>
```
jbogard: 9.1.0?

PR #3268, opened 2019-11-20
Honour AllowNullDestinationValues

### LINKED ISSUE #3267: Mapper.Map always returns null when source object is null
### Source/destination types

```csharp
public class DestModel
{
    public int Length { get; set; }
}

// Source is string
```

### Mapping configuration

```csharp
public class DestModelConverter : ITypeConverter<string, DestModel>
{
    public DestModel Convert(string source, DestModel destination, ResolutionContext)
    {
        return new DestModel
        {
            Length = source == null ? 0 : source.Length, // OR source?.Length ?? 0
            Value = source
        };
    }
}

var config = new MapperConfiguration(c => 
{
    c.AllowNullDestinationValues = false;
    c.CreateMap<string, DestModel>()
        .ConvertUsing<DestModelConverter>();
});
    
```

### Version: x.y.z
Version: 9.0.0

### Expected behavior

I expected the custom converter's `Convert` method to get fired when I call `mapper.Map` and the destination object to not be null.

### Actual behavior

Type Converter's Convert never gets called, and the destination object (returned from mapper.Map) is always null.

### Steps to reproduce

Here is a gist to reproduce: https://gist.github.com/welrocken/badeb4f2c90326accd828b3b257815a6
```csharp
{
    var config = new MapperConfiguration(c => 
    {
        c.AllowNullDestinationValues = false;
        c.CreateMap<string, DestModel>()
            .ConvertUsing<DestModelConverter>();
    });
    
    string source = null;
    
    try
    {
        var dest = mapper.Map<DestModel>(source);
        
        Debug.Assert(dest != null);
    }
    catch (Exception ex)
    {
        ex.ToString();
    }
}

public class DestModel
{
    public int Length { get; set; }
}

public class DestModelConverter : ITypeConverter<string, DestModel>
{
    public DestModel Convert(string source, DestModel destination, ResolutionContext)
    {
        return new DestModel
        {
            Length = source == null ? 0 : source.Length, // OR source?.Length ?? 0
            Value = source
        };
    }
}
```

### Extra Notes

This also happens to be the case with .ForMember and .ConstructUsing. It looks like no matter what the configuration is, when source is null (the root object), destination becomes null.

lbargaoanu:
I guess this will break some existing code. Fixes #3267.

### DIFF
--- a/src/AutoMapper/Configuration/PrimitiveExtensions.cs
+++ b/src/AutoMapper/Configuration/PrimitiveExtensions.cs
@@ -15,6 +15,8 @@ namespace AutoMapper.Configuration
             }
         }
 
+        public static bool IsNonStringEnumerable(this Type type) => type != typeof(string) && type.IsEnumerableType();
+
         public static bool IsSetType(this Type type)
             => type.ImplementsGenericInterface(typeof(ISet<>));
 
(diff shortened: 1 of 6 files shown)


src/AutoMapper/Configuration/PrimitiveExtensions.cs
```diff
@@ -15,6 +15,8 @@ public static void ForAll<T>(this IEnumerable<T> enumerable, Action<T> action)
             }
         }
 
+        public static bool IsNonStringEnumerable(this Type type) => type != typeof(string) && type.IsEnumerableType();
```
lbargaoanu: This is silly, but I don't have a better name.

PR #3270, opened 2019-11-22
Update index.rst

lbargaoanu:
I've messed this up somehow. But it's not exactly WYSIWYG.


lbargaoanu:
@jbogard Can we merge this? Because http://docs.automapper.org/en/latest/ is broken right now.

lbargaoanu:
I guess I still didn't get the syntax right.

PR #3274, opened 2019-11-29
Enabled custom validation for TypeConverters via Advanced.Validator

HenkKin:
If you use TypeConverters, this pull request allows you to add custom validators via Advanced.Validator

### DIFF
--- a/src/UnitTests/AdvancedConfiguration.cs
+++ b/src/UnitTests/AdvancedConfiguration.cs
@@ -1,0 +1,160 @@
+using System;
+using System.Linq.Expressions;
+using Shouldly;
+using Xunit;
+
+namespace AutoMapper.UnitTests
+{
+    public class AdvancedConfiguration
+    {
+        public class Source
+        {
+        }
+
+        public class Destination
+        {
+        }
+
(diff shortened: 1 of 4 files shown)


src/UnitTests/AdvancedConfiguration.cs
```diff
@@ -0,0 +1,160 @@
+using System;
```
lbargaoanu: Can you please rename this file/class to CustomValidations? And move the existing test here.

HenkKin: I renamed it

lbargaoanu: https://github.com/AutoMapper/AutoMapper/blob/7a00700de61cd3234b6a6eb4bf0f6bbe402369b1/src/UnitTests/ConfigurationValidation.cs#L9 belongs in this file too.

HenkKin: This test is moved to the new CustomValidations file. I have searched for all usages of cfg.Advanced.Validator in test files, There is no one left.

Issue #2782, opened 2018-09-05
Feature request: Validate mapping of read-only collection-typed properties


antoinebj:
Hi,

The project I'm working on uses read-only properties of type IList<> for collections both in source and destination types.
By default, AutoMapper ignores those properties because of their lack of a public setter. We therefore use the explicit MapFrom and UseDestinationValue to force the mapping to happen.
It would be nice if we didn't have to do that, but it works, so I can live with it.

However, if one of those collection properties is not mapped, the AssertConfigurationIsValid() method does **not** throw an exception, also because the lack of a public setter. And for that, there does not seem to be any workaround.

After looking at the source code, it appears that at one point there must have been some intent to consider read-only list/collection properties the same way as regular read/write properties.
I'm referring to this line of code in TypeDetails.BuildPublicAccessors:
```csharp
.Where(pi => pi.CanWrite || pi.PropertyType.IsListOrDictionaryType())
```

IsListOrDictionaryType() calls IsListType():
```csharp
public static bool IsListType(Type type)
    => typeof(IList).IsAssignableFrom(type);
```

Now, I don't know if that's intentional, but if it also checked for ICollection<> and not just the non-generic IList (BTW why not ICollection?), it appears that it would fix having to use MapFrom + UseDestinationValue, and also make the configuration validator take these kinds of properties into consideration.

Are there any reasons not to go in that direction?

lbargaoanu:
See #2743. It's just a hack. If you want to "improve" it, a PR is welcome. About validation, I guess it's not anything a few extra checks won't solve. But it might break people. The workaround would  be to use [the extension point](https://github.com/AutoMapper/AutoMapper/blob/dfce8185d2ed468b55e8a96708f0fd3aed2e7d88/src/UnitTests/ConfigurationValidation.cs#L30).

PR #3275, opened 2019-12-01
Add automatic mapping of writable collection members that have no write accessor

antoinebj:
This is a follow-up to issue #2782 that I had raised more than a year ago.
It also addresses the parts of #3036 which weren't covered by PR #2928.
See https://github.com/AutoMapper/AutoMapper/pull/3037#issuecomment-485367650

I created a feature flag on the mapping configuration to toggle this behavior.
By default, the feature is off and the behavior of AutoMapper remains the same.
This should allow releasing it without breaking existing code that is using the library (so, it should be fit for a 9.1).

When it is switched on, such properties will be auto-mapped:
```csharp
IList<int> Values { get; } = new List<int>();
```
Notice that `IList<>` is now supported, not just `List<>`.

I've also taken the liberty to simplify some logic in the `TypeDetails` class.

Take the following Where clause in its `BuildPublicAccessors` method:
```csharp
.Where(pi => pi.CanWrite || pi.PropertyType.IsListOrDictionaryType())
```

It seems to be redundant if its logic is put in the `PropertyWritable` method. I could not find a reason why the latter was filtering on `IEnumerable`, going so far as specifically excluding `string`, only to then filter further in `BuildPublicAccessors` on more specialized list and dictionary types.

When the feature is turned on, instead of only list and dictionary types, it takes all types implementing `ICollection<>` and `ICollection`, which covers the non-generic `IList` as well as writable dictionary types.

All unit tests pass, and I added 2 more to cover when the feature is on or off.

Feel free to suggest or make modifications to better suit the quality expectations of the project.

### DIFF
--- a/src/AutoMapper/Configuration/IProfileConfiguration.cs
+++ b/src/AutoMapper/Configuration/IProfileConfiguration.cs
@@ -2,7 +2,6 @@ using System;
 using System.Collections.Generic;
 using System.Reflection;
 using AutoMapper.Configuration.Conventions;
-using AutoMapper.Mappers;
 
 namespace AutoMapper.Configuration
 {
@@ -16,6 +15,7 @@ namespace AutoMapper.Configuration
         bool? AllowNullDestinationValues { get; }
         bool? AllowNullCollections { get; }
         bool? EnableNullPropagationForQueryMapping { get; }
+        bool? EnableMappingOfCollectionMembersWithoutWriteAccessor { get; }
         IEnumerable<Action<TypeMap, IMappingExpression>> AllTypeMapActions { get; }
         IEnumerable<Action<PropertyMap, IMemberConfigurationExpression>> AllPropertyMapActions { get; }
 
--- a/src/AutoMapper/Configuration/Internal/PrimitiveHelper.cs
+++ b/src/AutoMapper/Configuration/Internal/PrimitiveHelper.cs
@@ -31,8 +31,11 @@ namespace AutoMapper.Configuration.Internal
         public static Type GetTypeOfNullable(Type type) 
             => type.GetTypeInfo().GenericTypeArguments[0];
 
-        public static bool IsCollectionType(Type type) 
+        public static bool IsGenericCollectionType(Type type) 
             => type.ImplementsGenericInterface(typeof(ICollection<>));
+
+        public static bool IsCollectionType(Type type)
+            => typeof(ICollection).IsAssignableFrom(type) || IsGenericCollectionType(type);
 
         public static bool IsEnumerableType(Type type) 
             => typeof(IEnumerable).IsAssignableFrom(type);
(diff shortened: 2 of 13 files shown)


lbargaoanu:
I, for one, would rather see this outside AM. Just another special case.

antoinebj:
> I, for one, would rather see this outside AM. Just another special case.

I could be mistaken, but my belief is that it's a well-known practice to have collection-typed properties be read-only to avoid dealing with null checks and having to react to instance changes when logic requires reacting to collection changes. Some [static code analysis tools have rules to encourage it](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca2227), but AutoMapper seems to discourage it.

Moreover, I'm confused by the presence of this code in the `TypeDetails.BuildPublicAccessors()` method:
```csharp
.Where(pi => pi.CanWrite || pi.PropertyType.IsListOrDictionaryType())
```
It seems like it was intended that the `List<T>` type (but not `IList<T>` or `ICollection<T>`) be considered writable despite the property being read-only. So a property map is automatically created, but does nothing, because `UseDestinationValue` is not set due to the lack of a member expression.
It appears that someone either went too far, or not far enough.

Anyway, in my project I tried to implement this in the short term using the extensibility mechanisms that AM provides, but ended up having to hack my way through it.
The logic would be to go through each type map, creating the missing property maps, and setting `UseDestinationValue` (because we're not creating member expressions which would get that automatically).
That could be done with `ForAllMaps()`, but the inheritance and member inclusions happen after that, and when they happen, `UseDestinationValue` is not carried over. So, unless we can have `UseDestinationValue` be carried over, there needs to be a second step for setting `UseDestinationValue` again, after the inclusions. The first extension possibility after the inclusions is the call to `IRuntimeFeature.Seal()` on each item of `TypeMap.Features`, which happens after the map expression is built. So even if `UseDestinationValue` is set at that point, it's too late. To work around that, I had to hack through reflection to overwrite the map expression.
What I would need to do this cleanly is to move the call to `IRuntimeFeature.Seal()`. It would have to be right before the map expression is built, instead of after. But I don't know what kind of impact it could have on AutoMapper.Collection, for which the feature mechanism seems to have been created.

jbogard:
AutoMapper is intended to map to DTOs where these rules don’t apply.

antoinebj:
Thanks for your replies!

I can see why one could turn a blind eye on DTOs and treat them differently in that regard, but I don't see why they would _have to_, unless precisely to satisfy serialization and mapping libraries which only check the existence or lack of setter.

But ok, say we don't do that with DTOs, it would indeed solve mapping _to_ DTOs as you say.
But what about mapping _from_ DTOs? There is this `ReverseMap()`, so even if the library is intended to map _to_ DTOs, it also appears to have code that was written with mapping _from_ DTOs in mind.

Either way, I don't mind. If you're not keen to have this behavior built-in, then I'll simply ask if you expect AM to be extensible to the point where it should be plug-able into it.
If the answer is yes, then from my (possibly flawed) analysis that I detailed above, my (also possibly flawed) conclusion is that there needs to be some slight adjustments to the Features mechanism.

lbargaoanu:
`UseDestinationValue` is inherited now, use the MyGet build. Read-only members can be mapped through constructors, with private setters or with explicit `MapFrom`.

PR #3286, opened 2019-12-17
Fast expression compiler

jogibear9988:


jbogard:
Thanks, I meant to pick this one back up! I'll pull it down to run some perf comparisons.

PR #3291, opened 2019-12-21
Add support for passing value-typed members to IncludeMembers

### LINKED ISSUE #3276: IncludeMembers with value type
I'm following up on the issue #3090 which was closed with only a workaround.

Unfortunately I cannot use that workaround because I need to provide the equality expressions to AutoMapper.Collection, and they require the generic `IMappingExpression<TSource, TDestination>`, which does not provide the option to specify the member names as strings.

I have identified the cause of that issue that throws the exception "Only member accesses are allowed."
It happens when the members are of value types, or generic types which are unconstrained to reference types (`class` constraint). In the case of the generic types, the `IncludeMembers` needs to be called from a method that is unaware of the actual types assigned to the type parameters.

The expression objects generated are then not of type `MemberExpression`, but of type `UnaryExpression`. That is because the compiler wants to box the value types (or potential value types) into an `object`, as that is the return type of the lambda expressions taken as parameter of `IncludeMembers`.

Basically, the following method is incomplete, and should be amended to look into a `UnaryExpression` when it doesn't find a `MemberExpression`:
```csharp
public static bool IsMemberPath(this LambdaExpression exp) => exp.Body.GetMembers().FirstOrDefault()?.Expression == exp.Parameters.First();
```

### Steps to reproduce

I modified the existing unit test to reproduce the issue. You can copy paste that in the solution.
Adding the `class` constraint to the generic type and changing the struct to a class will make the test pass again. If either of those 2 things isn't done, the test will fail.

```csharp
public class IncludeMembers : AutoMapperSpecBase
{
    class Source<TInner> // where TInner : class
    {
        public TInner InnerSource { get; set; }
        public OtherInnerSource OtherInnerSource { get; set; }
    }
    class InnerSource
    {
        public string Description { get; set; }
    }
    struct OtherInnerSource // replace with class
    {
        public string Description { get; set; }
        public string Title { get; set; }
    }
    class Destination
    {
        public string Description { get; set; }
        public string Title { get; set; }
    }
    protected override MapperConfiguration Configuration => new MapperConfiguration(cfg=>
    {
        var map = cfg.CreateMap<Source<InnerSource>, Destination>();
        SetupMap(map);
        cfg.CreateMap<InnerSource, Destination>(MemberList.None);
        cfg.CreateMap<OtherInnerSource, Destination>(MemberList.None);
    });
    [Fact]
    public void Should_flatten()
    {
        var source = new Source<InnerSource> { Name = "name", InnerSource = new InnerSource { Description = "description" }, OtherInnerSource = new OtherInnerSource { Title = "title" } };
        var destination = Mapper.Map<Destination>(source);
        destination.Name.ShouldBe("name");
        destination.Description.ShouldBe("description");
        destination.Title.ShouldBe("title");
    }
    private static void SetupMap<TInner>(IMappingExpression<Source<TInner>, Destination> map) // where TInner: class
    {
        map.IncludeMembers(src => src.InnerSource, src => src.OtherInnerSource);
    }
}

```

On a side note, some while ago I submitted a PR for AutoMapper.Collection, which fixed an oversight also related to value types. Those poor value types are begging for attention 😉

antoinebj:
Fixes #3276 

I decided to go with replacing the expressions if they have a body that casts to object.
It's the strategy that has the smallest footprint on the source code.

### DIFF
--- a/src/AutoMapper/ExpressionExtensions.cs
+++ b/src/AutoMapper/ExpressionExtensions.cs
@@ -69,14 +69,5 @@ namespace AutoMapper
 
         public static Expression IfNullElse(this Expression expression, Expression then, Expression @else = null)
             => ExpressionFactory.IfNullElse(expression, then, @else);
-
-        public static LambdaExpression[] WithoutCastToObject<T>(this Expression<Func<T, object>>[] expressions)
-            => Array.ConvertAll(
-                expressions,
-                e =>
-                {
-                    var bodyIsCastToObject = (e.Body.NodeType == ExpressionType.Convert || e.Body.NodeType == ExpressionType.ConvertChecked) && e.Body.Type == typeof(object);
-                    return bodyIsCastToObject ? Lambda(((UnaryExpression)e.Body).Operand, e.Parameters) : e;
-                });
     }
 }
--- a/src/AutoMapper/ExpressionExtensions.cs
+++ b/src/AutoMapper/ExpressionExtensions.cs
@@ -69,5 +69,14 @@ namespace AutoMapper
 
         public static Expression IfNullElse(this Expression expression, Expression then, Expression @else = null)
             => ExpressionFactory.IfNullElse(expression, then, @else);
+
+        public static LambdaExpression[] WithoutCastToObject<T>(this Expression<Func<T, object>>[] expressions)
+            => Array.ConvertAll(
+                expressions,
+                e =>
+                {
+                    var bodyIsCastToObject = (e.Body.NodeType == ExpressionType.Convert || e.Body.NodeType == ExpressionType.ConvertChecked) && e.Body.Type == typeof(object);
+                    return bodyIsCastToObject ? Lambda(((UnaryExpression)e.Body).Operand, e.Parameters) : e;
+                });
     }
 }
--- a/src/UnitTests/IMappingExpression/IncludeMembers.cs
+++ b/src/UnitTests/IMappingExpression/IncludeMembers.cs
@@ -1420,26 +1420,4 @@ namespace AutoMapper.UnitTests.IMappingExpression
             cfg.CreateMap<InnerSource, Destination>(MemberList.None);
         });
     }
-    public class IncludeMembersFromUnconstrainedGenericValidation : AutoMapperSpecBase
-    {
-        class Source<T>
-        {
-            public T InnerSource { get; set; }
-        }
-        class InnerSource
-        {
-            public string Name { get; set; }
-        }
-        class Destination
-        {
-            public string Name { get; set; }
-        }
-        protected override MapperConfiguration Configuration => new MapperConfiguration(cfg =>
-        {
-            var map = cfg.CreateMap<Source<InnerSource>, Destination>();
-            IncludeMembersFromUnconstrainedGeneric(map);
-            cfg.CreateMap<InnerSource, Destination>(MemberList.None);
-        });
-        private static void IncludeMembersFromUnconstrainedGeneric<T>(IMappingExpression<Source<T>, Destination> map) => map.IncludeMembers(src => src.InnerSource);
-    }
 }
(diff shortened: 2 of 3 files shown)


src/AutoMapper/ExpressionExtensions.cs
```diff
@@ -69,5 +69,14 @@ public static Expression NullCheck(this Expression expression, Type destinationT
...
         public static Expression IfNullElse(this Expression expression, Expression then, Expression @else = null)
             => ExpressionFactory.IfNullElse(expression, then, @else);
+
+        public static LambdaExpression[] WithoutCastToObject<T>(this Expression<Func<T, object>>[] expressions)
+            => Array.ConvertAll(
+                expressions,
+                e =>
+                {
+                    var bodyIsCastToObject = (e.Body.NodeType == ExpressionType.Convert || e.Body.NodeType == ExpressionType.ConvertChecked) && e.Body.Type == typeof(object);
+                    return bodyIsCastToObject ? Lambda(((UnaryExpression)e.Body).Operand, e.Parameters) : e;
```
lbargaoanu: We don't need ConvertChecked here.

src/UnitTests/IMappingExpression/IncludeMembers.cs
```diff
@@ -1400,4 +1400,46 @@ public class CreateCustomerDto : CustomerDtoBase
...
+        {
+            public string Name { get; set; }
+        }
+        protected override MapperConfiguration Configuration => new MapperConfiguration(cfg =>
+        {
+            var map = cfg.CreateMap<Source<InnerSource>, Destination>();
+            IncludeMembersFromUnconstrainedGeneric(map);
+            cfg.CreateMap<InnerSource, Destination>(MemberList.None);
+        });
+        private static void IncludeMembersFromUnconstrainedGeneric<T>(IMappingExpression<Source<T>, Destination> map) => map.IncludeMembers(src => src.InnerSource);
```
lbargaoanu: I don't see the need for this function.

src/AutoMapper/ExpressionExtensions.cs
```diff
@@ -69,5 +69,14 @@ public static Expression NullCheck(this Expression expression, Type destinationT
 
         public static Expression IfNullElse(this Expression expression, Expression then, Expression @else = null)
             => ExpressionFactory.IfNullElse(expression, then, @else);
+
+        public static LambdaExpression[] WithoutCastToObject<T>(this Expression<Func<T, object>>[] expressions)
```
lbargaoanu: This is used in just one place, so just make it a local function at the call site.

src/UnitTests/IMappingExpression/IncludeMembers.cs
antoinebj: Without this function, the compiler can see that the type parameter (`InnerSource`) is a reference type and does not include a casting expression in the expression tree.

lbargaoanu: Just remove this test then. It's weird and not a use case we encourage.

src/AutoMapper/ExpressionExtensions.cs
lbargaoanu: You didn't move this.

antoinebj: What do you mean? I removed it from the ExpressionExtensions.cs file, which is now untouched by this PR. The code is now in the only method that uses it.

src/UnitTests/IMappingExpression/IncludeMembers.cs
antoinebj: I was also very hesitant to include it, so ok I'll remove it.

src/AutoMapper/ExpressionExtensions.cs
lbargaoanu: _Now_  it's gone :)

PR #3305, opened 2020-01-20
WIP: EnumerableToReadOnlyDictionaryMapper

### LINKED ISSUE #3300: Exception when mapping from List to IReadOnlyDictionary
### Source/destination types

```csharp
public class Source
{
    public List<int> Field;
}
public class Target
{
    public IReadOnlyDictionary<int, int> Field;

}
```

### Mapping configuration

```csharp
new MapperConfiguration(cfg =>
{
    cfg.CreateMap<Source, Target>();
});
```

### Expected behavior

Exception with a readable message that a mapping can't build between the types and an explanation of the issue

### Actual behavior

System.IndexOutOfRangeException is thrown

### Steps to reproduce

No additional steps

mrzammit:
…tion from List to IReadOnlyDictionary. (fixes #3300)

This will now behave like a map configuration from List to Dictionary. The configuration will not throw an IndexOutOfRange, but the mapping operation will throw a mapping exception unless a custom mapper is defined.

### DIFF
--- a/src/AutoMapper/Mappers/EnumerableToReadOnlyDictionaryMapper.cs
+++ b/src/AutoMapper/Mappers/EnumerableToReadOnlyDictionaryMapper.cs
@@ -1,0 +1,44 @@
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using System.Linq;
+using System.Linq.Expressions;
+using AutoMapper.Configuration;
+using AutoMapper.Internal;
+using AutoMapper.Mappers.Internal;
+
+namespace AutoMapper.Mappers
+{
+    using static Expression;
+    using static ExpressionFactory;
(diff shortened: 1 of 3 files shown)


lbargaoanu:
OK, now I see that the mapping is supposed to succeed :)
But the implementation is almost the same. So calling a helper function should eliminate the duplicated code.
Also we need better tests. The tests should call `Map` and check the results. A test with duplicated values should be added. And any other relevant test you might think of.

mrzammit:
I appreciate the feedback. I was trying to follow the existing style like DictionaryMapper/EnumerableToDictionaryMapper. 

I will take your comments and rework my change.

mrzammit:
I'm not going to be able to complete this at the moment. Thanks.

PR #3318, opened 2020-02-08
Fix for issue 3300: System.IndexOutOfRangeException is thrown

benster21:
https://github.com/AutoMapper/AutoMapper/issues/3300

Moving MakeGenericType before the new TypePair because MakeGenericType throws the proper exception


lbargaoanu:
Check https://github.com/AutoMapper/AutoMapper/pull/3305 about how a fix would like. He had the right idea, just the code was not up to par.

PR #3336, opened 2020-02-25
Reverse the string based MapFrom

### LINKED ISSUE #3329: Reverse mapping
While I was trying to find a solution to issue #3129, I came across two related problems.
I spent a couple hours looking at the AutoMapper code and found no easy way to fix.
I guess it needs a little more work and insight and I could use some help.

Basically there are 3 things, that need a fix:

1. `ReverseMap()` works in combination with `MapFrom(...expression...)`, but doesn't with 
    `MapFrom(...sourceMemberName...)`.

    So this **works** as expected:

    ```csharp
    cfg.CreateMap<Client, ClientDto>()
       .ForMember(d => d.Id, opt => opt.MapFrom(s => s.LocalId))
       .ReverseMap();
    ```
    But this **does not work**:
    ```csharp
    cfg.CreateMap<Client, ClientDto>()
       .ForMember(d => d.Id, opt => opt.MapFrom("LocalId"))
       .ReverseMap();
    ```
    I spotted the problem, basically the implementation(s) of IPropertyMapConfiguration.Reverse() do 
    not support the case without a source expression.

2. The issue #3129 addresses that the reverse mapping does not work properly when using the AutoMapAttribute together with the SourceMemberAttribute. There are two reasons. One reason is the problem 1, which I just stated and the next one is, that AutoMapper handles the SourceMemberAttribute's after the reverse mapping has been created.
If I switch the order, all unit tests run through, so I guess it comes down to fixing problem 1.

3. When creating a `ReverseMap()` with `MapFrom(...expression...)` on a destination member and if there is a source property with the same name as the destination property, it will get reverse mapped aswell. So a ClientDto with Id = 10 will reverse map to a Client with Id = 10 and LocalId = 10. I do not expect that, because it should be a 1 to 1 mapping, right?
To fix that we would need some Ignore() for the reverse mapping, when using MapFrom.
I don't have a solution, yet. 

So here are the classes:
```csharp
public class Client
{
    public int LocalId { get; set; }

    public int Id { get; set; }
}

[AutoMap(typeof(Client), ReverseMap = true)]
public class ClientDto
{
    [SourceMember("LocalId")]
    public int Id { get; set; }
}
```

timmi-on-rails:
- Added failing tests

Fixes #3329.

### DIFF
--- a/src/AutoMapper/Configuration/MappingExpression.cs
+++ b/src/AutoMapper/Configuration/MappingExpression.cs
@@ -18,10 +18,7 @@ namespace AutoMapper.Configuration
         {
             var reversedTypes = new TypePair(Types.DestinationType, Types.SourceType);
             var reverseMap = new MappingExpression(reversedTypes, MemberList.None);
-            if(!reversedTypes.IsGenericTypeDefinition)
-            {
-                reverseMap.MemberConfigurations.AddRange(MemberConfigurations.Select(m => m.Reverse()).Where(m => m != null));
-            }
+            reverseMap.MemberConfigurations.AddRange(MemberConfigurations.Select(m => m.Reverse()).Where(m => m != null));
             ReverseMapExpression = reverseMap;
             reverseMap.IncludeMembers(MapToSourceMembers().Select(m => m.DestinationMember.Name).ToArray());
             foreach(var includedMemberName in IncludedMembersNames)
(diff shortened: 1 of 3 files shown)


timmi-on-rails:
Here we go, I am done.
I tryed to keep the changes minimalistic, because I don't understand everything along the way, so I had to rely on a hopefully good test coverage.

### Feedback codebase
For me it was hard to follow the code, because the open generic cases are always treated special and this makes the code more complex/nested/partially redundant.
I guess the open generic case should be handled more naturally.
Meaning, we need some abstraction for the `LambdaExpression` that can handle the open generic case.
A good starting point might be the `IPropertyMapConfiguration` interface. Maybe the 
`LamdbaExpression` could be replaced by `MemberPath`.
Haven't thought about deeply though.

lbargaoanu:
If this fixes the attributes stuff, you might want to add a test for it.

timmi-on-rails:
It doesn't fix attributes, but I can work out a fix and test for that.
Do you want me to put it in this merge request here or separate?

lbargaoanu:
Separate.

timmi-on-rails:
Ok, your changes look good.
As soon as you merged I will create the next pull request.

lbargaoanu:
@jbogard Merge?

PR #3349, opened 2020-03-19
Allow recursive queries

lbargaoanu:
For EFCore.


src/AutoMapper/AdvancedConfiguration.cs
```diff
@@ -40,5 +40,7 @@ public class AdvancedConfiguration
         public List<IExpressionResultConverter> QueryableResultConverters { get; } = ExpressionBuilder.DefaultResultConverters();
 
         public List<IExpressionBinder> QueryableBinders { get; } = ExpressionBuilder.DefaultBinders();
+
```
jbogard: I think this needs XML comments

src/AutoMapper/MapperConfiguration.cs
```diff
@@ -86,7 +87,9 @@ public void Validate(ValidationContext context)
 
         public IEnumerable<IExpressionResultConverter> ResultConverters { get; }
 
-        public IEnumerable<IExpressionBinder> Binders { get; }
+        public IEnumerable<IExpressionBinder> Binders { get; }
+        
+        public int RecursiveQueriesMaxDepth { get; }
```
lbargaoanu: A little late perhaps, but maybe we should move the `ProjectTo` configuration to an inner object? I think this stuff doesn't even warrant a major version bump. Just sayin' :)

src/AutoMapper/AdvancedConfiguration.cs
```diff
@@ -40,5 +40,11 @@ public class AdvancedConfiguration
...
+        /// <summary>
+        /// How many levels deep should recursive queries be expanded.
+        /// Must be zero for EF6. Can be greater than zero for EF Core.
+        /// </summary>
+        public int RecursiveQueriesMaxDepth { get; set; }
```
lbargaoanu: A little late perhaps, but maybe we should move the `ProjectTo` configuration to an inner object? I think this stuff doesn't even warrant a major version bump. Just sayin' :)

jbogard: Ah yeah, a little late...

PR #3354, opened 2020-03-20
disregard Convert node in ForPath

gsgabrielsouza:
Hello, I'm opening PR for a small change when creating lambda expressions dynamically, before it was a problem when using Expression.Convert because the verification in the ForPath method returned that it was not a valid path. I currently remove the 'Convert' method at the time of validation and simply read the expression. I will create a test to exemplify, I hope to be helping.


lbargaoanu:
This is not a reasonable use case. Perhaps I didn't make myself clear. Just because you can write an expression that breaks AM, that doesn't mean AM is the problem. So if you can come up with a proper failing test, using the API as intended, we'll take the PR.

Issue #3355, opened 2020-03-21
Feature: reverse naming conventions


jez9999:
I'm creating this issue to go with my pull request that implements the option to reverse source/destination member naming conventions when `ReverseMap()` is used on a mapping configuration.

Issue #3351, opened 2020-03-19
ReverseMap() doesn't reverse naming conventions


jez9999:
I'm aware that this bug has been submitted a couple of times before, but I don't understand why a fix hasn't been put in place.  Looking at this recent report of the issue:
https://github.com/AutoMapper/AutoMapper/issues/3259

The conclusion is:
> "I had to use two different profiles for mapping the models' bidirectionally based on a naming convention. Can't the ReverseMap() do this for us?"
> 
> "I don't see how. But a PR is welcome."

Don't see how?  How about, when generating the mappings during `ReverseMap`, have the option to pass in a boolean to `ReverseMap` that swaps `SourceMemberNamingConvention` and `DestinationMemberNamingConvention` for the mappings that it sets up?  Why is that so impossible?

lbargaoanu:
Well, give it a shot then.

jez9999:
@lbargaoanu Why close this, then?  Leave it open a while for me to submit a pull request?!

jez9999:
@jbogard What do you think about this?  If you're using naming conventions to map A to B, I can't really think of a circumstance where you wouldn't want to reverse those conventions when going the other way.  As it stands, I'd say `ReverseMap` is only useful right now if the naming conventions are the same for both objects.

lbargaoanu:
We don't need an issue, we need a PR. Feel free to discuss :)

jez9999:
You keep issues open when you acknowledge that something is a valid bug or feature request, but doesn't yet have a PR.  Almost all projects I've seen have taken this approach.  It's how you track stuff that is open for development, but hasn't yet been developed.  You don't close it if it's a valid request.

PR #3356, opened 2020-03-21
Reverse naming conventions for reverse maps by default

jez9999:
Fixes issue #3355 (described fully in #3351).


jez9999:
@jbogard Any chance of a review on this one?

jbogard:
So I think this is an easy, but wrong way to fix this problem. To me, the larger problem we have here is that there is configuration that affects a group of `TypeMap`s, and currently the *only way* to affect of type map creation is a `Profile`. Additionally, the final configuration result, `ProfileMap` is standalone. The most you can do is give it the root configuration, and those values are copied over.

The way you can adjust configuration so specific *kinds* of TypeMaps is what you're doing - applying a flag and propagating it everywhere. That's something I generally want to avoid.

Affecting `TypeMap` creation is even more difficult - the only hook for that is anything on a `ProfileMap` as it stands today. Even something like `ForAllTypeMaps` doesn't let you modify configuration in the `ProfileMap`.

As I see it, if we don't want to start adding one-off flags for behaviors like this, we can:

- Apply a sensible default - like automatically reverse the naming conventions anyway. Usually they're the same.
- Devise some new way of applying configuration to groups of `TypeMaps`, but in such a way that the result is the thing getting passed down to `TypeMapFactory`.

@lbargaoanu it would be a breaking behavioral change to reverse naming conventions on `ReverseMap`, but that does seem rather reasonable. Your only other option today is to separate profiles, but then it's no longer a reverse map.

lbargaoanu:
Yes, the fix is arbitrary. The way I see it, reverse maps are not that special. If the setting is per profile, I wouldn't want to special case reverse maps. But we could say that the setting is per map. We could then allow users to define their own extension method to opt in or to be able to configure reverse maps some other way. Or do it by default as you've said and take the breaking change. It could be done, but I'm not sure it's worth the trouble. It's clumsy, but it works. I don't think many people complained over the years. And `ReverseMap` is already complicated enough.

jez9999:
> 
> 
> Yes, the fix is arbitrary. The way I see it, reverse maps are not that special. If the setting is per profile, I wouldn't want to special case reverse maps. But we could say that the setting is per map. We could then allow users to define their own extension method to opt in or to be able to configure reverse maps some other way. Or do it by default as you've said and take the breaking change. It could be done, but I'm not sure it's worth the trouble. It's clumsy, but it works. I don't think many people complained over the years. And `ReverseMap` is already complicated enough.

AutoMapper has the naming convention functionality already.  It seems bizarre to say "this feature works fine unless you're reversing the mapping".  Surely that is a bug.  Yeah, I set 2 separate mappers up now to handle it.  But it's inelegant.

jbogard:
Well, reverse mapping was only recently a "thing" in AutoMapper. Of course lots of things don't make sense in the reverse direction - it was never originally designed to do this (I never use reverse mapping on any of my projects either). Just because it doesn't work the way you like, does not make it a bug.

I don't want to make the overall design worse with reverse mapping, since it's not a thing I nor any of my clients use. But I'm OK making reasonable defaults, it's just that naming conventions happen so early in the process there's no hook for it.

Rather than make this configurable, I'd vote for just reversing them by default. If you don't want that behavior, THEN we can have you split the mappings.

jez9999:
Suits me.

lbargaoanu:
Then the mappping configuration should have the naming conventions reversed and the code applying the conventions should just use what's set in the configuration.

jbogard:
The problem is naming conventions aren't applied at a `TypeMap` level, they're applied at a `ProfileMap` level. You can't reverse the naming conventions on a type map, you can only reverse how they're applied when we create them.

lbargaoanu:
That's why I was saying it makes sense for the setting to be changed per map. If it's per profile, why change it for a reverse map, it's in the same profile after all :)

jbogard:
The reverse mapping does a number of helpful things though, reversing the other configuration. To me this is in a similar vein - why wouldn't you want the member naming convention to be reversed?

lbargaoanu:
All I'm saying it makes a lot more sense if somehow the naming convention is per map, not per profile (the profile setting is just the default). Then, as you say, `ReverseMap` will just reserve the base map configuration.

jbogard:
Oh my god this api compat file

lbargaoanu:
I guess we should exclude somehow what's public but not in the API.

jbogard:
I modified this PR to reverse the naming conventions based on if the type map is a reverse map or not.

The naming conventions stuff tbqh I don't really understand too much, but we don't set naming conventions on a type map. We use the results of applying a naming convention to set a type map, so I thought the next best thing would be to at least mark a type map as a reverse map.

lbargaoanu:
It's ad hoc. But I do agree that code is difficult to understand/change.

jbogard:
It is ad-hoc, but I thought rather than the profiles or type maps having instructions for lower-level components, instead they describe what they are and lower level components adjust their mapping strategies accordingly.

lbargaoanu:
So you didn't want to simply pass the naming conventions to that function.

lbargaoanu:
Granted, that interface is terrible :)

src/AutoMapper/ProfileMap.cs
```diff
@@ -99,7 +99,7 @@ public void Register(IConfigurationProvider configurationProvider)
 
                 if (config.ReverseTypeMap != null)
                 {
-                    BuildTypeMap(configurationProvider, config.ReverseTypeMap);
+                    BuildTypeMap(configurationProvider, config.ReverseTypeMap, true);
```
lbargaoanu, line 102:
```suggestion
                    BuildTypeMap(configurationProvider, config.ReverseTypeMap, isReverseMap: true);

```

src/AutoMapper/TypeMap.cs
```diff
@@ -72,6 +73,8 @@ public PathMap FindOrCreatePathMapFor(LambdaExpression destinationExpression, Me
         public Type SourceType => SourceTypeDetails.Type;
         public Type DestinationType => DestinationTypeDetails.Type;
 
+        public bool IsReverseMap { get; }
```
lbargaoanu, line 76: Maybe have this flag on `MappingExpression` instead?

jbogard: I'm working on this but it needs a bit of refactoring since those objects aren't available everywhere we perform this reverse map creation.

jbogard: Moved this to `ITypeMapConfiguration`

jbogard:
I think we might have to figure out some other solution for the ApiCompat file. These merge conflicts are killing me

PR #3326, opened 2020-02-21
MapperConfigurationExpression.AddProfile use to ServiceCtor.

xmtoken:
MapperConfigurationExpression.AddProfile uses Activator.CreateInstance without using the services set in ConstructServicesUsing.
Modified to refer to ServiceCtor.

PR #3357, opened 2020-03-25
fix resolve

cr1mp:


lbargaoanu:
Duplicate of #3326.

jbogard:
@cr1mp just as a background, we've generally found DI for static configuration like profiles to be a bad idea. Instead, your maps themselves need to use DI.

PR #3389, opened 2020-04-22
Updated MemberFinderVisitor to only check members of the source type

### LINKED ISSUE #3386: MemberList.Source doesn't work with referenced objects
### Source/destination types

```csharp
public class A
{
    public string Text { get; set; }
}

public class B
{
    public string Text { get; set; }

    public string NotMapped { get; set; }
}
```

### Mapping configuration

```csharp
{
    cfg.CreateMap<A, B>(MemberList.Source)
       .ForMember(b => b.Text, x => x.MapFrom(a => a.Text ?? string.Empty));
});
```

### Expected behavior

`config.AssertConfigurationIsValid()` should not throw an exception.

### Actual behavior

`AutoMapperConfigurationException` was thrown with the following text:

> Unmapped members were found. Review the types and members below. (...)
=================================================
A -> B (Source member list)
Unmapped properties:
Text

### Steps to reproduce

If I change `x.MapFrom(a => a.Text ?? string.Empty)` to `x.MapFrom(a => a.Text ?? "")`, it works without any problems.

So to reproduce the issue, a `MemberList.Source` must be used, then in a `ForMember` declaration an "outside" object should be used, even if it's a simple static member or a variable in the same scope.

btihanyi:
Fixes #3386

### DIFF
--- a/src/AutoMapper/DefaultMemberMap.cs
+++ b/src/AutoMapper/DefaultMemberMap.cs
@@ -43,7 +43,7 @@ namespace AutoMapper
             {
                 if (CustomMapExpression != null)
                 {
-                    var finder = new MemberFinderVisitor();
+                    var finder = new MemberFinderVisitor(TypeMap.SourceType);
                     finder.Visit(CustomMapExpression);
 
                     if (finder.Member != null)
--- a/src/AutoMapper/MemberFinderVisitor.cs
+++ b/src/AutoMapper/MemberFinderVisitor.cs
@@ -1,14 +1,26 @@
+using System;
 using System.Linq.Expressions;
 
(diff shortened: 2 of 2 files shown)


lbargaoanu:
You can always add an `if` to cover another case. I don't know :) It seems to me like the current situation is more clear.

btihanyi:
I could have added a `DoNotValidate()` call to those members, but I think the `MemberFinderVisitor` class should be updated because it gives back the first member expression it finds, including `string.Empty` in my issue example.

lbargaoanu:
Yes, because it's supposed to work only with simple expressions. It shouldn't even be called for your expression.

Issue #3391, opened 2020-04-23
Allow to change mapping behaviour maxdepth on lists


kfirp-bs:
### Source/destination types

```csharp
class NodeEntity : BaseEntity
{
   public string Name { get; set; }
   public List<NodeEntity> Children { get; set; } = new List<NodeEntity>();
}

class NodeDto : BaseDto
{
   public string Name { get; set; }
   public List<NodeDto> Children { get; set; } = new List<NodeDto>();
}
```

### Mapping configuration

```csharp
class NodeMappingProfile : Profile
{
   private const int MaxDepth = 2;
    NodeMappingProfile()
   {
         CreateMap<NodeEntity, NodeDto>()
                .MaxDepth(MaxDepth);
         CreateMap<NodeDto, NodeEntity>()
                .MaxDepth(MaxDepth);
   }
}
```

### Version: x.y.z
 
9.0.0
### Expected behavior

When exceeding max depth, `Children` list should be empty.

### Actual behavior

When exceeding maximum depth, `Children` list is not empty, and instead of mapped objects it contains nulls

### Steps to reproduce

```csharp

        [Fact]
        public void MapToDto_TwoDegreeChildren_StopsAfterFirstDegree()
        {
            var entity = new NodeEntity()
            {
                Id = Guid.NewGuid(),
                Name = "Grandfather",
                Children =
                {
                    new NodeEntity()
                    {
                        Id = Guid.NewGuid(),
                        Name = "Father",
                        Children =
                        {
                            new NodeEntity()
                            {
                                Id = Guid.NewGuid(),
                                Name = "Son"
                            }
                        }
                    }
                }
            };

            var dto = _mapper.Map<NodeDto>(entity);

            Assert.Equal(entity.Name, dto.Name);
            Assert.Equal(entity.Id, dto.Id);
            Assert.Equal(entity.Children.Count, dto.Children.Count);
            Assert.Equal(entity.Children.First().Name, dto.Children.First().Name);
            Assert.True(dto.Children.First().Children.IsNullOrEmpty()); // FAILS HERE - containing list of size 1 with a null element
        }
```

It was already discussed here: https://github.com/AutoMapper/AutoMapper/issues/2511 
I suggest to allow changing the behaviour and keeping current behaviour as default

lbargaoanu:
I think you should implement that in your own code.

kfirp-bs:
As a workaround I've added AfterMap to clear nulls

[...]

Issue #2511, opened 2018-01-26
MaxDepth on list sets the containing objects to null


JonnyMattsson:
When having a maxdepth setting the behaviour is different depending on if it is a single object or a list. 
A single object is set to null which is fine, but a list contains of null elements instead of being an empty list.

### Mapping configuration

```csharp
CreateMap<Company, CompanyResponse>().MaxDepth(2);
```

### Version: x.y.z
6.2.2

### Expected behavior

  "parentCompany": null,
    "subCompanies": [ ]

### Actual behavior

  "parentCompany": null,
    "subCompanies": [
      null,
      null,
      null,
      null,
      null,
      null
    ]

lbargaoanu:
Try setting it to 1 or 3.

lbargaoanu:
A repro would help. Make a [gist](https://gist.github.com/lbargaoanu/9c7233441c3a3413cc2b9b9ebb5964a9) that we can execute and see fail.

lbargaoanu:
Another thing is, this seems to be [the desired behavior](https://github.com/AutoMapper/AutoMapper/blob/16d508a120a5f3a2df402e58497e789d85fd8979/src/UnitTests/MaxDepthTests.cs#L65) :) cc @jbogard

jbogard:
Yep, I don't think we'll ever have consensus on how to count the depth.

PR #3393, opened 2020-04-23
Avoid collections with null values when reaching max depth

lbargaoanu:
Breaking change. From #3391 and #2511. Another `if`, another dollar :) We can always keep the current behaviour.


lbargaoanu:
All looks good, I'll rebase when needed.

PR #3401, opened 2020-05-04
Matching source values should override constructor parameters default…

### LINKED ISSUE #3378: Constructor mapping with optional arguments
At least the resolving part should be common to the other member maps. One problem is that optional values override matching values from the source.
Evolving things is cumbersome. I recently did that and added another `if`.

lbargaoanu:
… values. Fixes #3378.


lbargaoanu:
After we merge this, perhaps it would be a good time for a new release.

lbargaoanu:
Merge this so I can rebase the other?

PR #3411, opened 2020-05-11
EF Core; generalize constructor mapping for ProjectTo

### LINKED ISSUE #3410: ProjectTo constructor mapping
This issue is probably related to #3110 .
But i dont get the reason, if i write the Select-Projection by hand everything is fine.

### Source/destination types

```csharp
    public class A
    {
        public int Id { get; set; }

        public B B { get; set; }
    }

    public class B
    {
        public int I { get; }
    }

    public class DtoA
    {
        public DtoB B { get; }

        public DtoA(DtoB b)
        {
            B = b;
        }
    }

    public class DtoB
    {
        public int I { get; }

        public DtoB(int i)
        {
            I = i;
        }
    }
```

### Mapping configuration

```csharp
c.CreateMap<A, DtoA>();
c.CreateMap<B, DtoB>();
```

### Version

Automapper: 9.0.0 
EFCore: 3.1.3

### Expected behavior

```csharp
await context.Set<A>().AsNoTracking().ProjectTo<DtoA>(Mapper.ConfigurationProvider).FirstAsync();
```
 should have the same result as
```csharp
var a = mapper.Map<DtoA>(a);
```

### Actual behavior

`ProjectTo()` is throwing the following exception, while `mapper.Map` is working as expected:

```
System.ArgumentException : Expression of type 'B' cannot be used for constructor parameter of type 'DtoB' (Parameter 'arguments[0]')
```

Gist to reproduce:

https://gist.github.com/prochnowc/901b53f5fb2b9549d4f152d7b54e2c67

lbargaoanu:
Fixes #3410.

### DIFF
--- a/src/AutoMapper/QueryableExtensions/ExpressionBuilder.cs
+++ b/src/AutoMapper/QueryableExtensions/ExpressionBuilder.cs
@@ -127,22 +127,16 @@ namespace AutoMapper.QueryableExtensions
 
         private Expression CreateMapExpressionCore(ExpressionRequest request, Expression instanceParameter, TypePairCount typePairCount, LetPropertyMaps letPropertyMaps, out TypeMap typeMap)
         {
-            typeMap = _configurationProvider.ResolveTypeMap(request.SourceType, request.DestinationType);
-
-            if(typeMap == null)
-            {
-                throw QueryMapperHelper.MissingMapException(request.SourceType, request.DestinationType);
-            }
-
-            if(typeMap.CustomMapExpression != null)
-            {
-                return typeMap.CustomMapExpression.ReplaceParameters(instanceParameter);
-            }
+            typeMap = _configurationProvider.ResolveTypeMap(request.SourceType, request.DestinationType) ?? throw QueryMapperHelper.MissingMapException(request.SourceType, request.DestinationType);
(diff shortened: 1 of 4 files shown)


src/AutoMapper/QueryableExtensions/ExpressionBuilder.cs
```diff
@@ -229,39 +222,51 @@ private static int GetDepth(ExpressionRequest request, TypePairCount typePairCou
...
 
-            protected override Expression VisitNew(NewExpression node)
+        public Expression NewExpression(ConstructorMap constructorMap, ExpressionRequest request, Expression instanceParameter, TypePairCount typePairCount, LetPropertyMaps letPropertyMaps)
+        {
+            var parameters = constructorMap.CtorParams.Select(map =>
             {
-                NewExpression = node;
-                return base.VisitNew(node);
-            }
+                var resolvedSource = ResolveExpression(map, request.SourceType, instanceParameter, letPropertyMaps);
```
lbargaoanu, line 243: I guess what should happen here is the same as for property maps. But that seems like a big change.

lbargaoanu:
No more conflicts I believe.

src/AutoMapper/QueryableExtensions/ExpressionBuilder.cs
jbogard: Yeah, let's see how this goes first.

PR #3415, opened 2020-05-13
Exclude some public members from api checks

lbargaoanu:
We can keep adding the attribute on new PR-s. I expect less conflicts. When we do, we can squash (it's built-in in VS now) and solve the conflicts mechanically (just take whatever version). The file will be regenerated.


lbargaoanu:
@jbogard Can we merge this?

PR #3424, opened 2020-05-20
Allow any expression with IncludeMembers

lbargaoanu:
The idea is to cache the included members. In a variable for `Map` and in a let clause for `ProjectTo`. I guess _any_ expression doesn't necessarily sound like a good idea. We'll see :)


lbargaoanu:
@jbogard Review? :)

lbargaoanu:
I don't think that restriction is mentioned anywhere. And for people already using it wouldn't be of much help. But I'll add smth if you think it's useful.

PR #3438, opened 2020-06-05
Clean up Mapper and friends

lbargaoanu:
Extracted `IMapperBase` for consistency. That causes some problems with dynamic dispatch in net461, hence the test changes. We pass the context ourselves, so I hid those overloads. We could simply drop `ResolutionContext.Mapper` and have `ResolutionContext.Map` methods, but that would be more impactful. I guess we could also have both.


src/AutoMapper/ResolutionContext.cs
```diff
@@ -1,21 +1,18 @@
...
-    public class ResolutionContext : IRuntimeMapper
+    public class ResolutionContext : IInternalRuntimeMapper
     {
         private Dictionary<ContextCacheKey, object> _instanceCache;
         private Dictionary<TypePair, int> _typeDepth;
-        private readonly IRuntimeMapper _inner;
+        private readonly IInternalRuntimeMapper _inner;
 
-        public ResolutionContext(IMappingOperationOptions options, IRuntimeMapper mapper)
+        internal ResolutionContext(IMappingOperationOptions options, IInternalRuntimeMapper mapper)
```
lbargaoanu, line 15: If we must, we can keep this public and make `IInternalRuntimeMapper` public.

docs/Lists-and-arrays.md
```diff
@@ -60,6 +60,7 @@ var configuration = new MapperConfiguration(cfg => {
     cfg.CreateMap<Source, Destination>();
 });
 ```
+The setting can be applied globally and can be overridden per profile and per member with `AllowNull` and `DoNotAllowNull`.
```
lbargaoanu, line 63: Perhaps I should make a property `AllowNull`.

lbargaoanu:
@jbogard Review?

src/AutoMapper/QueryableExtensions/ExpressionBuilder.cs
```diff
@@ -268,45 +268,14 @@ private ExpressionResolutionResult ResolveExpression(IMemberMap propertyMap, Typ
                 return null;
             }
             result = matchingExpressionConverter.GetExpressionResolutionResult(result, propertyMap, letPropertyMaps);
-            if(propertyMap.NullSubstitute != null && result.Type.IsNullableType())
+            if(propertyMap.NullSubstitute != null && (result.Type.IsNullableType() || result.Type == typeof(string)))
```
lbargaoanu, line 271: Perhaps remove the type check altogether? And simply not use `NullSubstitute` when it doesn't make sense.

src/AutoMapper/QueryableExtensions/ExpressionBuilder.cs
```diff
@@ -268,45 +268,12 @@ private ExpressionResolutionResult ResolveExpression(IMemberMap propertyMap, Typ
                 return null;
             }
             result = matchingExpressionConverter.GetExpressionResolutionResult(result, propertyMap, letPropertyMaps);
-            if(propertyMap.NullSubstitute != null && result.Type.IsNullableType())
+            if(propertyMap.NullSubstitute != null && result.ResolutionExpression is MemberExpression && (result.Type.IsNullableType() || result.Type == typeof(string)))
```
lbargaoanu, line 271: Perhaps remove the type check altogether? And simply not use NullSubstitute when it doesn't make sense.

lbargaoanu:
@jbogard Ping :)

src/AutoMapper/ResolutionContext.cs
jbogard: Meh.

PR #3448, opened 2020-06-25
10.0 upgrade guide and some refactorings

lbargaoanu:


src/AutoMapper/Mappers/EnumToEnumMapper.cs
```diff
@@ -1,44 +1,26 @@
...
+            var destinationType = destExpression.Type;
+            var sourceToObject = sourceExpression.ToObject();
+            var toObject = Call(typeof(Enum), "ToObject", Type.EmptyTypes, Constant(destinationType), sourceToObject);
+            var castToObject = Convert(toObject, destinationType);
+            var isDefined = Call(typeof(Enum), "IsDefined", Type.EmptyTypes, Constant(sourceExpression.Type), sourceToObject);
+            var sourceToString = Call(sourceExpression, "ToString", Type.EmptyTypes);
+            var result = Variable(destinationType, "destinationEnumValue");
+            var ignoreCase = Constant(true);
+            var tryParse = Call(typeof(Enum), "TryParse", new[] { destinationType }, sourceToString, ignoreCase, result);
+            return Block(new[] { result }, Condition(isDefined, Condition(tryParse, result, castToObject), castToObject));
```
lbargaoanu, line 23: A different approach would be to cache the result by value. That might take a lot of memory, its true. And someone can always implement that outside AM if they want.

Issue #3422, opened 2020-05-19
Odd behavior in Explicit Expand with inherited models


JobaDiniz:
The full plug-and-play test can be found here: https://gist.github.com/JobaDiniz/628ceda9133843fad447152f1b9e8cc9 - just add to the test project, run it and you'll see the following results:

![image](https://user-images.githubusercontent.com/165290/82378001-d86b3600-99fa-11ea-909c-210968cc0162.png)

There are 2 test classes in the gist. The `ExplicitExpansionWithIncludeBase` explores the bug. The `ExplicitExpansionWithIncludeBaseUsingNewKeyword` shows that using the `new` keyword in the properties of inherited classes makes the tests pass, which is odd.

This comes from [2 .5 years ago](https://stackoverflow.com/questions/46914089/mapping-properties-from-base-class-doesnt-work-using-explicitexpansion) and I strongly believe it's a bug in AutoMapper.
It seems that AutoMapper expands members regardless their depth in the object graph, which in my opinion is wrong.

### Source/destination types

```csharp
/*********** Source Types ***********/
        abstract class EntityBase
        {
            public EntityBase() => Id = Guid.NewGuid();
            public Guid Id { get; set; }
            public User CreatedBy { get; set; }
            public User ModifiedBy { get; set; }
        }

        class User { }
        class Computer : EntityBase { }
        class Script : EntityBase
        {
            public Computer Computer { get; set; }
        }

        /*********** Destination Types ***********/
        class EntityBaseModel
        {
            public Guid Id { get; set; }
            public UserModel CreatedBy { get; set; }
            public UserModel ModifiedBy { get; set; }
        }

        class UserModel { }
        class ComputerModel : EntityBaseModel
        {
            public new UserModel CreatedBy { get; set; } //NOTE: the 'new' keyword here make the tests pass
        }
        class ScriptModel : EntityBaseModel
        {
            public new UserModel CreatedBy { get; set; } //NOTE: the 'new' keyword here make the tests pass
            public ComputerModel Computer { get; set; }
        }
```

### Mapping configuration

```csharp
 private IConfigurationProvider BuildConfiguration()
        {
            return new MapperConfiguration(cfg =>
            {
                cfg.CreateMap<User, UserModel>();
                cfg.CreateMap<EntityBase, EntityBaseModel>()
                .MaxDepth(1)
                .ForMember(d => d.ModifiedBy, o => o.ExplicitExpansion())
[...]

PR #3473, opened 2020-07-22
Fix for Issue #3422

BlaiseD:
Fix for Issue #3422

### DIFF
--- a/src/AutoMapper/MapMemberInfo.cs
+++ b/src/AutoMapper/MapMemberInfo.cs
@@ -1,0 +1,58 @@
+﻿using System;
+using System.ComponentModel;
+using System.Reflection;
+
+namespace AutoMapper
+{
+    /// <summary>
+    /// Used to ensure we capture the concrete parent type. For virtual members, AutoMapper's GetMemberInfo methods will return a
+    /// MemberInfo whose ReflectedType is the base class - not source or destination being mapped.
+    /// </summary>
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public class MapMemberInfo : MemberInfo, IEquatable<MemberInfo>
+    {
+        public MapMemberInfo(MemberInfo memberInfo, Type parentType)
+        {
+            _memberInfo = memberInfo;
+            _parentType = parentType;
+        }
+
+        private MemberInfo _memberInfo { get; set; }
+        private Type _parentType { get; set; }
+
+        public override Type DeclaringType => _memberInfo.DeclaringType;
+
+        public override MemberTypes MemberType => _memberInfo.MemberType;
+
+        public override string Name => _memberInfo.Name;
+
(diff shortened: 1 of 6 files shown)


lbargaoanu:
Breaking change. And also ad-hoc I'd say. Why would the parent of the last member in the path would be so special? Taking into account the whole path, that would make a lot more sense to me. That would be a breaking change too :) But the whole idea seems kind of subtle and prone to misunderstandings. Not really about the code...

lbargaoanu:
I presume `Include` in EF (kind of similar) works by considering the whole path.

BlaiseD:
Where are you seeing "last member in the path"?  Both of the following set the parent for every item - unless I'm missing something.

```c#
                node.GetMemberExpressions()
                    .Select(e => new MapMemberInfo(e.Member, e.Expression.Type))
```

```c#
                foreach (var memberName in fullMemberName.Split('.'))
                {
                    var currentType = GetCurrentType(property, type);
                    property = currentType.GetFieldOrProperty(memberName);
                    yield return new MapMemberInfo(property, currentType);
                }
```

What would be your approach?  ([requested opinions earlier](https://github.com/AutoMapper/AutoMapper/issues/3422#issuecomment-658244272))

BlaiseD:
A breaking change would break existing tests - right? Or define breaking change.

lbargaoanu:
You interpret the same parameters in a different way. Clearly some code will break. I guess the problem is that the member list is flattened, as opposed to a structural path.

BlaiseD:
Seems to me like it's working as before with exception of the "member belongs to a base class" scenario - which is the fix.

I don't see `MemberInfo.ReflectedType` being used anywhere else in the code. 

The inheritance design is because we collect the member infos before the call to `ExpressionBuilder.GetMapExpression` -  I think changing that interface would have been a bigger break.

It's ok to come up with a new design once you've had time to think about it - this can stay on hold.  We know the underlying problem.

jbogard (approved):

src/AutoMapper/MapMemberInfo.cs
```diff
@@ -0,0 +1,58 @@
...
+        public override string Name => _memberInfo.Name;
+
+        public override Type ReflectedType => _parentType;
+
+        public bool Equals(MemberInfo other)
+        {
+            if (ReferenceEquals(null, other)) return false;
+            if (ReferenceEquals(this, other)) return true;
+
+            return this._memberInfo.Equals(other);
```
lbargaoanu, line 36: Always `false`. My co-workers traced a memory leak back to this line. The projection cache keeps growing in 10.1.1. No longer the case now, so it's fixed in 11.0.

BlaiseD: Yep -  here's from PropertyInfo source:
image

MemberInfo does the same.  Thanks for the heads-up.

PR #3487, opened 2020-08-23
Trim keys when mapping from dictionaries

### LINKED ISSUE #3483: Consider trimming keys when mapping from dictionaries
Consider improving the matching algorithm when mapping from dictionaries. The [docs](https://docs.automapper.org/en/latest/Dynamic-and-ExpandoObject-Mapping.html) states that 

> You can map straight from Dictionary<string, object> to objects, AutoMapper will line up the keys with property names.

..but white-spaces prevent the mapping, even though they're not legal characters in member names. Trimming white-spaces from these strings could help avoid some silent bugs.

### Steps to reproduce

```csharp
class Foo
{
    public int Baz { get; set; }
}

void Example()
{
    var input = new Dictionary<string, object>
    {
        {" Bar", "test"},
        {"Baz ", 1213}
    };

    var mapper = new MapperConfiguration(_ => { }).CreateMapper();

    var mapped = mapper.Map<Foo>(input);
}
```

### Expected behavior

I'd expect `mapped` to contain "test" and 1213.

### Actual behavior

 `mapped` contains just the defaults.

dim5:
- Now when mapping from string dictionaries keys surrounded with white-spaces are considered
  - Preferring the one with the least amount of white-spaces, when multiple entries are present
- Added matching tests

Fixes #3483

### DIFF
--- a/src/AutoMapper/Mappers/FromStringDictionaryMapper.cs
+++ b/src/AutoMapper/Mappers/FromStringDictionaryMapper.cs
@@ -33,16 +33,19 @@ namespace AutoMapper.Mappers
         private static TDestination Map<TDestination>(StringDictionary source, TDestination destination, ResolutionContext context, ProfileMap profileMap)
         {
             var destTypeDetails = profileMap.CreateTypeDetails(typeof(TDestination));
-            var members = from name in source.Keys
-                          join member in destTypeDetails.PublicWriteAccessors on name equals member.Name
-                          select member;
+
+            var memberMatches = from member in destTypeDetails.PublicWriteAccessors
+                                join key in source.Keys on member.Name equals key.Trim() into matchingKeys
+                                where matchingKeys.Any()
+                                select new { member, sourceName = matchingKeys.OrderBy(n => n.Length).First() };
+
             object boxedDestination = destination;
-            foreach (var member in members)
+            foreach (var match in memberMatches)
             {
-                var value = context.MapMember(member, source[member.Name], boxedDestination);
-                member.SetMemberValue(boxedDestination, value);
+                var value = context.MapMember(match.member, source[match.sourceName], boxedDestination);
+                match.member.SetMemberValue(boxedDestination, value);
             }
-            return (TDestination) boxedDestination;
+            return (TDestination)boxedDestination;
         }
     }
 }
--- a/src/AutoMapper/Mappers/FromStringDictionaryMapper.cs
+++ b/src/AutoMapper/Mappers/FromStringDictionaryMapper.cs
@@ -1,4 +1,5 @@
-﻿using System.Linq;
+﻿using System;
+using System.Linq;
 using System.Linq.Expressions;
 using System.Reflection;
 using AutoMapper.Execution;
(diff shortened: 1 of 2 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/AutoMapper/AutoMapper?pullRequest=3487) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/AutoMapper/AutoMapper?pullRequest=3487) before we can accept your contribution.<br/><br/>:x: dim5 [sign now](https://cla.dotnetfoundation.org/AutoMapper/AutoMapper?pullRequest=3487)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/AutoMapper/AutoMapper?pullRequest=3487) it.</sub>

src/AutoMapper/Mappers/FromStringDictionaryMapper.cs
```diff
@@ -33,16 +33,19 @@ public class FromStringDictionaryMapper : IObjectMapper
...
             var destTypeDetails = profileMap.CreateTypeDetails(typeof(TDestination));
-            var members = from name in source.Keys
-                          join member in destTypeDetails.PublicWriteAccessors on name equals member.Name
-                          select member;
+
+            var memberMatches = from member in destTypeDetails.PublicWriteAccessors
+                                join key in source.Keys on member.Name equals key.Trim() into matchingKeys
+                                where matchingKeys.Any()
+                                select new { member, sourceName = matchingKeys.OrderBy(n => n.Length).First() };
+
```
lbargaoanu, line 41: This is pretty much an arbitrary choice. I'd rather throw.

dim5: Was debating that myself, I'll post an update soon.

src/AutoMapper/Mappers/FromStringDictionaryMapper.cs
```diff
@@ -33,16 +34,24 @@ public class FromStringDictionaryMapper : IObjectMapper
...
             object boxedDestination = destination;
-            foreach (var member in members)
+            foreach (var match in memberMatches)
             {
-                var value = context.MapMember(member, source[member.Name], boxedDestination);
-                member.SetMemberValue(boxedDestination, value);
+                if (match.sourceNames.Count() > 1)
+                {
+                    throw new InvalidOperationException($"Multiple matching keys were found in the source dictionary for {match.member.Name}.");
```
lbargaoanu, line 48: Throw `AutoMapperMappingException` instead. And be sure to specify the types being mapped. We want to give as much details as possible.

dim5: As far as I can tell, this exception gets wrapped as an InnerException in an `AutoMapperMappingException` - which has some type info.

If I threw an `AutoMapperMappingException` instead, wouldn't it just get wrapped as well? (resulting in an `AutoMapperMappingException` containing an `AutoMapperMappingException`)

lbargaoanu: OK, but the more important point is what's in that exception. I'm guessing it does not contain the type pair we're mapping, but if it does, then you're home free. In your test, check the type pair in the exception.

lbargaoanu (approved):
Thanks!

lbargaoanu:
@jbogard I think merge this one so I can rebase my PR.

PR #3504, opened 2020-09-21
Separate internal APIs

lbargaoanu:
https://github.com/AutoMapper/AutoMapper/pull/3504#discussion_r499995066


src/AutoMapper/QueryableExtensions/ExpressionBuilder.cs
```diff
@@ -18,7 +18,7 @@ namespace AutoMapper.QueryableExtensions
 
     public interface IExpressionBuilder
```
lbargaoanu, line 19: I don't consider this a part of our API. I think we should hide `MapperConfiguration.ExpressionBuilder` and only expose it through an implementation interface. I'm sure other members can get the same treatment.

lbargaoanu: @jbogard If you like the idea, perhaps it would be easier to write the code yourself after merging this PR. I'm not sure you would agree with my choices :)

lbargaoanu: Also, we could have an extension method `Advanced()` or smth, in the `AM.Internal` namespace.

src/UnitTests/Constructors.cs
```diff
@@ -8,6 +8,18 @@
...
 namespace AutoMapper.UnitTests.Constructors
 {
+    public class ConstructorValidation : AutoMapperSpecBase
+    {
+        class Source
+        {
+        }
+        class Destination
+        {
+            public Destination(int value = 2) => Value = value;
```
lbargaoanu, line 18: People want to say that the property `Value` is mapped because this constructor is used.

src/AutoMapper/Configuration/MapperConfigurationExpression.cs
```diff
@@ -26,8 +68,6 @@ public void CreateProfile(string profileName, Action<IProfileExpression> config)
 
         public IList<IObjectMapper> Mappers { get; }
 
-        public AdvancedConfiguration Advanced { get; } = new AdvancedConfiguration();
```
lbargaoanu, line 29: Not exactly correct, but it seems reasonable to replace `Advanced` with `Internal()`. Easier to upgrade. If some setting is too important for that, we could move it in the public interface.

src/AutoMapper/InternalApi.cs
```diff
@@ -0,0 +1,175 @@
...
+        IEnumerable<ProfileMap> GetProfiles();
+    }
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public interface IProfileExpressionInternal : IProfileExpression
+    {
+        /// <summary>
+        /// Specify common configuration for all type maps.
+        /// </summary>
+        /// <param name="configuration">configuration callback</param>
+        void ForAllMaps(Action<TypeMap, IMappingExpression> configuration);
```
lbargaoanu, line 167: We could compromise and let these two be.

SingThatSong: We use ForAllMaps for perfomance optimization: disable object validation before all mappings and reenable it after, for exapmle. Please keep it public.

lbargaoanu: Everything is still public. The separation is logical.

SingThatSong: I'm not sure why you deleted my comment, but I assure you it is not public.
MapperConfiguration uses IMapperConfigurationExpression as a parameter, which is inherited directly from IProfileExpression.
You have removed ForAllMaps method from IProfileExpression, hence it's not available for configuration.

lbargaoanu: Check the tests.

src/AutoMapper/TypeDetails.cs
```diff
@@ -38,7 +38,10 @@ private void PossibleNames(ProfileMap config)
             {
                 foreach (var memberName in PossibleNames(member.Name, config.Prefixes, config.Postfixes))
                 {
-                    _nameToMember[memberName] = member;
+                    if (!_nameToMember.ContainsKey(memberName))
```
lbargaoanu, line 41: This is silly, but I don't have a better idea.

src/AutoMapper/TypeDetails.cs
```diff
@@ -138,77 +107,34 @@ select extensionMethod.MakeGenericMethod(genericInterfaceArguments)
...
-            // Scan all types for public properties and fields
-            return typesToScan
-                .Where(x => x != null) // filter out null types (e.g. type.BaseType == null)
-                .SelectMany(x => x.GetDeclaredMembers()
-                    .Where(mi => mi.DeclaringType != null && mi.DeclaringType == x)
-                    .Where(
-                        m =>
-                            m is FieldInfo && fieldAvailableFor((FieldInfo)m) ||
-                            m is PropertyInfo && propertyAvailableFor((PropertyInfo)m) &&
-                            !((PropertyInfo)m).GetIndexParameters().Any())
```
lbargaoanu, line 200: I removed this because it's expensive. I think it's rare and people can just explicitly ignore index properties.

src/AutoMapper/Configuration/IProfileConfiguration.cs
```diff
@@ -12,6 +12,8 @@ namespace AutoMapper.Configuration
     public interface IProfileConfiguration
     {
         IEnumerable<IMemberConfiguration> MemberConfigurations { get; }
+        bool? FieldMappingEnabled { get; }
+        bool? MethodMappingEnabled { get; }
```
lbargaoanu, line 16: Another approach would be to use `ShouldMap*` instead. Make them expressions and check for `false`.

src/UnitTests/ArraysAndLists.cs
```diff
@@ -673,71 +674,6 @@ public void Should_assign_the_value_directly()
...
-
-        protected override void Because_of()
-        {
-            _destination = Mapper.Map<Source, Destination>(_source);
-        }
-
-        [Fact]
-        public void Should_map_the_list_of_source_items()
-        {
-            // here not the EnumerableMapper is used, but just the AssignableMapper!
```
lbargaoanu, line 731: If it's `IEnumerable` (except `string`), it's a collection, and it should be mapped as such.

jbogard:
Ping me when you want a review 😆 but I'm gonna unsubscribe for now

jbogard (approved):

PR #3522, opened 2020-10-16
EF Core tests project

hannan-m:
Removed an extra semicolon.

### DIFF
--- a/src/IntegrationTests.Core/IncludeMembers.cs
+++ b/src/IntegrationTests.Core/IncludeMembers.cs
@@ -1,0 +1,89 @@
+﻿using System.Linq;
+using AutoMapper.UnitTests;
+using Microsoft.EntityFrameworkCore;
+using Shouldly;
+using Xunit;
+
+namespace AutoMapper.IntegrationTests.Core
+{
+    public class IncludeMembers : AutoMapperSpecBase
+    {
+        class Source
+        {
+            public int Id { get; set; }
+            public string Name { get; set; }
+            public InnerSource InnerSource { get; set; }
+            public OtherInnerSource OtherInnerSource { get; set; }
+        }
+
+        class InnerSource
+        {
+            public int Id { get; set; }
+            public string Name { get; set; }
+            public string Description { get; set; }
+        }
+
+        class OtherInnerSource
+        {
+            public int Id { get; set; }
+            public string Name { get; set; }
+            public string Description { get; set; }
+            public string Title { get; set; }
+        }
+
+        class Destination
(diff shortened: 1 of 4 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/AutoMapper/AutoMapper?pullRequest=3522) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/AutoMapper/AutoMapper?pullRequest=3522) before we can accept your contribution.<br/><br/>:x: hannan-m [sign now](https://cla.dotnetfoundation.org/AutoMapper/AutoMapper?pullRequest=3522)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/AutoMapper/AutoMapper?pullRequest=3522) it.</sub>

lbargaoanu (approved):
Thanks!

hannan-m:
Apologies, my bad made commit to the wrong branch.

src/IntegrationTests.Core/IncludeMembers.cs
```diff
@@ -0,0 +1,89 @@
...
+            public Context()
+            {
+                Database.EnsureDeleted();
+                Database.EnsureCreated();
+                Seed();
+            }
+
+            protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder)
+            {
+                optionsBuilder.UseInMemoryDatabase("Local");
```
lbargaoanu, line 53: Looks good, but this is a problem. Because we want the tests to run with SQL Server, a real database, and we want to configure that in only one place for all the tests. I don't know how one would do that with EF Core, I guess a base context class would work, but maybe there is a better way. In EF6, it's set [here](https://github.com/AutoMapper/AutoMapper/blob/f31e398a596438a939bf19d5bb97ed218124731d/src/IntegrationTests/App.config#L14).

hannan-m: Thank you for your review. I will work on changing the setup according to the need.

mycroes:
This tests with EF Core 3, which I guess many people might still be using for a while. OTOH EF Core 5 has been released, so I would suggest testing against both. Also because there's obvious issues that occur with v3 or v5 only (like the fix for v5 in AutoMapper v10.1.1, which actually breaks stuff on v3 again...)

lbargaoanu:
This is stale. A PR is welcome.

Issue #3563, opened 2021-01-25
Query projection with subquery can generate non-compiling expression


Erythnul:
Hey guys,

I've run into an issue with query projection when subqueries are included.
I've spent some time debugging and seeing if I could get it fixed up, I've got a pull request coming up with a fix that definitely works for my scenario and most probably other ones as well.
The code isn't the most elegant, but I'm not exactly at home in your source so I tried to keep my additions small and easily readable. If you've got a better way of fixing this problem, I'll gladly change the source (or feel free to change it yourself and fix up the pull request before merging).

More details in the reproduction repo:

https://github.com/Erythnul/AutomapperIssue

### Source/destination types

```csharp

//Source
public class Order
{
    public Guid Id { get; set; }
    public Guid CustomerId { get; set; }

    public int? MostImportantOrderLine { get; set; }

    public ICollection<OrderLine> OrderLines { get; set; } = null!;
}
public class OrderLine
{
    public Guid Id { get; set; }
    public Guid OrderId { get; set; }

    public int OrderLineNumber { get; set; }
    public string Description { get; set; } = null!;

    public Order Order { get; set; } = null!;
}

//Destination
public class OrderModel
{
    public OrderSubModel? OrderSubModel { get; set; }
    public OrderLineModel? MostImportantOrderLine { get; set; }
}

public class OrderSubModel
{
    public Guid? OrderId { get; set; }
}

public class OrderLineModel
{
    public Guid OrderLineId { get; set; }
    public string Description { get; set; } = string.Empty;
}
```

### Mapping configuration

```csharp
new MapperConfiguration(cfg =>
{
    cfg.CreateMap<Entities.Order, Dtos.OrderModel>()
        .ForMember(dst => dst.OrderSubModel, opt => opt.MapFrom(src => src.MostImportantOrderLine != null ? src : null))
        .ForMember(dst => dst.MostImportantOrderLine, opt =>
        {
            opt.MapFrom(src => src.OrderLines.FirstOrDefault(x => x.OrderLineNumber == src.MostImportantOrderLine));
        });

    cfg.CreateMap<Entities.Order, Dtos.OrderSubModel>()
        .ForMember(dst => dst.OrderId, opt => opt.MapFrom(src => src.Id));

    cfg.CreateMap<Entities.OrderLine, Dtos.OrderLineModel>()
        .ForMember(dst => dst.OrderLineId, opt => opt.MapFrom(src => src.Id))
[...]

PR #3565, opened 2021-01-25
Evaluate Conditional Expressions for projections with subqueries

Erythnul:
Add failing integration test as discussed in #3563


src/IntegrationTests/MemberWithSubQueryProjections.cs
```diff
@@ -0,0 +1,98 @@
...
+
+        [Fact]
+        public void Should_Create_Compilable_Expressions()
+        {
+            using (var context = new Context())
+            {
+                var resultQuery = ProjectTo<CustomerViewModel>(context.Customers);
+                var result = resultQuery.Single();
+
+                result.ShouldNotBeNull();
```
lbargaoanu, line 94: Assert here that the result has the values from the db.

Erythnul: Added the additional asserts, I only asserted the notNull because the current state throws an exception anyway, but the asserts are indeed an improvement.

lbargaoanu:
Thanks!

PR #3658, opened 2021-05-29
Updated example to work with Simple Injector v5

dotnetjunkie:
The original example doesn't work with the OOTB configuration of Simple Injector v5, because v5 requires all components to be registered, while MapperProvider isn't. This removes the need to resolve MapperProvider and simplifies the code by making MapperProvider static.

### DIFF
--- a/docs/Dependency-injection.md
+++ b/docs/Dependency-injection.md
@@ -81,36 +81,23 @@ public class MyRegistrar
         container.RegisterSingleton<IService, SomeService>();
 
         // Automapper
-        container.RegisterSingleton(() => GetMapper(container));
-    }
-
-    private AutoMapper.IMapper GetMapper(Container container)
-    {
-        var mp = container.GetInstance<MapperProvider>();
-        return mp.GetMapper();
+        container.RegisterSingleton(() => MapperProvider.GetMapper(container));
     }
 }
 
-public class MapperProvider
+public static class MapperProvider
 {
-    private readonly Container _container;
-
-    public MapperProvider(Container container)
-    {
-        _container = container;
-    }
-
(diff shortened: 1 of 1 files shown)


dotnetjunkie:
Additional note: the proposed change will make the example work with for both Simple Injector v4 and v5.

docs/Dependency-injection.md
```diff
@@ -81,36 +81,23 @@ public class MyRegistrar
...
-    {
-        _container = container;
-    }
-
-    public IMapper GetMapper()
+    public static IMapper GetMapper(Container container)
     {
         var mce = new MapperConfigurationExpression();
-        mce.ConstructServicesUsing(_container.GetInstance);
+        mce.ConstructServicesUsing(container.GetInstance);
```
lbargaoanu, line 93: I think this line is not needed.

dotnetjunkie: That's not something I can't comment on. I'm not thst familiar with AutoMapper (just with Simlle Injector), so removing such line is not a decision I can make.

lbargaoanu: Yes, but if you already have this code, please test that evertyhing still works without it. I'm unfamiliar with Simple Injector.

dotnetjunkie: That"s not up to me to test whether the code works without it. I'm just providing an updated version that works with v5 as well.

lbargaoanu: Well, if you didn't test this code, why would we consider merging this PR? :)

dotnetjunkie:
If you don't want help in getting helpful documentation for your users, feel free.to close this pull request.

PR #3727, opened 2021-09-02
Add GetProjectionExpression method to IConfigurationProvider

buvinghausen:
ProjectTo is nice but what is really needed is access to the underlying Expression<Func<TSource, TDestination>> which can then simply be passed into whatever expression capable API without having the encumbrance of IQueryable.

For example in Mongo you can pass projection expressions into their IFindFluent interface.  Even if you use the IMongoQueryable interface you have to cast it back to IMongoQueryable after you call ProjectTo.  This removes those barriers and makes using the projection capabilities easier and in a data store agnostic manner.

### DIFF
--- a/src/AutoMapper/Configuration/MapperConfiguration.cs
+++ b/src/AutoMapper/Configuration/MapperConfiguration.cs
@@ -47,6 +47,14 @@ namespace AutoMapper
         /// Use if you want AutoMapper to compile all mappings up front instead of deferring expression compilation for each first map.
         /// </summary>
         void CompileMappings();
+
+        /// <summary>
+        /// Gets the projection expression which can then be fed into any expression capable data store regardless of IQueryable compatibility
+        /// </summary>
+        /// <typeparam name="TSource">Source type</typeparam>
+        /// <typeparam name="TDestination">Destination type</typeparam>
+        /// <returns>The underlying projection expression</returns>
+        Expression<Func<TSource, TDestination>> GetProjectionExpression<TSource, TDestination>();
     }
     public class MapperConfiguration : IGlobalConfiguration
     {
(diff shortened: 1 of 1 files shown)


lbargaoanu:
Look into expression mapping.

buvinghausen:
@lbargaoanu thank you so much for the insightful feedback.

How do I get a projection expression out of AutoMapper to integrate [here](https://github.com/mongodb/mongo-csharp-driver/blob/master/src/MongoDB.Driver/IFindFluentExtensions.cs#L57)?

buvinghausen:
@lbargaoanu I looked through expression mapping and it doesn't contain the functionality I'm looking for.  I don't want to map expressions I only want to get the expression based on source and destination types.

How do I ask AutoMapper or any derivative project what the Expression<Func<TSource, TDestination>> is for the two types?

[Integration point here...](https://github.com/mongodb/mongo-csharp-driver/blob/master/src/MongoDB.Driver/IFindFluentExtensions.cs#L57)

jbogard:
Yes that does what you need, you get the underlying map projection expression. Though the naming is a bit off, it could be `GetProjectionExpression`. But if you're looking to get the expression for `ProjectTo`, that's the one.

buvinghausen:
Thanks @jbogard I will open a new PR with that function name here in a bit.

jbogard:
Will the existing method not work for you? Changing the method name is a breaking change. We also don't want to expose this directly on the mapper - if you're using `ProjectTo` you don't use the `IMapper` interface, that's for in-memory mapping.

buvinghausen:
@jbogard ok perhaps I would better phrase this...  ProjectionBuilder is not publicly accessible.  If I wanted to make an accessibility point for the end user plebians (myself) to run the GetMapExpression function what would be the best manner to expose that functionality?  I was just assuming IMapper because I have the IMapper already mapped into my controller I just want to get the projection expression from AutoMapper when the end user doesn't provide one themselves.

buvinghausen:
@jbogard thank you for the help and sorry for wasting your time. I found the SNAFU on my end was not including the AutoMapper.QueryableExtensions namespace and thus not seeing the extension method was available.

I now have my own extension method to cut down on the verbosity.

```csharp
public static Expression<Func<TSource, TDestination>> GetProjectionExpression<TSource, TDestination>(this IMapper mapper) =>
    mapper.ConfigurationProvider.ExpressionBuilder.GetMapExpression<TSource, TDestination>();
```

PR #3730, opened 2021-09-08
Mention that IncludeMembers doesn't do runtime mapping

hmijail:
Applying the learnings from https://stackoverflow.com/questions/63292509/map-class-with-abstract-property-to-destination/63304821?noredirect=1#answer-63304821

### DIFF
--- a/docs/Mapping-inheritance.md
+++ b/docs/Mapping-inheritance.md
@@ -65,6 +65,12 @@ Assert.IsType<OnlineOrderDto>(mapped);
 
 You will notice that because the mapped object is a OnlineOrder, AutoMapper has seen you have a more specific mapping for OnlineOrder than OrderDto, and automatically chosen that.
 
+Note that this does not apply to situations where the mapping happens at configuration time, like `IncludeMembers` does. To get the runtime-mapping behavior in that case, a solution is to use `AfterMap`:
+```c#
+CreateMap<Source, Destination>()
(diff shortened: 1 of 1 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/AutoMapper/AutoMapper?pullRequest=3730) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/AutoMapper/AutoMapper?pullRequest=3730) before we can accept your contribution.<br/><br/>:x: hmijail [sign now](https://cla.dotnetfoundation.org/AutoMapper/AutoMapper?pullRequest=3730)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/AutoMapper/AutoMapper?pullRequest=3730) it.</sub>

lbargaoanu:
Thanks, but I don't think it's worth to have that in the docs. I've added that note about the config time behaviour, that should be enough.

hmijail:
IMHO, the existing note only covers half of the problem. I tried adding this note here because this is the other half (+ suggested solution). I think this would have saved me _a lot_ of head scratching.

Relatedly, I was wondering whether it'd make sense to detect cases which are doomed to go wrong, like an `IncludeMembers` on a member whose declared type is an abstract class. Since the member will necessarily be a child class at runtime, such mapping will never work. Sounds like something that `AssertConfigurationIsValid` should report.

PR #3821, opened 2022-01-06
Assign collections we can't handle 

lbargaoanu:
For better or worse, this favours backward compatibility without giving up the performance gains.


jbogard (approved):

jbogard:
YOLO

PR #3830, opened 2022-01-10
Adding information to exception message; Addresses #3828

jbogard:


lbargaoanu:
Ideally this would throw when calling `Include`/`IncludeBase`, but yes, I'm tired of changing exception messages too :)

Issue #3869, opened 2022-01-26
v11 InvalidOperationException: Stack Empty on mapping which used to generate a subquery


Erythnul:
### Gist
https://gist.github.com/Erythnul/38a29c9012a3fdcc77fe117c51386ce9

### Summary
Hi!

I've been reworking some of our more interesting mappers to start our upgrade to 11.0.0 and have run into the problem outlined in the gist. I can open a PR with a failing unit or integration test if necessary.

Looking forward to hearing from you!

Kind regards.

### Expected behavior

In 10.1.1 this would generate an expression like

```csharp
EnumerableQuery<AEntity>.Select(
    dtoAEntity => new Object_1451902296___DtoSubWrapper#DtoSub_Id
    {
        __DtoSubWrapper#DtoSub = dtoAEntity.BEntity.CEntities.FirstOrDefault(x => x.Id == dtoAEntity.CEntityId),
        Id = dtoAEntity.Id
    }).Select(
    dtoLet => new Dto
    {
        DtoSubWrapper = new DtoSubWrapper
        {
            DtoSub = (dtoLet.__DtoSubWrapper#DtoSub == null)
                ? null
                : new DtoSub
                {
                    Id = dtoLet.__DtoSubWrapper#DtoSub.Id,
                    SubString = dtoLet.__DtoSubWrapper#DtoSub.SubString
                }
        },
        Id = dtoLet.Id
    })
```

Not sure if that should be different with the reworked pipeline in 11.0.0, but something like this?

### Actual behavior

System.InvalidOperationException: Stack Empty

Stacktrace: 
```
    Stack`1.ThrowForEmptyStack()
    Stack`1.Peek()
    <> c.< GetSourceExpression > b__4_1(Expression left, Expression right) line 269
    Enumerable.Aggregate[TSource, TAccumulate](IEnumerable`1 source, TAccumulate seed, Func`3 func)
    SubQueryPath.GetSourceExpression(Expression parameter) line 268
    <> c__DisplayClass11_0.< GetSubQueryExpression > b__0(SubQueryPath path) line 230
    SelectListIterator`2.ToArray()
    Enumerable.ToArray[TSource](IEnumerable`1 source)
    FirstPassLetPropertyMaps.GetSubQueryExpression(ProjectionBuilder builder, Expression projection, TypeMap typeMap, ProjectionRequest & request, ParameterExpression instanceParameter) line 229
[...]

PR #3874, opened 2022-01-27
Fix EF6 relationships and Id initialization

Erythnul:
Fixing up some relationships and Id Initialization for CEntities. Original code was based on in-memory unittest, which had less relationship issues. #3869


lbargaoanu:
Thanks!

PR #3881, opened 2022-01-31
Converting integration tests to EF Core

### LINKED ISSUE #3521: EF Core integration tests
Anybody interested in EF Core is welcome to add a test project similar to AutoMapper.IntegrationTests. All existing integration tests use EF6. [Here](https://github.com/AutoMapper/AutoMapper/blob/f31e398a596438a939bf19d5bb97ed218124731d/src/IntegrationTests/IncludeMembers.cs#L13) is an example. Ideally, all existing tests would run on EF Core too.

jbogard:
One step towards allowing xplat/*nix-based development

Fixes #3521


jbogard:
Both `CreateDatabaseIfNotExists` and `DropCreateDatabaseAlways` don't exist in EF Core so I created ones that do similar - drop and create. Technically `CreateDatabaseIfNotExists` also drops the database but I think all tests should drop and create anyway.

src/IntegrationTests/BuiltInTypes/ByteArray.cs
```diff
@@ -1,75 +1,76 @@
...
+using System.Threading.Tasks;
+using AutoMapper.UnitTests;
+using Microsoft.EntityFrameworkCore;
 using Shouldly;
 using Xunit;
 
-namespace AutoMapper.IntegrationTests
+namespace AutoMapper.IntegrationTests.BuiltInTypes;
+
+public class ByteArrayColumns : AutoMapperSpecBase, IAsyncLifetime
```
lbargaoanu, line 11: Maybe have a base class `IntegrationTest<TDbInitializer> : IAsyncLifetime`?

jbogard: Yeah that can work. Though I'm not sure I want to have even more inheritance in our tests classes. The whole "Establish_Context" "Because_of" should probably all go away.

lbargaoanu: Yeah, those should go  :)

jbogard: In order to target the linux build servers, I'll have to also move to SQlite from LocalDB, so I was gonna wait for any additional refactorings to see what that looks like.

lbargaoanu: Some SQL things don't work on SQLite, so I don't know about that, probably just run what does work.

jbogard:
@lbargaoanu can you review this? I can click the red button but would rather not :)

lbargaoanu (approved):
I still think we need to refactor to a base class, but we can do that anytime.

PR #3938, opened 2022-03-11
allow building on Linux (and probably macOS)

adamralph:
I wanted to test the next major of MinVer on a few repos which consume it, but I was stuck with AutoMapper because it currently only build on Windows.

### DIFF
--- a/src/UnitTests/Internationalization.cs
+++ b/src/UnitTests/Internationalization.cs
@@ -16,12 +16,12 @@ namespace AutoMapper.UnitTests
 
             public class Customer
             {
-                public string Æøå { get; set; }
+                public string Ã†Ã¸Ã¥ { get; set; }
             }
 
             public class OrderDto
             {
-                public string CustomerÆøå { get; set; }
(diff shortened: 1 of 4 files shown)


src/UnitTests/Internationalization.cs
```diff
@@ -16,12 +16,12 @@ public class Order
 
             public class Customer
             {
-                public string ��� { get; set; }
+                public string Æøå { get; set; }
```
adamralph, line 19: The encoding was wrong on these characters. They were encoded in "Western European (Windows)", but the file is encoded in UTF-8. I can only guess that on Windows, some kind of encoding munging is done to allow it pass through the compiler, but on Linux no such "convenience" is present, so compilation was failing.

adamralph:
Sorry, I force pushed a couple of corrections, but it's all good to go now.

jbogard:
Does the build actually run locally (build.ps1)?

jbogard (approved):

PR #3942, opened 2022-03-17
Use RegexOptions.Compiled

EngRajabi:
improvment regex with compile flag


src/IntegrationTests/ExplicitExpansion/ProjectionWithExplicitExpansion.cs
```diff
@@ -19,7 +19,7 @@ public static class Ext
     public static void SqlShouldNotSelectColumn(this string sqlSelect, string columnName)=> sqlSelect.ShouldNotContain($"[{columnName}]");
     public static void SqlFromShouldStartWith  (this string sqlSelect, string tableName)
     {
-        Regex regex = new Regex($@"FROM(\s+)\[{tableName}\](\s+)AS");
+        Regex regex = new Regex($@"FROM(\s+)\[{tableName}\](\s+)AS", RegexOptions.Compiled);
```
lbargaoanu, line 22: Revert all the test changes, I don't see the point.

lbargaoanu:
Did you measure the impact? What does it add to the config time. What are the gains when mapping, when the regex actually does smth and when it doesn't.

EngRajabi:
Yes, I measured it.

https://medium.com/@mohsen_rajabi/how-to-write-a-regex-very-fast-in-c-best-practice-875d386c0485

lbargaoanu:
That's generic stuff, I'm talking about AM here. I've spent some time profiling AM config and I never saw time spent doing Regex stuff. Probably because it didn't apply and it usually doesn't. So what kind of config it takes to justify this change. How many uses it takes to compensate for the time spent compiling? When the regeex is not used, it's still compiled? This is not a problem now from what I've seen, so I would like to see a config where it is a problem.

lbargaoanu:
I've done some tests, and I'm reasonably confident that for my use case this is a regression. But I'm sure it depends :)
You can change the naming convention from the outside, so I guess the real question is what's the right default. When the setting starts paying off. I don't have the time for a deep analysis, but it seems it takes a lot of maps for that to happen. And probably it matters whether the regex matches or not. So we probably have the right default and people can profile their configs and make a decision accordingly. Note that, if you care about performance, you should probably [disable](https://gist.github.com/lbargaoanu/9948bf66d452ba6b816252f9965143ee) the features you don't need.
If you want to take a deeper look, you're welcome to do so and we can discuss your findings.

EngRajabi:
Thanks for your review.
I wanted to help the program run faster by adding the Compile flag

PR #3949, opened 2022-03-27
Add support for projecting to custom collections.

fiseni:
See discussion [here](https://github.com/AutoMapper/AutoMapper/discussions/3893).

The existing behavior is preserved. In the current implementation of [EnumerableProjectionMapper](https://github.com/AutoMapper/AutoMapper/blob/master/src/AutoMapper/QueryableExtensions/ProjectionMappers/EnumerableProjectionMapper.cs), it is assumed that destination collection type can be assigned from `List`, hence `ToList` extension method is utilized. In the scope of this PR, we explicitly check if that's the case. If not, we give it one more chance by trying to use a suitable constructor.

[lbargaoanu](https://github.com/lbargaoanu), sorry I didn't understand exactly your suggestion. If we're doing that check on top of existing code, wouldn't we have unnecessary overhead for the standard usage? Let me know what you want me to change here.

### DIFF
--- a/src/AutoMapper/QueryableExtensions/ProjectionMappers/EnumerableProjectionMapper.cs
+++ b/src/AutoMapper/QueryableExtensions/ProjectionMappers/EnumerableProjectionMapper.cs
@@ -1,3 +1,4 @@
+using System.Collections.Generic;
 using System.ComponentModel;
 using System.Linq;
 using System.Linq.Expressions;
@@ -34,17 +35,19 @@ namespace AutoMapper.QueryableExtensions.Impl
             }
             if (!memberMap.DestinationType.IsAssignableFrom(sourceExpression.Type))
             {
-                if (memberMap.DestinationType.IsArray)
+                var convertFunction = memberMap.DestinationType.IsArray ? ToArrayMethod : ToListMethod;
+                convertFunction = convertFunction.MakeGenericMethod(destinationListType);
+                if (memberMap.DestinationType.IsAssignableFrom(convertFunction.ReturnType))
                 {
-                    sourceExpression = Call(ToArrayMethod.MakeGenericMethod(destinationListType), sourceExpression);
+                    sourceExpression = Call(convertFunction, sourceExpression);
                 }
-                else if (memberMap.IsDestinationTypeAssignableFromList)
+                else
                 {
-                    sourceExpression = Call(ToListMethod.MakeGenericMethod(destinationListType), sourceExpression);
-                }
-                else if (memberMap.DestinationTypeIEnumerableCtor is not null)
-                {
-                    sourceExpression = New(memberMap.DestinationTypeIEnumerableCtor, sourceExpression);
+                    var ctorInfo = memberMap.DestinationType.GetConstructor(new[] { typeof(IEnumerable<>)?.MakeGenericType(destinationListType) });
+                    if (ctorInfo is not null)
+                    {
+                        sourceExpression = New(ctorInfo, sourceExpression);
+                    }
                 }
             }
             return sourceExpression;
--- a/src/AutoMapper/QueryableExtensions/ProjectionMappers/EnumerableProjectionMapper.cs
+++ b/src/AutoMapper/QueryableExtensions/ProjectionMappers/EnumerableProjectionMapper.cs
@@ -43,7 +43,7 @@ namespace AutoMapper.QueryableExtensions.Impl
                 }
                 else
                 {
-                    var ctorInfo = memberMap.DestinationType.GetConstructor(new[] { typeof(IEnumerable<>)?.MakeGenericType(destinationListType) });
+                    var ctorInfo = memberMap.DestinationType.GetConstructor(new[] { sourceExpression.Type });
                     if (ctorInfo is not null)
                     {
                         sourceExpression = New(ctorInfo, sourceExpression);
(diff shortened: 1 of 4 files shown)


src/AutoMapper/QueryableExtensions/ProjectionMappers/EnumerableProjectionMapper.cs
```diff
@@ -34,8 +34,18 @@ public Expression Project(IGlobalConfiguration configuration, MemberMap memberMa
             }
             if (!memberMap.DestinationType.IsAssignableFrom(sourceExpression.Type))
             {
-                var convertFunction = memberMap.DestinationType.IsArray ? ToArrayMethod : ToListMethod;
-                sourceExpression = Call(convertFunction.MakeGenericMethod(destinationListType), sourceExpression);
```
lbargaoanu, line 38:
```C#
converFunction = convertFunction.MakeGenericMethod(destinationListType);
if(memberMap.DestinationType.IsAssignableFrom(convertFuncion.ReturnType))
{
    sourceExpression = Call(convertFunction.MakeGenericMethod(destinationListType), sourceExpression);
}
else
{
     // new code
}
```
Let's revert the caching in the `PropertyMap` and have everything here. Even if it were useful, which I don't think iit is, this is a corner case.

fiseni: Sure, updated.

src/AutoMapper/QueryableExtensions/ProjectionMappers/EnumerableProjectionMapper.cs
```diff
@@ -35,7 +36,19 @@ public Expression Project(IGlobalConfiguration configuration, MemberMap memberMa
...
                 var convertFunction = memberMap.DestinationType.IsArray ? ToArrayMethod : ToListMethod;
-                sourceExpression = Call(convertFunction.MakeGenericMethod(destinationListType), sourceExpression);
+                convertFunction = convertFunction.MakeGenericMethod(destinationListType);
+                if (memberMap.DestinationType.IsAssignableFrom(convertFunction.ReturnType))
+                {
+                    sourceExpression = Call(convertFunction, sourceExpression);
+                }
+                else
+                {
+                    var ctorInfo = memberMap.DestinationType.GetConstructor(new[] { typeof(IEnumerable<>)?.MakeGenericType(destinationListType) });
```
lbargaoanu, line 46:
```suggestion
                    var ctorInfo = memberMap.DestinationType.GetConstructor(new[] { sourceExpression.Type });
```

src/AutoMapper/QueryableExtensions/ProjectionMappers/EnumerableProjectionMapper.cs
```diff
@@ -35,7 +36,19 @@ public Expression Project(IGlobalConfiguration configuration, MemberMap memberMa
...
+                if (memberMap.DestinationType.IsAssignableFrom(convertFunction.ReturnType))
+                {
+                    sourceExpression = Call(convertFunction, sourceExpression);
+                }
+                else
+                {
+                    var ctorInfo = memberMap.DestinationType.GetConstructor(new[] { sourceExpression.Type });
+                    if (ctorInfo is not null)
+                    {
+                        sourceExpression = New(ctorInfo, sourceExpression);
```
lbargaoanu, line 49: This probably needs a cast because the returned constructor parameter might not match exactly, but I'm not sure how likely is that to happen, so perhaps it's better to stick with the simpler code until proven otherwise.

lbargaoanu:
Thanks!

src/AutoMapper/QueryableExtensions/ProjectionMappers/EnumerableProjectionMapper.cs
fiseni: I think the cast will mostly fail, since `sourceExpression.Type` will be `IEnumerable<>` in the majority of the cases (product of Select statement). If we want to be more "loose" and support more specific ctor parameters, then perhaps the following will cover more cases?
```c#
var convertFunction = memberMap.DestinationType.IsArray ? ToArrayMethod : ToListMethod;
sourceExpression = Call(convertFunction.MakeGenericMethod(destinationListType), sourceExpression);
if (!memberMap.DestinationType.IsAssignableFrom(convertFunction.ReturnType))
{
    var ctorInfo = memberMap.DestinationType.GetConstructor(new[] { sourceExpression.Type });
    if (ctorInfo is not null)
    {
        sourceExpression = New(ctorInfo, sourceExpression);
    }
}
```

lbargaoanu: I don't follow :) All I'm saying is that the constructor parameter can be a base type of `sourceExpression.Type` and then the `New` expression constructor will fail, but I'd rather see it fail before adding that ugly piece of code.

fiseni: Being the base type of it would be really rare. I meant the cases where ctor param is a derived type of `sourceExpression.Type`. The above code will have a better chance of success for those cases. The generated statement will be something like `new(__.ToList())`. Anyhow, not important :) Thanks for your help.
```c#
public class CustomList<T> : List<T>
{
    public CustomList(IList<T> collection) : base(collection)
    {
    }
}
```

Issue #3988, opened 2022-05-30
[dotnet-sdk-7.0.100-preview.5.22257.3] MaxInteger[T](System.Collections.Generic.IEnumerable`1[T])' violates the constraint of type parameter 'T' exception


Junjun-zhao:
**Description**: Apps that refer to the AutoMapper package failed to launch against .NET 7 dotnet-sdk-7.0.100-preview.5.22257.3 runtime with exception. 

Please refer to more details information from  https://github.com/dotnet/runtime/issues/69119

**Application Name**: nopCommerceNetCore, blazorboilerplate, CastIt, grandnode, BlogSimpleAPI, NorthwindTraders, DevBetterWeb, piranha.core, Boilerplate
**OS**: Windows 10 RS5
**CPU**: X64
**.NET Build Number**: dotnet-sdk-7.0.100-preview.5.22257.3

**Findings:**

1. The apps are thrown exception on new methods, like MaxInteger(), MaxtFloat(), which are not exist in previous version: 
![finding](https://user-images.githubusercontent.com/83067197/167610180-9b7fdee8-f55d-4c64-b51b-0df525742da7.png)
2. The exception is thrown from MethodInfo.MakeGeneric method.
3. All these apps are using AutoMapper , and when debuging with AutoMapper , we get below exception:

```
System.ArgumentException
  HResult=0x80070057
  Message=GenericArguments[0], 'Nop.Web.Framework.Models.BaseNopModel', on 'T MaxInteger[T](System.Collections.Generic.IEnumerable`1[T])' violates the constraint of type 'T'.
  Source=System.Private.CoreLib
  StackTrace:
   at System.RuntimeType.ValidateGenericArguments(MemberInfo definition, RuntimeType[] genericArguments, Exception e)
   at System.Reflection.RuntimeMethodInfo.MakeGenericMethod(Type[] methodInstantiation)
   at AutoMapper.TypeDetails.<>c__DisplayClass28_1.<BuildPublicNoArgExtensionMethods>b__10(MethodInfo extensionMethod) in F:\AutoMapper-10.1.1\AutoMapper-10.1.1\src\AutoMapper\TypeDetails.cs:line 161
   at System.Linq.Enumerable.WhereSelectArrayIterator`2.MoveNext()
   at System.Linq.Enumerable.ConcatIterator`1.MoveNext()
   at System.Linq.Enumerable.<SelectManyIterator>d__227`3.MoveNext()
[...]

PR #3999, opened 2022-06-08
Suppress generic constraint exceptions in GetPublicNoArgExtensionMethods

stephentoub:
GetPublicNoArgExtensionMethods tries to instantiate arbitrary generic methods with arbitrary type arguments, which can fail if those arguments don't meet any constraints defined on the generic parameters. Automapper isn't doing the required checking for constraint validation today, resulting in exceptions escaping; multiple existing Automapper unit tests fail on .NET 7 as a result.  This works around it by eating the ArgumentException that comes from MakeGenericMethod (there's no exposed compatibility-checking function, and so while not usually recommended, the most reliable thing here is to simply catch the errors stemming from the validation MakeGenericMethod performs).

Fixes https://github.com/AutoMapper/AutoMapper/issues/3988

### DIFF
--- a/src/AutoMapper/Internal/TypeDetails.cs
+++ b/src/AutoMapper/Internal/TypeDetails.cs
@@ -123,12 +123,29 @@ namespace AutoMapper.Internal
                     from extensionMethod in sourceExtensionMethodSearch
                     where extensionMethod.IsGenericMethodDefinition
                         && extensionMethod.GetGenericArguments().Length == genericInterfaceArguments.Length
-                    select extensionMethod.MakeGenericMethod(genericInterfaceArguments)
+                    let constructedGeneric = MakeGenericMethod(extensionMethod, genericInterfaceArguments)
+                    where constructedGeneric != null
+                    select constructedGeneric
                 )
                 from methodMatch in matchedMethods
                 where methodMatch.GetParameters()[0].ParameterType.IsAssignableFrom(genericInterface)
                 select methodMatch
             );
+
+            // Use method.MakeGenericMethod(genericArguments) wrapped in a try/catch(ArgumentException)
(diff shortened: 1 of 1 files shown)


jbogard:
Related feature request to not have to do this exception silliness:

https://github.com/dotnet/runtime/issues/28033

le sigh

jbogard (approved):

lbargaoanu:
Maybe we should simply skip all methods with generic constraints.

PriyaPurkayastha:
@jbogard @lbargaoanu we (.NET team) would like to collaborate with you on this issue. We recognize that AutoMapper library is used by many applications, and it is important for us that these applications continue to work with .NET 7. Can you please contact us on dotnet@microsoft.com? Thanks!

lbargaoanu:
I'm on vacation for the next two weeks :)
@jbogard If someone is interested in speeding up AM configuration, the first thing to do is to disable the [expensive stuff](https://gist.github.com/lbargaoanu/9948bf66d452ba6b816252f9965143ee). That's the configuration I've profiled until now. But it's on by default. So I've tried to minimize the damage (in my branch). Mainly by calling `MakeGenericMethod` only when the extension method actually matches a destination name.

PriyaPurkayastha:
Thanks @lbargaoanu. We wanted to get some information as follows:
1. When is the planned release date for AutoMapper 11.0.2 which contains this fix that Stephen submitted?
2. We found that several apps in our AppCompat lab use older versions of AutoMapper. Can you please let us know what your Servicing plans look like? Do you normally service previous AutoMapper major/minor versions, or would your customers need to pick up the latest version when they use .NET 7? If you do service previous releases, what can we do to get the fix in those releases as well?

.NET 7 GA is this November, so we have a couple of weeks to evaluate the options. Have a great vacation! We look forward to hearing back from you.

PR #4005, opened 2022-06-21
Support multiple .NET Frameworks (add net6.0)

bdovaz:
Right now it only targets netstandard2.1 and makes it dependent on: Microsoft.CSharp (>= 4.7.0).

That dependency is very old (2019): https://www.nuget.org/packages/Microsoft.CSharp/#versions-body-tab

As of .net5.0+ it is no longer needed so it can be removed.

In our case this dependency is a problem because it has dependencies of packages of System.* 4.3 that have **vulnerabilities** that obviously we do not want.

### DIFF
--- a/src/AutoMapper/AutoMapper.csproj
+++ b/src/AutoMapper/AutoMapper.csproj
@@ -3,7 +3,7 @@
   <PropertyGroup>
     <Summary>A convention-based object-object mapper.</Summary>
     <Description>A convention-based object-object mapper.</Description>
-    <TargetFrameworks>netstandard2.1</TargetFrameworks>
+    <TargetFrameworks>net6.0;netstandard2.1</TargetFrameworks>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <AssemblyName>AutoMapper</AssemblyName>
     <AssemblyOriginatorKeyFile>..\..\AutoMapper.snk</AssemblyOriginatorKeyFile>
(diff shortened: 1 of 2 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/AutoMapper/AutoMapper?pullRequest=4005) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/AutoMapper/AutoMapper?pullRequest=4005) before we can accept your contribution.<br/><br/>:x: bdovaz [sign now](https://cla.dotnetfoundation.org/AutoMapper/AutoMapper?pullRequest=4005)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/AutoMapper/AutoMapper?pullRequest=4005) it.</sub>

jbogard:
What vulnerabilities exactly? And what runtime are you targeting?

bdovaz:
CVE-2018-8292 (System.Net.Http):

Official announcement: https://github.com/dotnet/announcements/issues/88

CVE-2019-0820 (System.Text.RegularExpressions):

Official announcement: https://github.com/dotnet/announcements/issues/111

@jbogard The solution they propose in case you have a direct dependency is to update it to the minimum version that fixes the vulnerability but in this case they are transitive dependencies through the **Microsoft.CSharp** dependency so it is that dependency who determines it (and there are no new versions, from .NET 5, it is not necessary).

jbogard:
Are those two packages pulled in from the `Microsoft.CSharp` dependency? I'm not seeing that in my test projects.

bdovaz:
AutoMapper > Microsoft.CSharp (>= 4.7.0) > NETStandard.Library (>= 1.6.1):

https://www.nuget.org/packages/NETStandard.Library/1.6.1#dependencies-body-tab

System.Net.Http (>= 4.3.0)
System.Text.RegularExpressions (>= 4.3.0)

bdovaz:
I don't know why it happens in some cases but regardless of that, so much harm does it do to do multi targeting including .NET 6 which on top of that is LTS?

jbogard:
I don't want to accept a PR for a bug where the fix does not actually address the supposed bug. I can't reproduce this, in any runtime I have that supports `netstandard2.1`. .NET Core 3.1 runtime will load the dependency based on the runtime version, NOT the transitive version. Is this behavior you're seeing in production, or is it part of some automated vulnarability tool? Reports from Twitter on this indicate it's a problem with the tool, and they mark it as a false positive.

bdovaz:
I still insist, forget about vulnerability.

What's the problem with accepting this PR? By targeting .NET 6.0 we simplify the dependencies and remove one that is unnecessary.

Does it hurt that much? Nowadays it is very common for nuget packages to target .net5.0 or .net6.0 precisely for that reason, to avoid bringing redundant or old dependencies that only exist in .netstandard2.0.

bdovaz:
With this attitude you are going to get the community to not help.

You have focused on vulnerabilities when I am saying what it helps to reduce the complexity of dependencies to a project that targets .net6.0.

jbogard:
We multitargeted for years but I don't want to anymore. It slows down compilation and makes testing more complex. If you can point to a specific issue you are running into, I can look at this but so far I'm not seeing any problems pointed out.

lbargaoanu:
@jbogard Support for .net core 3.1 ends this year. So next year we can consider targeting only .net 6.

PR #4011, opened 2022-06-29
Consider member value resolvers and value converters for source valid…

lbargaoanu:
…ation


src/AutoMapper/MemberMap.cs
```diff
@@ -39,7 +39,20 @@ public class MemberMap
         public virtual ValueResolverConfiguration ValueResolverConfig { get => default; set { } }
         public virtual ValueResolverConfiguration ValueConverterConfig { get => default; set { } }
         public virtual IReadOnlyCollection<ValueTransformerConfiguration> ValueTransformers => Array.Empty<ValueTransformerConfiguration>();
-        public MemberInfo SourceMember => CustomMapExpression.GetMember() ?? SourceMembers.LastOrDefault();
+        public MemberInfo SourceMember => this switch
+        {
```
lbargaoanu, line 43: I think it would make a lot of sense to have only one resolver object per member map (we have 5 kinds of resolvers). It would break cases when people use a different configuration for `Map` and `ProjectTo`, but I guess that's a small price to pay :)

src/AutoMapper/ResolutionContext.cs
```diff
@@ -10,33 +10,36 @@ public class ResolutionContext : IInternalRuntimeMapper
...
         /// Context items from <see cref="Options"/>
         /// </summary>
         public IDictionary<string, object> Items
         {
             get
             {
-                CheckDefault();
+                if (Options == null)
+                {
+                    ThrowInvalidMap();
```
lbargaoanu, line 33: We could simply return null here but I'm guessing people would keep asking why `Items` is null.

jbogard (approved):

jbogard:
Ready to go?

lbargaoanu:
Yes, I think so.

PR #4105, opened 2022-10-14
Emphasized the ProjectTo method call location

jbogard:


docs/Queryable-Extensions.md
```diff
@@ -56,9 +56,9 @@ public List<OrderLineDTO> GetLinesForOrder(int orderId)
 ```
 
 The `.ProjectTo<OrderLineDTO>()` will tell AutoMapper's mapping engine to emit a `select` clause to the IQueryable that will inform entity framework that it only needs to query the Name column of the Item table, same as if you manually projected your `IQueryable` to an `OrderLineDTO` with a `Select` clause.
-
-`ProjectTo` must be the last call in the chain. ORMs work with entities, not DTOs. So apply any filtering and sorting on entities and, as the last step, project to DTOs.
-
+
+**`ProjectTo` must be the last call in the LINQ method chain. ORMs work with entities, not DTOs. Apply any filtering and sorting on entities and, as the last step, project to DTOs. Query providers are highly complex and making the `ProjectTo` call last ensures the query provider works as closely as designed to build valid queries against the underlying query target (SQL, Mongo QL etc.).**
```
lbargaoanu, line 60: Maybe drop the bold thing and just make a separate paragraph with a title.

lbargaoanu (approved):

PR #4202, opened 2023-01-17
Target .NET 6

lbargaoanu:


src/AutoMapper/Mapper.cs
```diff
@@ -28,15 +28,15 @@ public interface IMapperBase
...
+    TDestination? Map<TSource, TDestination>(TSource? source, TDestination? destination);
     /// <summary>
     /// Execute a mapping from the source object to a new destination object with explicit <see cref="System.Type"/> objects
     /// </summary>
     /// <param name="source">Source object to map from</param>
     /// <param name="sourceType">Source type to use</param>
     /// <param name="destinationType">Destination type to create</param>
     /// <returns>Mapped destination object</returns>
-    object Map(object source, Type sourceType, Type destinationType);
+    object? Map(object? source, Type? sourceType, Type destinationType);
```
lbargaoanu, line 39: Technically `destinationType` can be null here, it would fallback to object. But null doesn't really makes sense, so I want to say this overload is different.

src/AutoMapper/Mapper.cs
```diff
@@ -84,7 +84,7 @@ public interface IMapper : IMapperBase
     /// <param name="destinationType">Destination type to create</param>
     /// <param name="opts">Mapping options</param>
     /// <returns>Mapped destination object</returns>
-    object Map(object source, Type sourceType, Type destinationType, Action<IObjectMappingOperationOptions> opts);
+    object? Map(object? source, Type? sourceType, Type destinationType, Action<IObjectMappingOperationOptions> opts);
```
lbargaoanu, line 87: Technically `destinationType` can be null here, it would fallback to object. But null doesn't really makes sense, so I want to say this overload is different.

BlaiseD:
Does this leave Xamarin.Forms developers behind? MAUI may not be 100% - sounds early to me.

lbargaoanu:
Where does it say what Xamarin.Forms targets?

BlaiseD:
Maybe I'm wrong - I looked [here](https://learn.microsoft.com/en-us/answers/questions/636510/xamarin-and-net-6).

Sound like my [ignorance](https://github.com/xamarin/xamarin-android/wiki/Migrating-Xamarin.Android-Applications-to-.NET-6).

lbargaoanu:
Anyway, this is about the future :) We'd probably release AM 13 about the same time as .NET 8.

docs/Queryable-Extensions.md
```diff
@@ -208,7 +208,15 @@ You may also use a dictionary to build the projection values:
...
 
-However, using a dictionary will result in hard-coded values in the query instead of a parameterized query, so use with caution.
+However, using a dictionary will result in hard-coded values in the query instead of a parameterized query, so use with caution.
+
+### Recursive queries
+
+Ideally, you would avoid queries that call themselves (do some research). But if you must, you need to enable them:
+
+```c#
+configuration.Internal().RecursiveQueriesMaxDepth = someRandomNumber;
```
lbargaoanu, line 218: The default is zero. We could choose a number :)

jbogard: One? Three? Five? Whatever we pick people will complain.

jbogard (approved):

docs/Queryable-Extensions.md
lbargaoanu: Word :)

PR #4266, opened 2023-03-29
Replace Expression.Default with Expression.Constant

statler:
This is a change to resolve the issues in https://github.com/AutoMapper/AutoMapper/discussions/4265

Essentially it replaces Expression.Default with a hopefully equivalent Expression.Constant.

Seems to work. Passes tests.


lbargaoanu:
We don't support EF6 anymore. You can do that in your own fork.

statler:
> We don't support EF6 anymore. You can do that in your own fork.

Where is this specified? **The docs say you don't support .Net Framework.** That is a completely different thing. I am running EF6 in .Net 7, not framework. I could not find it anywhere that EF6 is not supported. The above changes are all that is required to make it work in EF6, so why not?

PR #4272, opened 2023-04-03
added Polimorphic Projections

JanEggers:
I consider this a draft that does what I need.

If the design is approved I will also create some tests.

image

image

new queries contain the expected subtypes.

fixes #4271

### DIFF
--- a/src/AutoMapper/QueryableExtensions/ProjectionBuilder.cs
+++ b/src/AutoMapper/QueryableExtensions/ProjectionBuilder.cs
@@ -63,7 +63,6 @@ public class ProjectionBuilder : IProjectionBuilder
         {
             return customProjection;
         }
-        var propertiesProjections = new List<MemberBinding>();
         int depth;
         if (OverMaxDepth())
         {
@@ -72,23 +71,35 @@ public class ProjectionBuilder : IProjectionBuilder
                 return null;
             }
         }
-        else
-        {
-            ProjectProperties();
-        }
+
         var constructorExpression = CreateDestination();
-        var expression = MemberInit(constructorExpression, propertiesProjections);
+        var propertyBindings = ProjectProperties(typeMap, false);
+        var expression = (Expression)MemberInit(constructorExpression, propertyBindings);
+        foreach (var derivedType in _configuration.GetIncludedTypeMaps(typeMap.IncludedDerivedTypes))
+        {
+            var derivedContructorExpression = New(derivedType.DestinationType);
+            var derivedExpression = MemberInit(derivedContructorExpression, propertyBindings.Concat(ProjectProperties(derivedType, true)));
+            var condition = TypeIs(instanceParameter, derivedType.SourceType);
+
+            expression = Condition(condition, Convert(derivedExpression, typeMap.DestinationType), expression);
+        }
         return expression;
         bool OverMaxDepth()
         {
             depth = letPropertyMaps.IncrementDepth(request);
             return typeMap.MaxDepth > 0 && depth >= typeMap.MaxDepth;
         }
-        void ProjectProperties()
+        List<MemberBinding> ProjectProperties(TypeMap localTypeMap, bool polimorph)
         {
-            foreach (var propertyMap in typeMap.PropertyMaps)
+            var propertiesProjections = new List<MemberBinding>();
+            foreach (var propertyMap in localTypeMap.PropertyMaps)
             {
                 if (!propertyMap.CanResolveValue || !propertyMap.CanBeSet || typeMap.ConstructorParameterMatches(propertyMap.DestinationName))
+                {
+                    continue;
+                }
+
+                if (polimorph && propertyMap.DestinationMember.DeclaringType != localTypeMap.DestinationType)
                 {
                     continue;
                 }
@@ -98,6 +109,8 @@ public class ProjectionBuilder : IProjectionBuilder
                     propertiesProjections.Add(Bind(propertyMap.DestinationMember, propertyProjection));
                 }
             }
+
+            return propertiesProjections;
         }
         Expression TryProjectMember(MemberMap memberMap, Expression defaultSource = null)
         {
@@ -140,7 +153,7 @@ public class ProjectionBuilder : IProjectionBuilder
                     var resolvedSource = memberMap switch
                     {
                         { CustomMapExpression: LambdaExpression mapFrom } => MapFromExpression(mapFrom),
-                        { SourceMembers.Length: > 0  } => memberMap.ChainSourceMembers(CheckCustomSource()),
+                        { SourceMembers.Length: > 0  } => memberMap.ChainSourceMembers(CheckCustomSource(memberMap)),
                         _ => defaultSource ?? throw CannotMap(memberMap, request.SourceType)
                     };
                     if (NullSubstitute())
@@ -164,13 +177,23 @@ public class ProjectionBuilder : IProjectionBuilder
                         return letPropertyMaps.GetSubQueryMarker(newMapFrom);
                     }
                     bool NullSubstitute() => memberMap.NullSubstitute != null && resolvedSource is MemberExpression && (resolvedSource.Type.IsNullableType() || resolvedSource.Type == typeof(string));
-                    Expression CheckCustomSource()
+                    Expression CheckCustomSource(MemberMap member = null)
                     {
                         if (customSource == null)
                         {
+                            if (member != null && member.SourceMember.DeclaringType != instanceParameter.Type)
+                            {
+                                return Convert(instanceParameter, member.TypeMap.SourceType);
+                            }
+
                             return instanceParameter;
                         }
-                        return customSource.IsMemberPath(out _) ? customSource.ReplaceParameters(instanceParameter) : letPropertyMaps.GetSubQueryMarker(customSource);
+                        if (customSource.IsMemberPath(out _)) 
+                        {
+                            return customSource.ReplaceParameters(instanceParameter);
+                        } 
+                        
+                        return letPropertyMaps.GetSubQueryMarker(customSource);
                     }
                 }
                 IProjectionMapper GetProjectionMapper()
@@ -225,12 +248,21 @@ public class ProjectionBuilder : IProjectionBuilder
         public override LetPropertyMaps New() => new FirstPassLetPropertyMaps(Configuration, GetCurrentPath(), BuiltProjections);
         public override QueryExpressions GetSubQueryExpression(ProjectionBuilder builder, Expression projection, TypeMap typeMap, in ProjectionRequest request, ParameterExpression instanceParameter)
         {
+            var sParam = Parameter(typeMap.SourceType, "s");
+
             var letMapInfos = _savedPaths.Select(path =>
                 (path.LetExpression,
                 MapFromSource : path.GetSourceExpression(instanceParameter),
                 Property : path.GetPropertyDescription(),
-                path.Marker)).ToArray();
-            var properties = letMapInfos.Select(m => m.Property).Concat(GePropertiesVisitor.Retrieve(projection, instanceParameter));
+                path.Marker)).Append(new
+            (
+                Lambda(sParam, sParam),
+                instanceParameter,
+                new PropertyDescription("__src", typeMap.SourceType),
+                instanceParameter
+            )).ToArray();
+
+            var properties = letMapInfos.Select(m => m.Property);
             var letType = ProxyGenerator.GetSimilarType(typeof(object), properties);
             TypeMap letTypeMap;
             lock(Configuration)
@@ -250,7 +282,6 @@ public class ProjectionBuilder : IProjectionBuilder
                     letPropertyMap.MapFrom(Lambda(letMapInfo.LetExpression.ReplaceParameters(letMapInfo.MapFromSource), secondParameter));
                     projection = projection.Replace(letMapInfo.Marker, Property(secondParameter, letProperty));
                 }
-                projection = new ReplaceMemberAccessesVisitor(instanceParameter, secondParameter).Visit(projection);
             }
         }
         readonly record struct SubQueryPath(MemberProjection[] Members, LambdaExpression LetExpression, Expression Marker)
@@ -269,43 +300,6 @@ public class ProjectionBuilder : IProjectionBuilder
             public PropertyDescription GetPropertyDescription() => new("__" + string.Join("#", Members.Select(p => p.MemberMap.DestinationName)), LetExpression.Body.Type);
             internal bool IsEquivalentTo(SubQueryPath other) => LetExpression == other.LetExpression && Members.Length == other.Members.Length &&
                 Members.Take(Members.Length - 1).Zip(other.Members, (left, right) => left.MemberMap == right.MemberMap).All(item => item);
-        }
-        class GePropertiesVisitor : ExpressionVisitor
-        {
-            private readonly Expression _target;
-            public List<MemberInfo> Members { get; } = new();
-            public GePropertiesVisitor(Expression target) => _target = target;
-            protected override Expression VisitMember(MemberExpression node)
-            {
-                if(node.Expression == _target)
-                {
-                    Members.TryAdd(node.Member);
-                }
-                return base.VisitMember(node);
-            }
-            public static IEnumerable<PropertyDescription> Retrieve(Expression expression, Expression target)
-            {
-                var visitor = new GePropertiesVisitor(target);
-                visitor.Visit(expression);
-                return visitor.Members.Select(member => new PropertyDescription(member.Name, member.GetMemberType()));
-            }
-        }
-        class ReplaceMemberAccessesVisitor : ExpressionVisitor
-        {
-            private readonly Expression _oldObject, _newObject;
-            public ReplaceMemberAccessesVisitor(Expression oldObject, Expression newObject)
-            {
-                _oldObject = oldObject;
-                _newObject = newObject;
-            }
-            protected override Expression VisitMember(MemberExpression node)
-            {
-                if(node.Expression != _oldObject)
-                {
-                    return base.VisitMember(node);
-                }
-                return PropertyOrField(_newObject, node.Member.Name);
-            }
         }
     }
 }
--- a/src/AutoMapper/QueryableExtensions/ProjectionBuilder.cs
+++ b/src/AutoMapper/QueryableExtensions/ProjectionBuilder.cs
@@ -272,7 +272,7 @@ public class ProjectionBuilder : IProjectionBuilder
         {
             var sParam = Parameter(typeMap.SourceType, "s");
 
-            var letMapInfos = _savedPaths.Select(path =>
+            var letMapInfos = _savedPaths.Select(path => new LetMapInfo
                 (path.LetExpression,
                 MapFromSource : path.GetSourceExpression(instanceParameter),
                 Property : path.GetPropertyDescription(),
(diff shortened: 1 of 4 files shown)


JanEggers:
image

here is also the generated expression tree typically the lambda would only contain the last else path that creates the base class

lbargaoanu:
We have [this](https://github.com/AutoMapper/AutoMapper/blob/73ba4003e49e37e5fffec49b67e7d9ea1cbb8a8a/src/UnitTests/Projection/ProjectionMappers.cs) extension point. I suggest that you try to implement it outside AM while we discuss things.

lbargaoanu:
Moved from the discussion
@lbargaoanu tried your suggestion in the pr. it may be possible but not viable. I would have to replace calling into ProjectionBuilder.CreateProjection to replace CreateDestination to be able to change the new expression. that way i have to effectivly copy all the code in Projectionbuilder. 

maybe you could improve the design so GetProjectionMapper is also called by EnumerableProjectionMapper. right now EnumerableProjectionMapper does not call GetProjectionMapper making it hard to create a custom mapper for elements inside a list.

lbargaoanu:
Did you try calling `CreateProjection` for each of the included types?

JanEggers:
yes. but i only get 3 calls for IsMatch Parent.Id, Parent.Children, Child.Id.

but i dont get a call for Parent.Children ElementType aka ChildModel  => ChildDto

lbargaoanu:
Whoever handles that, if you're before them in the list, you can take over.

JanEggers:
tried again. its 60 lines of code and im still missing the equivalent of ProjectProperties.

I think this should be in the box

https://github.com/JanEggers/GraphqlRepro/blob/2e32c3a29003c8eada2124a22d1207c485d31e54/GraphQlRepro/Program.cs#L69-L124

lbargaoanu:
I've looked a little deeper and is [used](https://github.com/AutoMapper/AutoMapper/commit/5b0cf17a3b19dcbd7ec214efb96455309c330643) to be the case that the list of projection mappers was checked on every projection request. If that helps, we'll consider bringing it back.

JanEggers:
for me it is not viable implementing a custom projection mapper. it is less code to just manually write the expression.

lbargaoanu:
Then that's the way to go.

lbargaoanu:
How would this work with our let clause? Examples: https://github.com/AutoMapper/AutoMapper/blob/master/src/IntegrationTests/CustomMapFrom/MapObjectPropertyFromSubQuery.cs and some and the ones using methods from https://github.com/AutoMapper/AutoMapper/blob/master/src/IntegrationTests/IncludeMembers.cs.

JanEggers:
I dont understand the question. what is a let clause? 

the tests are all green on my branch so I asume everything is working fine.

can you maybe create a test of what you dont expect to work?

JanEggers:
I understand the problem now. the let will introduce an intermediate projection that looses the type information. I will try to rewrite the projection in a way that preserves the type information.

JanEggers:
I updated all the tests in IncludeMembers to also use inhertance.

I also fixed the let clause be including the whole source in the intermediate projection so the type information is not lost.

lbargaoanu:
That defeats the very purpose of `ProjectTo`, fetching only the required data to fill the dto.

JanEggers:
> That defeats the very purpose of `ProjectTo`, fetching only the required data to fill the dto.

nope its just an intermidiate projection that is discarded. only the columns present in the final select are fetched.

lbargaoanu:
As far as I can tell, that's correct. Revert all changes to files other than ProjectionBuilder. Don't use `IncludeAllDerived`, explicitly include the required maps.
Add new tests (with `CreateMap`) for both files I mentioned that use inheritance. Add at least one an extra column to the derived classes and assert they have the expected data.
When that is done, I'll take a look at the tests to try and see if they work as expected.
After that we can discuss the implementation.

JanEggers:
> Revert all changes to files other than ProjectionBuilder

can you elaborate why? I was under the impression that if inheritance is supported by projection it should be available by AddProjection?

lbargaoanu:
I feel it's a niche use case and I believe the majority of people can safely ignore it exists. The rest can just use `CreateMap` or an `Advanced` extension method that casts `CreateProjection` to `IMappingExpression`.

lbargaoanu:
A thing I forgot to mention is that I have a branch ahead of master that surely conflicts with this one. I think we'll merge that one soon enough and you'll have to rebase. Not exactly major changes, but conflicts nevertheless.

lbargaoanu:
Constructor mapping has to work too. Some tests are [here](https://github.com/AutoMapper/AutoMapper/blob/master/src/UnitTests/Projection/ConstructorTests.cs).

JanEggers:
I updated the tests in [MapObjectPropertyFromSubQuery](https://github.com/AutoMapper/AutoMapper/blob/master/src/IntegrationTests/CustomMapFrom/MapObjectPropertyFromSubQuery.cs) and did the rebase

two of the tests are failing because there are now multiple FirstOrDefault calls. I updated the code to reuse the baseclass expressions for the derived classes but the issue persists. any advice whats wrong or how to fix it is very welcome.

also i noticed when mixing addprojection and addmapping I get an exception while projecting. I dont know if this is by design.

I will check / extend the constructor tests

src/IntegrationTests/IncludeMembers.cs
```diff
@@ -920,22 +1536,45 @@ public class Destination
...
         }
     }
     protected override MapperConfiguration CreateConfiguration() => new(cfg =>
     {
-        cfg.CreateProjection<Source, Destination>().IncludeMembers(s => s.InnerSources.FirstOrDefault(), s => s.OtherInnerSources.FirstOrDefault());
+        cfg.CreateMap<Source, Destination>().IncludeMembers(s => s.InnerSources.FirstOrDefault(), s => s.OtherInnerSources.FirstOrDefault())
+            .Include<SourceA, DestinationA>()
+            .Include<SourceB, DestinationB>();
+        cfg.CreateMap<SourceA, DestinationA>();
+        cfg.CreateMap<SourceB, DestinationB>();
```
lbargaoanu, line 1577: None of these derived maps actually have included members. I want to see at least two such cases, one for a simple included member, chained properties, and one that uses the let clause, such as this one, with a method. 
Check [the execution plan](https://docs.automapper.org/en/latest/Understanding-your-mapping.html). That extension is very useful :)

lbargaoanu:
Yes, don't mix `CreateMap` with `CreateProjection`. About the failing test, no hurry, it can wait. We'll figure it out eventually.

src/IntegrationTests/CustomMapFrom/MapObjectPropertyFromSubQuery.cs
```diff
@@ -689,27 +1007,43 @@ public class DatabaseInitializer : DropCreateDatabaseAlways<ClientContext>
...
 {
     protected override MapperConfiguration CreateConfiguration() => new(cfg =>
     {
-        cfg.CreateProjection<ProductArticle, ProductArticleModel>();
-        cfg.CreateProjection<Product, ProductModel>()
+        cfg.CreateMap<ProductArticle, ProductArticleModel>()
+            .Include<ProductArticleA, ProductArticleAModel>()
+            .Include<ProductArticleB, ProductArticleBModel>();
+        cfg.CreateMap<ProductArticleA, ProductArticleAModel>();
+        cfg.CreateMap<ProductArticleB, ProductArticleBModel>();
```
lbargaoanu, line 1041: Similar to `IncludeMembers`, at least a few of these tests should have some configuration in the derived map. Certainly a subquery using the let clause and perhaps some random `MapFrom` to see that it's used.

src/IntegrationTests/IncludeMembers.cs
lbargaoanu: Unless I've missed smth, you didn't address this and a similar comment for the subquery test.

JanEggers: yes will do today. was busy with the family over the weekend :)

JanEggers: I extended the tests. basicaly doing the same thing as for the base class in the derived class. that uncovered some problems with the current approach. I was able to fix some of the issues but there is at least one more issues remaining.  im not quite certain whats the issue as the expression it self looks ok but the entity framework transformation of the query cannot be translated to sql because it contains a a lamda that should probably be a property expression.

JanEggers: my current thinking is that it would be better to have multiple intermediate let statements that would also improve expression caching.

e.g:

instead of:
```c#
.Select(s => new {
    Source = s,
    Inner = s.InnerSources.FirstOrDefault(),
    Inner#Details = s.InnerSources.FirstOrDefault().InnerSourceDetails.FirstOrDefault().
})
```
generate:
```c#
.Select(s => new {
    Source = s,
    Inner = s.InnerSources.FirstOrDefault()
}).Select(s => new {
    s.Source,
    s.Inner,
    Details = s.Inner.InnerSourceDetails.FirstOrDefault()
})
```

lbargaoanu: One let clause should be enough for everyone :) It's not clear to me why EF would not handle the first expression. Except that it's too complicated. And if that's the case, just simplify your test so things don't compose. But of course I still want to see a subquery with a method in a derived class.

JanEggers: im investigating

src/IntegrationTests/IncludeMembers.cs
```diff
@@ -1630,4 +1630,105 @@ public void Should_flatten()
             result.ShouldBeOfType<Destination>();
         }
     }
+}
+public class IncludeOnlySelectedMembers : IntegrationTest<IncludeOnlySelectedMembers.DatabaseInitializer>
```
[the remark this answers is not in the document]
lbargaoanu: This needs to use an expression with `FirstOrDefault` to use the let clause.

src/UnitTests/Projection/ConstructorTests.cs
```diff
@@ -65,51 +124,175 @@ class Source
...
+            .Include<SourceWrapperB, DestinationB>();
+        cfg.CreateMap<SourceWrapperA, DestinationA>()
+            .IncludeMembers(s => s.A);
+        cfg.CreateMap<SourceWrapperB, DestinationB>()
+            .IncludeMembers(s => s.B);
+        cfg.CreateMap<Source, Destination>();
+        cfg.CreateMap<Source, DestinationA>();
+        cfg.CreateMap<Source, DestinationB>();
+        cfg.CreateMap<SourceA, DestinationA>();
+        cfg.CreateMap<SourceB, DestinationB>();
```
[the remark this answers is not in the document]
JanEggers: does nor work either. for some reason it converts SourceA and SourceB toString instead of using the Properties on the objects to call the constructor

JanEggers:
finally found a solution for the tests failing because ef was not not able to translate expressions.
the issue was caused by incorrect chaining of conditional source expressions. I am not 100% certain its the correct way to handle it as it is quite complicated to wrap your head around the expression trees....

still two failing tests remaining

src/AutoMapper/QueryableExtensions/ProjectionBuilder.cs
```diff
@@ -243,16 +297,43 @@ public override QueryExpressions GetSubQueryExpression(ProjectionBuilder builder
...
+                        source = Condition(condition.Test, letMapInfo.LetExpression.ReplaceParameters(condition.IfTrue), letMapInfo.Marker);
+                    }
+                    else if (extractor.Parameter.Type != type && source.Type != type)
+                    {
+                        source = Condition(TypeIs(source, type), letMapInfo.LetExpression.ReplaceParameters(Convert(source, type)), letMapInfo.Marker);
+                        for (int j = i+1; j < letMapInfos.Length; j++)
+                        {
+                            if (letMapInfos[j].MapFromSource.Type == letMapInfo.Marker.Type)
+                            {
+                                letMapInfos[j] = letMapInfos[j] with { MapFromSource = source };
```
lbargaoanu, line 322: The chaining should happen in `GetSourceExpression`, not here.

JanEggers: I changed it on purpose as a resultig expression of one let property can be a source of another property. im not sure how to  code this recursion when the code is in  GetSourceExpression

JanEggers: understood

JanEggers: I was able to simplify code and reuse it some more let me know what you think...

src/UnitTests/Projection/ConstructorTests.cs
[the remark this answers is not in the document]
JanEggers: changed member names now this test is green

src/UnitTests/Projection/ConstructorTests.cs
```diff
@@ -43,20 +63,59 @@ public DestinationItem(DestinationValue destinationValue)
...
     {
-        cfg.CreateProjection<Source, Destination>().ForCtorParam("item", o => o.MapFrom(s => s.Items.FirstOrDefault()));
-        cfg.CreateProjection<SourceItem, DestinationItem>().ForCtorParam("destinationValue", o=>o.MapFrom(s=>s.Values.FirstOrDefault()));
-        cfg.CreateProjection<SourceValue, DestinationValue>();
+        cfg.CreateMap<Source, Destination>()
+            .ForCtorParam("item", o => o.MapFrom(s => s.Items.FirstOrDefault()))
+            .Include<SourceA, DestinationA>()
+            .Include<SourceB, DestinationB>();
+        cfg.CreateMap<SourceA, DestinationA>()
+            .ForCtorParam("item", o => o.MapFrom(s => s.Items.FirstOrDefault())) //TODO: required for CTOR Validation but should not be required and messes up mapping
```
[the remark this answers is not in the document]
JanEggers: I changed the let property names to include the source types, this way there is no issue in having the same properties for different source types. the issue was that there have been 3 item properties due to the redundand constructor mappings.

JanEggers:
tests are green now on my system let me know if I should add more tests change something in the implementation

lbargaoanu:
I see

  Message: 
System.NullReferenceException : Object reference not set to an instance of an object.

  Stack Trace: 
ConstructorLetClause.Should_construct_correctly() line 111

src/UnitTests/Projection/ConstructorTests.cs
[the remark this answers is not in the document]
lbargaoanu: That's arbitrary. It should work in the same way as it does for properties. By building a path within the query that says how we got to that point.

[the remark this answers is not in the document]
JanEggers: the problem is that it is an expression with a parameter. the new code checks the type of the parameter and creates a convert expression to the type of the parameter. and that results in 3 different source expressions that access the same property.

if the 3 properties should be treated as the same property I need to change the code to not look at the type of the parameter but look at the declaring type of the first memberexpression. 

is this a better solution?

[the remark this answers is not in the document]
lbargaoanu: That cast is not needed in this case. Maybe try to have it only when it's actually needed, perhaps when accessing a member of a derived type.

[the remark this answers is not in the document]
JanEggers: done

src/IntegrationTests/IncludeMembers.cs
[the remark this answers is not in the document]
JanEggers: I changed the test like you requested locally but it is broken, although not a regression. the FirstOrDefault let expression brings in the whole object instead of the requested properties. 

I am investigating if this can be fixed by adding a projection that only includes the required properties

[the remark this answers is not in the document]
JanEggers: found a solution

src/AutoMapper/QueryableExtensions/ProjectionBuilder.cs
```diff
@@ -165,11 +165,10 @@ Expression MapFromExpression(LambdaExpression mapFrom)
                     {
                         if (memberTypeMap == null || mapFrom.IsMemberPath(out _) || mapFrom.Body is ParameterExpression)
                         {
-                            var parmaterExtractor = new ParameterExpressionExtractorVisitor();
-                            parmaterExtractor.Visit(mapFrom.Body);
-                            if (parmaterExtractor.Parameter != null)
+                            var member = MemberAccessExpressionExtractorVisitor.Extract(mapFrom.Body, mapFrom.Parameters.First());
+                            if (member != null)
```
lbargaoanu, line 169: What does this actually mean? Why would accessing smth on the parameter matter?

src/IntegrationTests/IncludeMembers.cs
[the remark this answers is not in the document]
lbargaoanu: But that just brings back the one piece of code you removed.

src/AutoMapper/QueryableExtensions/ProjectionBuilder.cs
JanEggers: the code needs to decide weather a type conversion is required.

before I used the parameter type to detect this.

but that resulted in unnessesay conversions in case the parameter is of the derived type but the used property is defined in the base class.  

the new code only adds a conversion if the member of the parameter is actually declared on the derived type

src/IntegrationTests/IncludeMembers.cs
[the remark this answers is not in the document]
JanEggers: the original code was not preserving type information. 
if you have suggestions on what code should be restored please let me know. then i will try to use that code instead.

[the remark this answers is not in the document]
JanEggers: also i tried to repro on master but strangely there the query is correct even if the let clause contains the whole object included by firstOrDefault. im still investigating whats the difference there...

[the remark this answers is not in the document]
JanEggers: I was able to repro the issue.
as soon as the let statement contains a conditional statement entity framework selects all the properties instead of the required properties.

https://github.com/JanEggers/AutoMapper/tree/queryContainsMoreThanRequiredProperties

src/AutoMapper/QueryableExtensions/ProjectionBuilder.cs
JanEggers: Ahh I was not thinking about using multiple members in one expression tree I will update the code

JanEggers: done. i now extract all member expressions on the given parameter and use the most derived type to determine weather a conversion is required.

lbargaoanu:
I think it's fair to say we're not trending in the right direction :)
I think we should simply not mix inheritance with the let clause. So have everything inline in a single `Select`.
Just leave this branch for future reference and start over. Hopefully the new approach can work with less complexity.

JanEggers:
> So have everything inline in a single Select

I tried that suggestion but removing the let clause creates much bigger queries as the subqueries are now longer reused.

to clarify did you mean i should remove the let clause completely or just avoid using it when dealing with inheritance?

lbargaoanu:
Just avoid using it with inherintance, as is the case now for chained properties expressions.
And yes, the resulting queries won't look good, but at least the code generating them would be manageable.

PR #4273, opened 2023-04-04
Hotfix/fix problem with struct mapping

Maximys:
After reading [this](https://github.com/AutoMapper/AutoMapper/discussions/4199) discussion, I had try to make mapping from DateTime to DateOnly and I had get exception.
Current pr fix this problem.


lbargaoanu:
Let's start over.
Open an issue with a repro as requested, with a database, and we'll take it from there.

PR #4278, opened 2023-04-08
Add the DI code

lbargaoanu:
I've pushed the 12.0.1 DI update. Maybe archive that repo?


lbargaoanu:
@jbogard Can we merge this one? Because right now I don't think you can use the MyGet build with DI.

jbogard (approved):

jbogard:
Old repo is now archived

PR #4279, opened 2023-04-14
Runtime polymorphic projections no let

JanEggers:
same as #4272 but avoiding let when using inheritance.

from my pov the complexity is on par with #4272 it just moved from chaining sources to deciding wether or not to use let clause


lbargaoanu:
If there are included types you, don't use the let clause. It's simple.

lbargaoanu:
I don't see why existing tests would need to change. Just add new tests.

PR #4292, opened 2023-04-21
Fix ReverseMap destination validation issue

### LINKED ISSUE #4291: ReverseMap validation doesn't work since 12.0.0
We used to use `CreateMap<>(MemberList.Destination).ReverseMap().ValidateMemberList(MemberList.Destination));` to make sure that the objects are identical.
This technique was working in 11.0.1 (and it is mentioned several times here and on SO), but it stopped working in 12.0.0. The Reverse validation doesn't seem to occur.

### Source/destination types

```csharp
    public class Source
    {
        public int SomeValue  { get; set; }
        public int SomeValue2 { get; set; }
    }

    public class Destination
    {
        public int SomeValue  { get; set; }
    }
```

### Mapping configuration

```csharp
        var configuration = new MapperConfiguration(cfg =>
                                    cfg.CreateMap<Source, Destination>(MemberList.Destination)
                                       .ReverseMap()
                                       .ValidateMemberList(MemberList.Destination));
```

### Expected behavior

AssertConfigurationIsValid() should throw, as `SomeValue2` is missing from the reverse mapping `<Destination, Source>` and we ask to validate the destination.
And it throws in 11.0.1

### Actual behavior

AssertConfigurationIsValid() doesn't throw.

killergege:
Fixes #4291 

As mentioned in the issue, I'm not sure why this check was added to the property, but removing it fixes that specific problem.


killergege:
@dotnet-policy-service agree

src/AutoMapper/TypeMap.cs
```diff
@@ -76,14 +76,7 @@ public static Exception MissingMapException(Type sourceType, Type destinationTyp
     public MemberList ConfiguredMemberList
     {
         get => (_details?.ConfiguredMemberList).GetValueOrDefault();
-        set
-        {
-            if (value == default)
```
lbargaoanu, line 81: The check is here to avoid creating `_details` in the default case for `CreateMap`. So we want that _and_ your use case to work. Somehow :)

src/UnitTests/Bug/ReverseMapDestinationMembersValidation.cs
```diff
@@ -0,0 +1,33 @@
+using Xunit;
+
+namespace AutoMapper.UnitTests.Bug
+{
+    namespace AddingConfigurationForNonMatchingDestinationMember
+    {
+        public class ReverseMapDestinationMembersValidationBug : NonValidatingSpecBase
```
lbargaoanu, line 7: Let's move this in ConfigurationValidation.cs.

src/UnitTests/Bug/ReverseMapDestinationMembersValidation.cs
```diff
@@ -0,0 +1,33 @@
...
+            protected override MapperConfiguration CreateConfiguration() => new(cfg =>
+            {
+                cfg.CreateMap<Source, Destination>(MemberList.Destination).ReverseMap()
+                    .ValidateMemberList(MemberList.Destination);
+            });
+
+            [Fact]
+            public void Should_show_configuration_error()
+            {
+                typeof(AutoMapperConfigurationException).ShouldBeThrownBy(AssertConfigurationIsValid);
```
lbargaoanu, line 29: We should also check that the initial map has no `_details`. That probably requires a new `TypeMap.HasDetails`.

lbargaoanu:
Try the [MyGet](https://docs.automapper.org/en/latest/The-MyGet-build.html) build.

PR #4304, opened 2023-06-14
Support mapping of polymorphic collection with abstract parent type

kvpt:
When mapping a polymorphic list with an abstract parent, the expression throw an exception on create.
It try to use a new expression of the abstract type which is not possible.
So replace it with a default value expression which is allowed and not throw an error.


src/UnitTests/Projection/ProjectAbstractWithInheritance.cs
lbargaoanu, line 1: This needs to be an integration test, running with a DB, [here](https://github.com/AutoMapper/AutoMapper/tree/master/src/IntegrationTests/Inheritance).

src/AutoMapper/QueryableExtensions/ProjectionBuilder.cs
```diff
@@ -64,7 +64,10 @@ public QueryExpressions CreateProjection(in ProjectionRequest request, LetProper
     QueryExpressions CreateProjection(in ProjectionRequest request, LetPropertyMaps letPropertyMaps, TypeMap typeMap, TypeMap[] polymorphicMaps)
     {
         var instanceParameter = Parameter(request.SourceType, "dto" + request.SourceType.Name);
-        var projection = CreateProjectionCore(request, instanceParameter, typeMap, letPropertyMaps);
+        var projection = !typeMap.DestinationType.IsAbstract
```
lbargaoanu, line 67: A better fix is to simply skip abstract types when computing polymorphic maps.

kvpt: I disagree with that, or perhaps I don't see how to do it this way at this level.
Because the abstract type mapping don't come from the polymorphic map but from the EnumerableProjectionMapper.Project method.
This is because it is this abstract type that is used for the collection property in the entity.

lbargaoanu: I guess both make sense, projecting to an abstract type when there are included maps and also including an abstract map, presumably for configuration reuse.

PR #4307, opened 2023-06-23
Use `TryAdd` in `OpenTypeMapConfigs`

tiasmt:
Summary of the changes
 - Use `TryAdd` when adding to `OpenTypeMapConfigs`, to prevent exception being thrown when not calling `AssertConfigurationIsValid`.


lbargaoanu:
Let's start over. Open an issue with a repro.

PR #4320, opened 2023-07-19
Add integration test case

kvpt:
Add an integration test case reproducing the issue with polymorphic element projection.

### DIFF
--- a/src/IntegrationTests/Inheritance/ProjectToAbstractTypeWithInheritance.cs
+++ b/src/IntegrationTests/Inheritance/ProjectToAbstractTypeWithInheritance.cs
@@ -14,11 +14,18 @@ public class ProjectToAbstractTypeWithInheritance : IntegrationTest<ProjectToAbs
         public string Name { get; set; }
         public int StepGroupId { get; set; }
         public virtual StepGroup StepGroup { get; set; }
+        public virtual ICollection<StepInput> StepInputs { get; set; } = new HashSet<StepInput>();
     }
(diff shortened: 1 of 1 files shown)


src/IntegrationTests/Inheritance/ProjectToAbstractTypeWithInheritance.cs
```diff
@@ -89,6 +113,14 @@ protected override void Seed(Context context)
                 }
             });
 
+            context.SaveChanges();
```
lbargaoanu, line 116: I don't think this is needed.

src/IntegrationTests/Inheritance/ProjectToAbstractTypeWithInheritance.cs
```diff
@@ -98,7 +130,15 @@ public void ProjectCollectionWithElementInheritingAbstractClass()
     {
         using var context = new Context();
         var steps = ProjectTo<StepGroupModel>(context.StepGroups).Single().Steps;
-        steps[0].ShouldBeOfType<CheckingStepModel>().Name.ShouldBe("CheckingStep");
-        steps[1].ShouldBeOfType<InstructionStepModel>().Name.ShouldBe("InstructionStep");
+        steps[0].ShouldBeOfType<CheckingStepModel>();
```
lbargaoanu, line 133: Revert this and modify the other test in the same direction.

lbargaoanu (changes requested):

PR #4327, opened 2023-08-03
add options to disable and override ExplicitExpansion

Angelinsky7:
replace #4326
Pr to implement https://github.com/AutoMapper/AutoMapper/discussions/4325
from discussion from here : https://github.com/AutoMapper/AutoMapper/discussions/4323


Angelinsky7:
@lbargaoanu i think i corrected almost everything. (apart from the override parameter that i really would like to talk about)

src/IntegrationTests/ExplicitExpansion/DisableExplicitExpansion.cs
```diff
@@ -0,0 +1,111 @@
...
+    public class Context : LocalDbContext {
+        public DbSet<Entity> Entities { get; set; }
+    }
+    public class DatabaseInitializer : DropCreateDatabaseAlways<Context> {
+        protected override void Seed(Context context) {
+            context.Entities.Add(new() { Name = "Name" });
+            base.Seed(context);
+        }
+    }
+    protected override MapperConfiguration CreateConfiguration() => new(c => c.CreateProjection<Entity, Dto>().ForCtorParam("Name", o => o.ExplicitExpansion(false)));
```
lbargaoanu, line 102: This should use mapping inheritance, the same as the test below.

Angelinsky7: @lbargaoanu i added a new test for constructor mapping with inheritance (sorry for the long delay i needed to change my computer)

Angelinsky7:
if you do that, you loose all the advantage of doing this pr !
i don't understand why you would want to make all user need to change all their code by needing to pass a value (it breaks compatibility)

lbargaoanu:
Obviosly it doesn't, all the existing tests pass.

Angelinsky7:
@lbargaoanu for the matter of this, i'm against this pr in this state...

Angelinsky7:
@lbargaoanu yeah of course you removed all my tests :-)

lbargaoanu:
Well, that's what's needed here. The `ForAllMembers` stuff is `ForAllOtherMembers` in disguise, nothing that makes sense for AM. If you want that, just add an `if` as I've already said.

Angelinsky7:
i'm sorry but you didn't understand my usecase and why i was doing this pr... and your solution doesn't work.

Angelinsky7:
but again, don't bother

PR #4328, opened 2023-08-04
Private methods that don't access any instance members were made `static`

Lehonti:


lbargaoanu:
Kind of irelevant.

Lehonti:
Not irrelevant. You can use that to determine which methods you can safely refactor out of the class, into utility classes, for example.

lbargaoanu:
That's not the case here. These methods are static by accident.

PR #4341, opened 2023-09-01
add missing creation of mapper local var in example

jamesmanning:
made the change through the github we ui, I guess it added an EOL or something for the last line to change.


lbargaoanu:
It;s not relevent where the mapper is coming from.

PR #4387, opened 2024-02-07
Adding annotations based on MS guidance

jbogard:
Reviewed the changes and added tests basically to confirm the behavior based on this guidance:

https://github.com/dotnet/runtime/blob/main/docs/coding-guidelines/api-guidelines/nullability.md

This should cover the 99.99% scenarios. People can break it intentionally but this at least shows the default behaviors?

### DIFF
--- a/src/AutoMapper/Mapper.cs
+++ b/src/AutoMapper/Mapper.cs
@@ -1,3 +1,5 @@
+using System.Diagnostics.CodeAnalysis;
+
 namespace AutoMapper;
 using IObjectMappingOperationOptions = IMappingOperationOptions<object, object>;
 using Factory = Func<Type, object>;
@@ -11,7 +13,7 @@ public interface IMapperBase
     /// <typeparam name="TDestination">Destination type to create</typeparam>
     /// <param name="source">Source object to map from</param>
     /// <returns>Mapped destination object</returns>
-    TDestination? Map<TDestination>(object? source);
+    TDestination Map<TDestination>(object source);
     /// <summary>
     /// Execute a mapping from the source object to a new destination object.
     /// </summary>
@@ -19,6 +21,7 @@ public interface IMapperBase
     /// <typeparam name="TDestination">Destination type to create</typeparam>
     /// <param name="source">Source object to map from</param>
     /// <returns>Mapped destination object</returns>
+    [return: NotNullIfNotNull(nameof(source))]
     TDestination? Map<TSource, TDestination>(TSource? source);
     /// <summary>
     /// Execute a mapping from the source object to the existing destination object.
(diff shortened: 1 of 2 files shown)


lbargaoanu:
I'm sure there are people out there that have maps that return null for non-null source. You'll just lie to them? :) Does MS do that anywhere? I've checked System.Text.Json and it does have nullable both on the source and the result.

src/AutoMapper/Mapper.cs
```diff
@@ -155,8 +161,18 @@ public Mapper(IConfigurationProvider configuration, Factory serviceCtor)
     ResolutionContext IInternalRuntimeMapper.DefaultContext => DefaultContext;
     Factory IInternalRuntimeMapper.ServiceCtor => _serviceCtor;
     public IConfigurationProvider ConfigurationProvider => _configuration;
-    public TDestination Map<TDestination>(object source) => Map(source, default(TDestination));
-    public TDestination Map<TDestination>(object source, Action<IMappingOperationOptions<object, TDestination>> opts) => Map(source, default, opts);
+    public TDestination Map<TDestination>(object source)
+    {
+        ArgumentNullException.ThrowIfNull(source);
```
lbargaoanu, line 166: Really? :)

EklipZgit:
> I'm sure there are people out there that have maps that return null for non-null source. You'll just lie to them? :) Does MS do that anywhere? I've checked System.Text.Json and it does have nullable both on the source and the result.

https://github.com/dotnet/runtime/blob/main/docs/coding-guidelines/api-guidelines/nullability.md#:~:text=How%20common%20is%20it%20in%20the%20case%20of%20(3)%20for%20such%20invocations%20to%20then%20dereference%20the%20result%20rather%20than%20passing%20it%20off%20to%20something%20else%20that%20accepts%20a%20T%3F%3F

The intended behavior is not to map objects to nulls. In the wild, I've never seen it done. In cases where it is done, it is YOU doing it, and not someone outside of your control, so YOU know you need to handle nulls. The library should not make EVERYONE handle nulls when they know there are no nulls.

I think this PR is perfectly reasonable, thanks @jbogard

lbargaoanu:
This is not good :) If you don't want the headache, let's just revert the whole thing. We can simply be nullable oblivious as before.

src/AutoMapper/Mapper.cs
jbogard: Ha yes. We should be checking for nulls here when we rely on a not-null source object for getting the type.

jbogard:
Then I'd just vote revert. I just tried this on a codebase and had similar headaches - it's not adding anything meaningful in the vast majority of cases. 

This captures the vast majority of cases that I can think of, as well as our expected built-in behavior. You can get around all the compile-time null checks anyway but this is about describing expected behaviors. But it's not 100% accurate.

lbargaoanu:
It's not worth the trouble. We can simply revert and let people write their own stuff if they must.

PR #4402, opened 2024-02-17
Fix polymorphic mapping when some derived types have explicit mappings and others do not

kev-andrews:
as discussed in #4396

### DIFF
--- a/src/AutoMapper/QueryableExtensions/ProjectionBuilder.cs
+++ b/src/AutoMapper/QueryableExtensions/ProjectionBuilder.cs
@@ -77,7 +77,7 @@ public sealed class ProjectionBuilder : IProjectionBuilder
         {
             var sourceType = derivedMap.SourceType;
             var derivedRequest = request.InnerRequest(sourceType, derivedMap.DestinationType);
-            var derivedProjection = CreateProjectionCore(derivedRequest, letPropertyMaps, derivedMap, Convert(source, sourceType));
+            var derivedProjection = CreateProjectionCore(derivedRequest, letPropertyMaps, derivedMap, TypeAs(source, sourceType));
             projection = Condition(TypeIs(source, sourceType), derivedProjection, projection, projection.Type);
         }
         return projection;
--- a/src/IntegrationTests/Inheritance/PolymorphismTests.cs
+++ b/src/IntegrationTests/Inheritance/PolymorphismTests.cs
@@ -1,0 +1,104 @@
+﻿namespace AutoMapper.IntegrationTests.Inheritance;
+
+public class PolymorphismTests : IntegrationTest<PolymorphismTests.DatabaseInitializer>
+{
+    public abstract class Vehicle
+    {
+        public int Id { get; set; }
+        public string Name { get; set; }
+    }
+
+    public class Car : Vehicle
+    {
+        public int AmountDoors { get; set; }
+    }
+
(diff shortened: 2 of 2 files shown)


kev-andrews:
@dotnet-policy-service agree

src/UnitTests/Projection/PolymorphismTests.cs
```diff
@@ -0,0 +1,63 @@
+﻿namespace AutoMapper.UnitTests.Projection;
+
+public class InMemoryMapPolymorphic : AutoMapperSpecBase
```
lbargaoanu, line 3: This needs to be an integration test in the dedicated folder.

src/AutoMapper/QueryableExtensions/ProjectionBuilder.cs
```diff
@@ -76,6 +76,10 @@ Expression CreateProjection(in ProjectionRequest request, LetPropertyMaps letPro
         foreach(var derivedMap in polymorphicMaps)
         {
             var sourceType = derivedMap.SourceType;
+            if (!request.SourceType.IsAssignableFrom(sourceType))
+            {
+                continue;
+            }
             var derivedRequest = request.InnerRequest(sourceType, derivedMap.DestinationType);
             var derivedProjection = CreateProjectionCore(derivedRequest, letPropertyMaps, derivedMap, Convert(source, sourceType));
```
lbargaoanu, line 84: I think the only change needed is to replace `Convert` with `TypeAs`.

lbargaoanu (changes requested):

src/UnitTests/Projection/PolymorphismTests.cs
kev-andrews: moved

src/AutoMapper/QueryableExtensions/ProjectionBuilder.cs
kev-andrews: you are correct, there is a difference in the generated Expressions and SQL though:

my initial solution does not add polymorphic maps that are not applicable to the expression tree:

```
DbSet<Car>()
    .Select(dtoCar => new VehicleModel{ Name = dtoCar.Name }
    )
    
SELECT [v].[Name]
FROM [Vehicles] AS [v]
WHERE [v].[Discriminator] = N'Car'
```

the approach of adding them with the type check results in:

```
DbSet<Car>()
    .Select(dtoCar => (dtoCar is Bicycle) ? new BicycleModel{ 
        EBike = (dtoCar as Bicycle).EBike, 
        Name = (dtoCar as Bicycle).Name 
    }
     : (dtoCar is Motorcycle) ? new MotorcycleModel{ 
        HasSidecar = (dtoCar as Motorcycle).HasSidecar, 
        Name = (dtoCar as Motorcycle).Name 
    }
     : new VehicleModel{ Name = dtoCar.Name }
    )

SELECT [v].[Id], [v].[Discriminator], [v].[Name], [v].[AmountDoors]
FROM [Vehicles] AS [v]
WHERE [v].[Discriminator] = N'Car'
```

if this is not an issue i can commit the version you proposed with less changes to the codebase

kev-andrews: ok, i cleaned up the commits to reflect the requested changes

lbargaoanu (approved):

PR #4403, opened 2024-02-17
fix: error in Build.ps1 - the wrong target to the copy compat version

dadhi:
I was not able to `.\Build.ps1` the project after cloning until I fixed the path.


dadhi:
@dotnet-policy-service agree

Issue #4412, opened 2024-03-10
Maps containing generics not being added to configuration using MapperConfiguration(config => c.AddMaps(assembly))


JerrettDavis:
### Source/destination types and Mapping configuration

```csharp
public class MyClass
{
    public int Id { get; set; }
    public string Name { get; set; } = null!;
}

public record MyClassDto 
{
    public int Id { get; init; }
    public string Name { get; init; } = null!;
    
    private class Mapping : Profile
    {
        public Mapping() => CreateMap<MyClass, MyClassDto>();
    }
}

public record MyGenericClassDto<T>
{
    public T Id { get; init; } = default!;
    public string Name { get; init; } = null!;
    
    private class Mapping : Profile
    {
        public Mapping() => CreateMap<MyClass, MyGenericClassDto<int>>();
    }
}
```

### Expected behavior

Expected `new MapperConfiguration(c => c.AddMaps(assembly))` to add **all** maps defined in `Profile` classes in an assembly to a given configuration.

### Actual behavior

All Maps, except those with Generics seem to be added.

### Steps to reproduce

Create a NUnit test project. reference a project implementing the above demo classes (`MyClass`, `MyClassDto`, `MyGenericClassDto`). Use the below code to execute 4 test runs. 3 of the 4 runs will succeed, while all 4 should.

We can see that the configuration explicitly creating the generic map succeeds in the second test, but it's not discovered in the assembly scan during the first test.

```csharp
public class MappingTests
{
    [Test]
    // Will succeed
    [TestCase(typeof(MyClass), typeof(MyClassDto))]
    // Will fail
    [TestCase(typeof(MyClass), typeof(MyGenericClassDto<int>))]
    public void ShouldSupportMappingFromSourceToDestination_AddMaps(
        Type source, 
        Type destination)
    {
        var instance = GetInstanceOf(source);
        var configuration = new MapperConfiguration(config => 
            config.AddMaps(Assembly.GetAssembly(typeof(MyClass))));
        var mapper = configuration.CreateMapper();

        mapper.Map(instance, source, destination);
    }
    
    [Test]
    // Will succeed
    [TestCase(typeof(MyClass), typeof(MyClassDto))]
    // Will succeed
    [TestCase(typeof(MyClass), typeof(MyGenericClassDto<int>))]
    public void ShouldSupportMappingFromSourceToDestination_CreateMap(
        Type source, 
        Type destination)
    {
        var instance = GetInstanceOf(source);
[...]

PR #4414, opened 2024-03-12
Add generic support to assembly scanning for nested profiles

JerrettDavis:
Added the ability for assembly scanning to add profiles nested in a generic class.

#4412


JerrettDavis:
@dotnet-policy-service agree

xeekworx (approved):

lbargaoanu:
That's just weird and doesn't add anything useful. Your code is the problem, not AM.

xeekworx:
> That's just weird and doesn't add anything useful. Your code is the problem, not AM.

I am also interested in this change. As far as I can tell AM doesn't have support for generic mappings. Does AM provide this support already or are you unwilling to add support for it (without explanation). Why was the related issue closed?

PR #4420, opened 2024-03-19
Docs/fix-links

alexravenna:
It looked like all of the relative links in the documentation were not functioning because the Markdown files were referencing `.html` pages instead of `.md` ones. This PR fixes that. Also, 1 or 2 of the links to sub-sections were incorrect, so I fixed those too.


lbargaoanu:
Go to the docs site and you'll see that the links work.

alexravenna:
@lbargaoanu: What do you think made me start this PR in the first place? I was using the docs site and was confused that none of the links to other docs pages worked.

Try it for yourself:
[https://docs.automapper.org/en/stable/Getting-started.html#Dependency-injection.html#asp-net-core](Getting Started)

On that page alone there are 4 links that don't work.

jbogard:
No ideas, I just went through the upgrade guide from ReadTheDocs. I didn't do the original conversion either so I'm not entirely sure how it all works.

alexravenna:
[FluentValidation](https://github.com/FluentValidation/FluentValidation/tree/main) seems to be using RTD v2 with Markdown files and doesn't use file endings for local/relative links at all.

Example: https://github.com/FluentValidation/FluentValidation/blob/4df7f57d58cf1ecac5abbbff0c60cf0c81c75b11/docs/start.md?plain=1#L131

Maybe that's the solution? 😃

lbargaoanu:
[But they do](https://github.com/FluentValidation/FluentValidation/blob/4df7f57d58cf1ecac5abbbff0c60cf0c81c75b11/docs/start.md?plain=1#L142). It must be smth in conf.py. Maybe that recommonmark thing. We used to have smth like that, but it's gone now :)

lbargaoanu:
I guess myst replaces recommonmark.

PR #4423, opened 2024-03-22
Set html_baseurl

lbargaoanu:
Maybe give this a try.
Another idea to try is setting [default_role](https://www.sphinx-doc.org/en/master/usage/referencing.html#cross-referencing-anything) to `any` or `doc`.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -26,7 +26,7 @@ intersphinx_disabled_domains = ['std']
 templates_path = ['_templates']
 
 # -- Options for HTML output
-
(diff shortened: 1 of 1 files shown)


jbogard:
Sure, let's give it a try.

lbargaoanu:
I can't hurt :) But my guess is that, in the end, the problem is https://github.com/AutoMapper/AutoMapper/pull/4420#issuecomment-2019526931.

lbargaoanu:
Maybe just revert the recent docs changes?

jbogard:
I dunno the build was failing and it was preventing any releases.

PR #4425, opened 2024-04-01
Readability improvements

nimanikoo:
Hello, I hope you are well
I wanted to thank you for contributing to open source and building this project for the .NET community.
I was curious to review the code in this repository and found some points that might be useful to improve.
I hope these points are helpful and I would be happy to contribute to the progress of this project.
In each commit and change I made, I referenced the Microsoft documentation as well as the reason for the change and refactoring.
If you agree, we can also upgrade to .NET 8 and C# 12 in the next step. I think this will be beneficial for the project in terms of performance.

Please review and let me know the results.


nimanikoo:
@dotnet-policy-service agree

lbargaoanu:
There is little value here.

nimanikoo:
The way you code and your informal behavior shows how much you value it.
I'm really sorry for you.

@lbargaoanu

PR #4432, opened 2024-04-11
Allow package version ranges

dariogriffo:
One good practice that will allow projects referencing different versions of MinVer/options to be resolved during build without conflicts is to use version ranges
https://blog.inedo.com/nuget/package-versioning/
MinVer not allowed beyond 5 
rest is safe to have 9 as upper bound.
I couldn't run integration tests, since the the debian is not supported as platform for `LocalDb`


lbargaoanu:
We're fine with using open ranges.

PR #4446, opened 2024-04-30
WIP: Use FastExpressionCompiler for compilation and diagnostics

dadhi:
Hi @jbogard,

Time goes, and I was able to improve the [FastExpressionCompiler](https://github.com/dadhi/FastExpressionCompiler) enough to pass AutoMapper unit tests. 

This PR is just a demonstration. The minimal FEC files are included directly.

Today, in addition to compilation speed and some invocation speed, FEC has capabilities to inspect and output Expressions and C# as demonstrated in `MapperConfiguration`.

The sources of FEC used here correspond to [FEC v4.2.0](https://github.com/dadhi/FastExpressionCompiler/releases/tag/v4.2.0)

If it is an interest to you, I may add `MapperConfiguration` hook, so the consumer may switch to the FEC if he wants to.

Anyway, it was fun for me to make things work :-)
Cheers!


src/AutoMapper/AutoMapper.csproj
```diff
@@ -53,5 +53,6 @@
     <Using Include="AutoMapper.Execution.ExpressionBuilder" Static="true" />
     <Using Include="AutoMapper.Execution" />
     <Using Include="AutoMapper.Configuration" />
+    <Using Include="FastExpressionCompiler" />
```
lbargaoanu, line 56: In my view, the best solution would be just a hook to allow changing the compiler. So AM doesn't need to reference FEC. We would advertise the change in the upgrade guide and we could link to a unit test in this repo that shows it in action. Perhaps an extension method to `IMapperConfiguration` to simplify the usage.

dadhi: Hi @lbargaoanu 

Yeah, I was thinking about something like moving the `CompileFast` to the delegate.
It may look something like this:

```cs
new MapperConfiguration(cfg => cfg.UseCompiler(expr => expr.CompileFast()));
```

I am not familiar with the details of AM configuration, but it would be good to set `Compiler` globally, e.g. statically. 
IMHO, this option does not require the granularity anyway.

This way, I may just add another UnitTests project, bring the FEC dependency into it, reference the source UnitTests and configure the FEC statically.  Then run all the tests for the regression.

lbargaoanu: Yes, smth like that. I don't know about running the tests with FEC here, in this repo. What would we do when smth breaks, disable the offending test?

dadhi: @lbargaoanu Yes, Ignore and issue in my repo. This will do.

lbargaoanu: I think we should start with a small PR that just introduces the hook in the [internal API](https://github.com/AutoMapper/AutoMapper/blob/1fbdf63dd99bff60b8fea096897ce0ee921f54f0/src/AutoMapper/Internal/InternalApi.cs#L48). The test can simply use the built in compiler and assert that it's called. We can add FEC afterwards.

jbogard: Yeah I agree, let's start small. Unfortunately I think the only way to "know" that FEC works is to run against all the tests.

lbargaoanu: Yes, but I think it's kind of weird to have the tests here when the fix is in another repo :)

dadhi: I may run the tests on my side as well via devcontainer or similar, via checking out AutoMapper, copying my own project into it, referencing AM tests, configuring AM with FEC and running the tests. Likely, it will work too.

lbargaoanu: If it helps, we could have a hook in our tests to allow changing the compiler when running the tests, maybe with an [environment variable](https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-test#synopsis). Just a thought, maybe there are better ways.

PR #4448, opened 2024-05-06
Update Enum-Mapping.md to describe how to map both string and int value of Enums at the same time without any additional configuration

twojnarowski:
- Add description on mapping Enum.ToString()
- Fix some gramaticall erros

### DIFF
--- a/docs/source/Enum-Mapping.md
+++ b/docs/source/Enum-Mapping.md
@@ -1,14 +1,14 @@
 # AutoMapper.Extensions.EnumMapping
 
-The built-in enum mapper is not configurable, it can only be replaced. Alternatively, AutoMapper supports convention based mapping of enum values in a separate package [AutoMapper.Extensions.EnumMapping](https://www.nuget.org/packages/AutoMapper.Extensions.EnumMapping/).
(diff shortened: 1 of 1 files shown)


twojnarowski:
@dotnet-policy-service agree

docs/source/Enum-Mapping.md
```diff
@@ -41,15 +41,43 @@ internal class YourProfile : Profile
     ...
 ```
 
+### Mapping Enum's names
```
lbargaoanu, line 44: That's too specific to be useful. Generally speaking you don't need the `ToString` suffix, you just need a `string` property and perhaps a `MapFrom`. We don't need this paragraph.

twojnarowski: `you just need a string property` is not correct in the case when I need to map both to int value and string value, because I can't have `string Value` and `int Value` properties in one object.

lbargaoanu: Again, not useful.

twojnarowski: Useful to me, I needed that, and I couldn't find anything about that in the docs and ha to figure it out by random trial and error...
Here https://docs.automapper.org/en/stable/Queryable-Extensions.html#string-conversion it is mentioned, that ToString() method is automatically added to to each non-string to string mapping. But nowhere it is said, that I can write ToString in a property name, to trigger this method.

PR #4455, opened 2024-05-27
Fix duplicate base mapping when inherited and IncludeAllDerived used

nicolea55:
Overcoming the "You cannot include a type map into itself" in case of mapping hierarchies


nicolea55:
@dotnet-policy-service agree

lbargaoanu:
A repro would help. Make a [gist](https://gist.github.com/lbargaoanu/9c7233441c3a3413cc2b9b9ebb5964a9) that we can execute and see fail.

PR #4513, opened 2024-12-20
Optimize Features.cs

KhanbalaRashidov:
Adjustments have been made in the code and repeated operations have been made into methods


lbargaoanu:
I feel the code is more readable the way it is.

PR #4526, opened 2025-02-08
Don't throw and catch on validation

### LINKED ISSUE #4523: Not all mapping errors are surfaced when a map contains more than one sort of misconfiguration
When a map contains unmapped properties _and_ properties with incompatible types, only the unmapped member error is thrown during validation. Once the unmapped member error is fixed, then the "unmappable member" error is thrown.

`AssertConfigurationIsValid` should surface all mapping errors upfront, to prevent the need for multiple passes when fixing configuration errors.

## Example

```csharp
class Foo
{
    public DateOnly IncompatibleProp { get; set; }
}

class FooDto
{
    public DateTime IncompatibleProp { get; set; }
   public string UnmappedProp { get; set; }
}

// ...

var configuration = new MapperConfiguration(cfg =>
{
    cfg.CreateMap<Foo, FooDto>();
});

```

This will throw this error:

```log
Unmapped members were found. Review the types and members below
...
Unmapped properties:
UnmappedProp
```

When the unmapped member is mapped, the code will then start throwing this error:

```log
The following member on AutomaticInterfaceExample.FooDto cannot be mapped: 
    IncompatibleProp 
```

I think it would be very helpful if the configuration exception showed both issues at once. I believe this could be done relatively easily by lifting the existing `configExceptions` collection higher in the `AssertConfigurationIsValid` method and having it collect type map errors together with dry run errors.

https://github.com/AutoMapper/AutoMapper/blob/b521096a3e05782df85fdb42356601a9bb3f35f8/src/AutoMapper/Configuration/ConfigurationValidator.cs#L28-L63

lbargaoanu:
Fixes #4523.


lbargaoanu:
@jbogard It's been a while, so I think we can release AM 14.

src/UnitTests/ConfigurationValidation.cs
```diff
@@ -1,5 +1,40 @@
...
+                        ex => ex.Errors.ShouldBeOfLength(1),
+                        ex => ex.Errors[0].UnmappedPropertyNames.ShouldContain("Bar")),
+                aex => aex.InnerExceptions[1]
+                    .ShouldBeOfType<AutoMapperConfigurationException>()
+                    .ShouldSatisfyAllConditions(
+                        ex => ex.MemberMap.ShouldNotBeNull(),
+                        ex => ex.MemberMap.DestinationName.ShouldBe("Foo"))
+            );
+    }
+}
```
simonmckenzie, line 37: Because the validation is no longer stopped by a `throw` at the first error, failures on multiple child properties are now surfaced. I think another test would be helpful here:

_I have verified that this test fails on the trunk but passes on this branch_

```suggestion
}

public class When_testing_a_dto_with_mismatches_in_multiple_children : AutoMapperSpecBase
{
    public class Source
    {
        public Type Foo { get; set; }
        public Type Bar { get; set; }
    }

    public class Destination
    {
        public int Foo { get; set; }
        public int Bar { get; set; }
    }

    protected override MapperConfiguration CreateConfiguration() =>
        new(cfg => { cfg.CreateMap<Source, Destination>(); });

    [Fact]
    public void Should_throw_for_both_mismatched_children()
    {
        new Action(AssertConfigurationIsValid)
            .ShouldThrow<AggregateException>()
            .ShouldSatisfyAllConditions(
                aex => aex.InnerExceptions.ShouldBeOfLength(2),
                aex => aex.InnerExceptions[0]
                    .ShouldBeOfType<AutoMapperConfigurationException>()
                    .ShouldSatisfyAllConditions(
                        ex => ex.MemberMap.ShouldNotBeNull(),
                        ex => ex.MemberMap.DestinationName.ShouldBe("Foo")),
                aex => aex.InnerExceptions[1]
                    .ShouldBeOfType<AutoMapperConfigurationException>()
                    .ShouldSatisfyAllConditions(
                        ex => ex.MemberMap.ShouldNotBeNull(),
                        ex => ex.MemberMap.DestinationName.ShouldBe("Bar"))
            );
    }
}

```

jbogard (approved):

PR #4632, opened 2026-04-17
Support LicenseKey via environment variable fallback

Nit8:
# PR: Auto-Detect License Key from Environment Variables Fixes

## What Changed
Added support for reading the AutoMapper license key from an environment variable instead of requiring manual configuration in code.

## Why
In containerized and cloud environments, it's better to use environment variables for sensitive configuration instead of hardcoding values in code.

## How It Works

The LicenseKey property now checks for values in this order:
1. Explicit value set in code (cfg.LicenseKey = "key") - highest priority
2. Environment variable AUTOMAPPER_LICENSE_KEY - fallback
3. Null if neither is set - lowest priority

## Usage

NEW WAY - Using Environment Variable:
Set environment variable AUTOMAPPER_LICENSE_KEY before running your app
Then just add AutoMapper without setting the license key:

    services.AddAutoMapper(cfg => 
    {
        cfg.CreateMap<Foo, FooDto>();
    });

OLD WAY - Still Works (Backward Compatible):

    services.AddAutoMapper(cfg => 
    {
        cfg.LicenseKey = "your-license-key";
        cfg.CreateMap<Foo, FooDto>();
    });

BOTH SET - Code value wins:
If you set both the environment variable AND code value, the code value is used.

## Tests Added
- Tests for environment variable auto-detection
- Tests for old explicit assignment (backward compatibility)
- Tests for precedence when both are set
- Integration tests with dependency injection

## Backward Compatibility
100% backward compatible. All existing code continues to work exactly the same way.

### DIFF
--- a/src/AutoMapper/Configuration/MapperConfigurationExpression.cs
+++ b/src/AutoMapper/Configuration/MapperConfigurationExpression.cs
@@ -127,7 +127,13 @@ public sealed class MapperConfigurationExpression : Profile, IGlobalConfiguratio
     /// </summary>
     int IGlobalConfigurationExpression.RecursiveQueriesMaxDepth { get; set; }
 
-    public string LicenseKey { get; set; }
+    private string _licenseKey;
+
+    public string LicenseKey
+    {
+        get => _licenseKey ?? Environment.GetEnvironmentVariable("AUTOMAPPER_LICENSE_KEY");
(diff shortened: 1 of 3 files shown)


Nit8:
Will be pushing out a new fork pull request with better test cases and clean commit history.

PR #4634, opened 2026-04-25
Support LicenseKey via environment variable fallback

### LINKED ISSUE #4631: Support automatic discovery of license keys
Today AutoMapper requires manual setting of the license key:

```csharp
services.AddAutoMapper(cfg => {
    cfg.LicenseKey = "License key here";

    // Other configuration
});
```

While this allows for unlimited options for retrieving the license key (typically via the .NET configuration system), it still requires manual setting. This can be problematic for larger enterprises that share the license key amongst many systems/applications. This also requires code changes to set the license key.

Instead, we can support automatic discovery of license key, via:
- .NET configuration system (via a well-known key)
- Environment variables
- File system?

This should support registration via `IServiceCollection` and manual creation.

This may simplify some overloads back to releases prior to 15.x since a license key could be dynamically discovered instead of required in the configuration system.

Nit8:
# PR: Auto-Detect License Key from Environment Variables Fixes #4631 

## What Changed
Added support for reading the AutoMapper license key from an environment variable instead of requiring manual configuration in code.

## Why
In containerized and cloud environments, it's better to use environment variables for sensitive configuration instead of hardcoding values in code.

## How It Works

The LicenseKey property now checks for values in this order:
1. Explicit value set in code (cfg.LicenseKey = "key") - highest priority
2. Environment variable AUTOMAPPER_LICENSE_KEY - fallback
3. Null if neither is set - lowest priority

## Usage

NEW WAY - Using Environment Variable:
Set environment variable AUTOMAPPER_LICENSE_KEY before running your app
Then just add AutoMapper without setting the license key:

    services.AddAutoMapper(cfg => 
    {
        cfg.CreateMap<Foo, FooDto>();
    });

OLD WAY - Still Works (Backward Compatible):

    services.AddAutoMapper(cfg => 
    {
        cfg.LicenseKey = "your-license-key";
        cfg.CreateMap<Foo, FooDto>();
    });

BOTH SET - Code value wins:
If you set both the environment variable AND code value, the code value is used.

## Tests Added
- Tests for environment variable auto-detection
- Tests for old explicit assignment (backward compatibility)
- Tests for precedence when both are set
- Integration tests with dependency injection

## Backward Compatibility
100% backward compatible. All existing code continues to work exactly the same way.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,6 +39,10 @@ jobs:
       shell: pwsh
     - name: Report Test Results
       uses: dorny/test-reporter@v1
+      # Publishing the report needs checks:write, which GitHub forces to read-only
+      # for pull_request runs from forks. Don't let that fail the build; the test
+      # step itself already reflects pass/fail.
+      continue-on-error: true
       if: success() || failure()
       with:
         name: Test Results (Linux)
@@ -74,6 +78,9 @@ jobs:
         shell: pwsh
       - name: Report Test Results
         uses: dorny/test-reporter@v1
+        # See note on the Linux job: report publishing is best-effort and must
+        # never fail the build (forked-PR tokens are read-only).
+        continue-on-error: true
         if: success() || failure()
         with:
           name: Test Results (Windows)
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,6 +59,9 @@ jobs:
         with:
           fetch-depth: 0
       - name: Azure Login via OIDC
+        # Only needed to sign and publish packages on main. Skipped on PRs,
+        # where the Azure secrets aren't available (and are empty on forks).
+        if: github.ref == 'refs/heads/main'
         uses: azure/login@v2
         with:
           client-id: ${{ secrets.AZURE_CLIENT_ID }}
@@ -87,6 +90,8 @@ jobs:
           path: artifacts/**/*.trx
           reporter: dotnet-trx
       - name: Sign packages
+        # Signing uses the Azure Key Vault login above; only runs on main.
+        if: github.ref == 'refs/heads/main'
         run: |-
           foreach ($f in Get-ChildItem "./artifacts" -Filter "*.nupkg") {
             NuGetKeyVaultSignTool sign $f.FullName --file-digest sha256 --timestamp-rfc3161 http://timestamp.digicert.com --azure-key-vault-managed-identity --azure-key-vault-url ${{ secrets.AZURE_KEYVAULT_URI }} --azure-key-vault-certificate ${{ secrets.CODESIGN_CERT_NAME }}
--- a/docs/source/License-configuration.md
+++ b/docs/source/License-configuration.md
@@ -18,6 +18,21 @@ var mapperConfiguration = new MapperConfiguration(cfg => {
 
 You can obtain a valid license from the [AutoMapper website](https://automapper.io).
 
+### Auto-Discovery via Environment Variables
+
+If no license key is set in code, AutoMapper looks for one in environment variables. This is convenient for containerized and cloud environments, and for enterprises that share a single key across many services without code changes:
+
+- `AUTOMAPPER_LICENSE_KEY` – the AutoMapper-specific license key.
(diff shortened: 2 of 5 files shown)


Nit8:
@jbogard can you run this pipeline to check if anything goes wrong

Nit8:
Hi @jbogard, is this issue related to pipeline, am I missing something?

jbogard:
Hmm that is odd, I tried to re-run it but still getting that error.

Nit8:
yeah @jbogard thats strange does it have something to do with ci pipeline and about read and write permissions?

Nit8:
Hi @jbogard  I see you changed the permission into your release.yml and maybe same has to be done from ci.yml from read to write for this to complete as mentioned in my previous comment, and you can try firing the pipeline to verify. Thanks

SubashPaudyalTR:
This would be a great relief if we have this directly fed from the configuration, and introduce less noise ... as we are having more than dozens of services which is using automapper and we just acquired the enterprise license. We are looking for best option weather to go with per service configuration v/s one-place shared library. Both holds its own pros and cons, but if we get this env. var auto-discovery, it would be the real winner for us... with minimal changes. 

Waiting for this and any other valuable suggestions!!!

jbogard:
Hey all, I want this to land in the next release. I need this to also work with MediatR so I'm going to suggest some changes that allow it to work in environments that use a single Bundle key across both products in a single application.

jbogard:
Thanks for this, @Nit8 — really appreciate the contribution, and it pushed exactly the right feature forward. 🙏 I've pushed a few commits on top of your branch to get it merge-ready, and wanted to walk through what changed and why.

**What I kept:** your core idea — falling back to an environment variable when no key is set in code, fully backward compatible — is exactly right, and the precedence model (explicit code value wins) is preserved.

**What I changed:**

1. **Moved the discovery out of the property getter into `LicenseAccessor`.** Reading `Environment.GetEnvironmentVariable(...)` inside the `LicenseKey` getter made the property side-effecting and non-deterministic (it hit the environment on every access). I reverted `MapperConfigurationExpression.LicenseKey` back to a plain auto-property and centralized resolution in `LicenseAccessor.ResolveLicenseKey(...)` — the single place that actually consumes the key. This also gives us a clean seam for the IConfiguration / file-based discovery that #4631 also asks for.

2. **Added a second, shared `LUCKYPENNY_LICENSE_KEY` variable.** Lucky Penny licenses are sold as a bundle covering both AutoMapper and MediatR, so a company-wide variable lets both libraries read the same key. Resolution precedence is now:
   - `cfg.LicenseKey` (explicit code) →
   - `AUTOMAPPER_LICENSE_KEY` (product-specific) →
   - `LUCKYPENNY_LICENSE_KEY` (shared)

   Product-specific wins over the shared key, mirroring how Duende IdentityServer layers a product key over a company-wide one. (Note: the shared key still has to be for a license that includes AutoMapper — a MediatR-only license won't validate here.)

3. **Reworked the tests** to target `ResolveLicenseKey` directly for the precedence rules, and disabled parallelization on that class since it mutates process-global environment variables.

4. **Documented** the new auto-discovery section in `License-configuration.md`.

The env var names follow the flat all-caps convention used by other commercial .NET libraries (Particular's `PARTICULARSOFTWARE_LICENSE`, Telerik's `TELERIK_LICENSE`, etc.), since AutoMapper reads them directly rather than through IConfiguration binding.

Thanks again — your authorship on the PR stands, this just builds on it.

Nit8 (commented):
Thanks so much for taking the time to build on this rather than just closing it really means a lot.
The refactor to LicenseAccessor.ResolveLicenseKey() makes complete sense in hindsight; I hadn't considered the side-effect problem with hitting the environment on every property access.
Happy with everything you've done on top, and glad the core idea was worth building on. Looking forward to seeing it merged this is a library we use in our own project, and having run into this issue ourselves, it felt worth solving for everyone's benefit.
 Looking forward to contributing more to this amazing project!

jbogard:
Ported to MediatR in LuckyPennySoftware/MediatR#1175 (using `MEDIATR_LICENSE_KEY` as the product-specific variable, with the same shared `LUCKYPENNY_LICENSE_KEY` fallback).

Issue #4530, opened 2025-02-18
Cannot ignore source null values, if destination property is not nullable


utillity:
I use AutoMapper to ensure all properties are accounted for between Request, Database, and DTO (outgoing data-transfer-objects).

I have request objects coming via the API, to patch an object in the Database.

````c#
	[Fact]
	public void MapRequestToRecord()
	{
		var configuration = new MapperConfiguration(cfg =>
		{
			cfg.AllowNullDestinationValues = true;
			cfg.CreateMap<SomeRequest, SomeRecord>()
				// cannot access property-info here to check for nullability
				.ForAllMembers(opt => opt.Condition((s, t, srcV, destV, ctx) => srcV != null));
		});
		configuration.AssertConfigurationIsValid();
		var mapper = configuration.CreateMapper();

		mapper.Map<SomeRecord>(new SomeRequest()).Count.Should().Be(2);
		mapper.Map<SomeRecord>(new SomeRequest { Count = 3 }).Count.Should().Be(3);
	}

	private class SomeRequest
	{
		public int? Count { get; set; }
	}

	private class SomeRecord
	{
		public int Count { get; set; } = 2;
	}
````

The srcV is converted to 0, which is understandable. But in the Condition method I would like to know which property I'm on (PropertyInfo), so I can at least use reflection to check if the original source-value was null.

Is it possible to add this (either as param or to the ResolutionContext)?

BTW: tested with AutoMapper 14.0.0

lbargaoanu:
I think this is better suited for StackOverflow.

jbogard:
Reopening to track this — it is a real bug, not just nullable semantics. The third `Condition` argument is documented as the source member but we were passing the value after conversion to the destination member type, so a null `Nullable<T>` had already collapsed to `default(T)` before the condition ran. Fix in #4655.

jbogard:
Fixed by #4655, merged as 6e8697bc — shipping in 16.3.

`Condition` now receives the value resolved from the source object before conversion to the destination member type when the condition's member parameter can hold it, so a `ForAllMembers` condition sees a null `Nullable<T>` instead of `default(T)`. `Conditional-mapping.md` gained a "Nullable source members" section covering that, the `PreCondition` route for typed `ForMember` conditions, `UseDestinationValue` for the keep-the-destination case, and the `?? default(T)` rule when there is no condition.

Issue #3926, opened 2022-03-04
ForAllMembers expression Condition incorrectly passes Nullable<T> source member as T


xnoreq:
### Source/destination types

```csharp
    public class Destination
    {
        public bool Toggle { get; set; }
    }

    public class Source
    {
        public bool? Toggle { get; set; }
    }
```

### Mapping configuration

```csharp
    public class TestProfile : Profile
    {
        public TestProfile()
        {
            AllowNullCollections = true;
            AllowNullDestinationValues = true;

            CreateMap<Source, Destination>()
                .ForAllMembers(expression => expression.Condition((src, dst, srcMember) => srcMember != null));
        }
    }
```

### Version: 11.0.1
11.0.1

### Expected behavior

In the ForAllMembers condition the `srcMember` should be of type `bool?` (for the `Source.Toggle` property) and the condition should evaluate to `false` (given `Toggle = null` as in the example program below). The destination thing should therefore stay unmodified.

### Actual behavior
The `srcMember` is of type `bool`, so the information that the input was `null` **gets lost.** The ForAllMembers condition evaluates to `true` changing the destination thingy.

### Steps to reproduce

```csharp
    public class Program
    {
        static void Main(string[] args)
        {
            var config = new MapperConfiguration(cfg => cfg.AddProfile<TestProfile>());
            var mapper = config.CreateMapper();

            var thing = new Destination()
            {
                Toggle = true
            };

            var update = new Source()
            {
                Toggle = null
            };

            Console.WriteLine(thing.Toggle); // True
            mapper.Map(update, thing);
            Console.WriteLine(thing.Toggle); // False
        }
    }
```

jbogard:
Reopening to track this — it is a real bug, not just nullable semantics. The third `Condition` argument is documented as the source member but we were passing the value after conversion to the destination member type, so a null `Nullable<T>` had already collapsed to `default(T)` before the condition ran. Fix in #4655.

jbogard:
Fixed by #4655, merged as 6e8697bc — shipping in 16.3.

`Condition` now receives the value resolved from the source object before conversion to the destination member type when the condition's member parameter can hold it, so a `ForAllMembers` condition sees a null `Nullable<T>` instead of `default(T)`. `Conditional-mapping.md` gained a "Nullable source members" section covering that, the `PreCondition` route for typed `ForMember` conditions, `UseDestinationValue` for the keep-the-destination case, and the `?? default(T)` rule when there is no condition.

Issue #2999, opened 2019-02-24
Condition when mapping from Nullable<T> to T


MarkSFrancis:
When using a condition to check the value of a member of type Nullable<T> before mapping to a destination member with a type T, the value is converted to T before Condition is checked, and so checking for null will not work correctly.

### Source/destination types

```csharp
class Source
{
    public int? Value { get; set; }
}

class Destination
{
    public int Value { get; set; }
}
```

### Mapping configuration

```csharp
cfg.CreateMap<Source, Destination>().ForAllMembers(o => o.Condition((source, destination, sourceProperty, destinationProperty) =>
{
    sourceProperty.ShouldBeNull(); // Fails as it's already converted int? to int before it reaches this, and so the value is 0
    
    // Always true
    return sourceProperty != null;
}));
```

### Version: x.y.z
8.1.0

### Expected behavior

Type of `sourceProperty` should be `Nullable<Int32>`

### Actual behavior

Type of `sourceProperty` is `Int32`

### Steps to reproduce

```csharp
public class when_mapping_from_nullable_clr_member : AutoMapperSpecBase
{
    Source _source = new Source { Value = null };
    Destination _destination = new Destination { Value = 7 };

    class Source
    {
        public int? Value { get; set; }
    }

    class Destination
    {
        public int Value { get; set; }
    }

    protected override MapperConfiguration Configuration => new MapperConfiguration(cfg =>
    {
        cfg.CreateMap<Source, Destination>().ForAllMembers(o => o.Condition((source, destination, sourceProperty, destinationProperty) =>
        {
            sourceProperty.ShouldBeNull();

            // Always true
            return sourceProperty != null;
        }));
    });

    protected override void Because_of()
    {
        Mapper.Map(_source, _destination);
    }
}
```

lbargaoanu:
Yes, and there are known workarounds. You'll have to use whatever works best in your case.

MarkSFrancis:
@lbargaoanu Can you link some of these workarounds? I can't find anything in the docs mentioning them, and the most obvious issue I could find was #2918, which just says that documentation is needed (but again - still doesn't say what those workarounds are). Although this issue is closed, I can't find the related docs. 
Sorry if I'm just being block-headed and unable to see something obvious :/

MarkSFrancis:
On another note, is this something that I could help resolve? I'm happy to add the test for it and then try and fix this issue, submitting a PR if I'm successful. 
[...]

Issue #2918, opened 2018-12-16
Nullable source property to value type


jhenderson2099:
### Documentation Request
Please document how the (default) value for a destination property that doesn't exist in the source object is determined.  

**This is being logged at the request of @jbogard .**   Please refer to [https://groups.google.com/forum/#!topic/automapper-users/UX-liQGboMY](https://groups.google.com/forum/#!topic/automapper-users/UX-liQGboMY)

(Basically, "This is how it works if you works if the mapping is/isn't explicit/implicit and what you get as a default)

Suggestion is document it as part of _Inline Mapping_ 
[https://automapper.readthedocs.io/en/latest/Inline-Mapping.html]
(https://automapper.readthedocs.io/en/latest/Inline-Mapping.html)

### Example Source/destination types

```csharp
    public class Source
    {
        public double? Balance { get; set; }

    }


    public class Destination
    {
        public double Balance { get; set; }   // Not nullable!
    }
```

### Mapping configuration
```csharp
CreateMap<Source, Destination>(); // Implicit property-to-property mapping
```

or

```csharp
// Explicit property-to-property mapping
CreateMap<Source, Destination>()
    .ForMember(dest => dest.Balance, opt => opt.MapFrom(src => src.Balance))
```

### Version: x.y.z
7.0.0

### Questions to be answered
- If Source.Balance == null, then what will be the value of Destination.Balance?
- Does AutoMapper convert the null to zero, or does the instance of Destination simply get initialized to 0 and no mapping occurs?
- What would happen in this scenario if the two property names did not match and there was no explicit mapping?  What would be the value of Destination.Balance?

I think it works like this: if you have a nullable source property and the value of the source property is `null`and a non-nullable destination property (based upon my testing):
- If you have a mapping (explicit or implicit), then you get the default value for that type (e.g., `default(int)`)
- If you don't supply a mapping, and there is no implicit automapping that will work, and you **don't** supply a value in the Destination class's constructor, then you get the default value for that type
- If you don't supply a mapping, and there is no implicit automapping that will work, and you **do** supply a value in the Destination class's constructor, then you get the value that you specified in the constructor

### Steps to reproduce

N/A

lbargaoanu:
But all of this is quite unsurprising. So what's really the point of documenting it? Surely shorter documentation is better. It's easy to find out what happens at runtime using the execution plan. And that's also an unnecessary constraint for us in the future. We might need to change some details. It doesn't seem like a good idea to me.

[...]

PR #4655, opened 2026-09-04
Pass the source member value to Condition

### LINKED ISSUE #4627: Automapper keeps closing the issues related to Nullable source to Non-Nullable destination mapping
This is a problem which has been present from 2018 and still hasn't been solved, which is very useful for the PATCH scenarios.  
I won't write steps to reproduce because you can see them in the linked issues:

https://github.com/LuckyPennySoftware/AutoMapper/issues/2918
https://github.com/LuckyPennySoftware/AutoMapper/issues/2999
https://github.com/LuckyPennySoftware/AutoMapper/issues/3926
https://github.com/LuckyPennySoftware/AutoMapper/issues/4530

Amd you keep closing all similar issues, when there is obviously a need to support this behaviour.  
It's even missing in the documentation so it has to be discovered each time by a new individual.

jbogard:
Closes #4627. Also addresses #2999, #3926 and #4530.

## The bug

`CreatePropertyMapFunc` passed `mappedMemberVariable` — the value *after* conversion to the destination member type — as the third argument to `Condition`. That argument is documented as the source member (`IMemberConfigurationExpression.Condition`: "against the source, destination, **source** and destination members"), and the fourth argument is already the destination getter.

For a `Nullable<T>` source member and a non-nullable `T` destination member, null had already collapsed to `default(T)` by the time the condition ran. `ForAllMembers` types the member parameter as `object`, so the condition boxed the `0`/`false` rather than receiving a null — which makes the "only assign what the caller supplied" PATCH scenario unexpressible. That's what's been reported repeatedly since 2018.

Confirmed on `main`, source `{ Count = null }` mapped over a destination with `Count = 7`:

| config | before | after |
|---|---|---|
| `ForAllMembers(o => o.Condition((s,d,sm) => sm != null))` | **0** | **7** |
| `ForMember(d => d.Count, o => o.PreCondition(s => s.Count != null))` | 7 | 7 |

## The fix

Use the source-typed `resolvedValueVariable` when the condition's member parameter type is assignable from it, and fall back to the mapped value otherwise:

```csharp
var conditionMemberType = memberMap.Condition.Parameters[2].Type;
var conditionSourceMember = conditionMemberType.IsAssignableFrom(resolvedValueVariable.Type)
    ? (Expression)resolvedValueVariable
    : mappedMemberVariable;
```

The fallback is load-bearing. A `ForMember` condition types both member parameters as the *destination* member type (`TMember`), so passing an `int?` there would emit `Convert(int?, int)` and throw `InvalidOperationException` on null. Those conditions keep today's behavior, and `When_using_a_condition_typed_as_the_destination_member` guards it.

## Behavior change

For `ForAllMembers` (or any condition whose member parameter is `object` or a base type) where the source and destination member types differ, the condition now receives the source member instead of the mapped destination member. `When_using_a_condition_for_all_members_with_different_source_and_destination_member_types` pins this. It matches the documented contract, but it is observable. Targeting 16.3.

No public API surface change, so no ApiCompat baseline update.

## Not changed

`dest.Int = src.NullableInt ?? default(int)` when there is no condition. A name match always produces an assignment and a non-nullable destination member cannot represent absence — that part of #2918 is inherent, and is now written down in the docs instead.

## Docs

New "Nullable source members" section in `Conditional-mapping.md` covering the `ForAllMembers` pattern, why `ForMember` conditions can't see null and should use `PreCondition`, and the `?? default(T)` rule with a pointer to null substitution. The documentation gap is arguably the bigger half of #4627 — five issues over eight years never surfaced the `PreCondition` workaround. The blanket-PATCH workaround (`cfg.Internal().ForAllPropertyMaps(...)` closing over `pm.SourceMember`) is deliberately not documented here; whether it should be promoted out of `AutoMapper.Internal` is #4656.

## Testing

Full unit suite passes (1226). The three new behavior tests fail without the `TypeMapPlanBuilder` change and pass with it. Integration tests were not run locally (they need SQL Server via Testcontainers) — leaving those to CI.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Wm2m61jXrAs4FheXR5TufH

### DIFF
--- a/docs/source/Conditional-mapping.md
+++ b/docs/source/Conditional-mapping.md
@@ -23,6 +23,46 @@ var configuration = new MapperConfiguration(cfg => {
 ```
 If you have a resolver, see [here](Custom-value-resolvers.html#resolvers-and-conditions) for a concrete example.
 
+## Nullable source members
+
+The source member value passed to `Condition` is the value resolved from the source object, *before* it is converted to the destination member type. This matters when the source member is a `Nullable<T>` and the destination member is a non-nullable `T`: the condition sees `null`, not `default(T)`.
+
+This makes the common PATCH scenario -- "only assign members the caller actually supplied" -- expressible with `ForAllMembers`:
+
+```c#
+class Source {
+  public int? Count { get; set; }
+}
+
+class Destination {
+  public int Count { get; set; }
+}
+
+var configuration = new MapperConfiguration(cfg => {
+  cfg.CreateMap<Source, Destination>()
+    .ForAllMembers(opt => opt.Condition((src, dest, srcMember) => srcMember != null));
+}, loggerFactory);
+
+var destination = new Destination { Count = 7 };
+mapper.Map(new Source { Count = null }, destination);
+// destination.Count is still 7 -- the null source member was skipped
+```
+
+Note that `ForAllMembers` types the source member parameter as `object`, which is what allows a `Nullable<T>` to arrive intact. A `ForMember` condition types both member parameters as the *destination* member type, so a `Nullable<T>` source cannot be represented there and the condition receives the converted value instead (`0` for `int`, `false` for `bool`). To check a nullable source member for a single member, use a `PreCondition` against the source object:
+
+```c#
+cfg.CreateMap<Source, Destination>()
+  .ForMember(dest => dest.Count, opt => opt.PreCondition(src => src.Count != null));
+```
+
+Without a condition, a null `Nullable<T>` source member is always assigned as `default(T)`. AutoMapper does not decide to map zero -- a name match always produces an assignment, and a non-nullable destination member has no way to represent the absence of a value. Roughly:
+
+```c#
+dest.Count = src.Count ?? default(int);
+```
+
+If you need a value other than `default(T)`, see [Null substitution](Null-substitution.html).
+
 ## Preconditions
 
 Similarly, there is a PreCondition method. The difference is that it runs sooner in the mapping process, before the source value is resolved (think MapFrom). So the precondition is called, then we decide which will be the source of the mapping (resolving), then the condition is called and finally the destination value is assigned.
@@ -57,6 +97,8 @@ public interface ICondition<in TSource, in TDestination, in TMember>
     bool Evaluate(TSource source, TDestination destination, TMember sourceMember, TMember destMember, ResolutionContext context);
 }
 ```
+
+As with the lambda overloads, `sourceMember` is the value resolved from the source object before conversion to the destination member type -- see [Nullable source members](#nullable-source-members).
 
 `IPreCondition<TSource, TDestination>` is evaluated before source member resolution and does not have access to member values:
 
--- a/docs/source/Conditional-mapping.md
+++ b/docs/source/Conditional-mapping.md
@@ -55,6 +55,15 @@ cfg.CreateMap<Source, Destination>()
   .ForMember(dest => dest.Count, opt => opt.PreCondition(src => src.Count != null));
 ```
 
+For this particular shape -- keep whatever the destination already holds when the source member is null -- `UseDestinationValue` does the same job without a condition at all, and works for a `ForMember` too (where a condition can't see the null):
+
+```c#
+cfg.CreateMap<Source, Destination>()
+  .ForAllMembers(opt => opt.UseDestinationValue());
+```
+
+Reach for it when the members are scalars. It is not a general PATCH switch: on a member that is itself a mapped object, `UseDestinationValue` maps *into* the existing destination instance rather than replacing it, so the nested map still overwrites that object's own members with the source's (unset ones landing as defaults). Members you want left alone entirely still need a condition.
+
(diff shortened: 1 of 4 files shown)


docs/source/Conditional-mapping.md
```diff
@@ -23,6 +23,46 @@ var configuration = new MapperConfiguration(cfg => {
 ```
 If you have a resolver, see [here](Custom-value-resolvers.html#resolvers-and-conditions) for a concrete example.
 
+## Nullable source members
+
+The source member value passed to `Condition` is the value resolved from the source object, *before* it is converted to the destination member type. This matters when the source member is a `Nullable<T>` and the destination member is a non-nullable `T`: the condition sees `null`, not `default(T)`.
```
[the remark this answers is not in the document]
jbogard: Fixed in 34fbd20. Both spots now state the actual selection rule: the pre-conversion source value when the condition's member parameter can hold it, the converted value otherwise.

jbogard:
@codex review

docs/source/Conditional-mapping.md
```diff
@@ -23,6 +23,46 @@ var configuration = new MapperConfiguration(cfg => {
...
+```
+
+`ForAllMembers` types the source member parameter as `object`, which is what allows a `Nullable<T>` to arrive intact. A `ForMember` condition types both member parameters as the *destination* member type, so a `Nullable<T>` source cannot be represented there and the condition falls back to the converted value (`0` for `int`, `false` for `bool`). To check a nullable source member for a single member, use a `PreCondition` against the source object:
+
+```c#
+cfg.CreateMap<Source, Destination>()
+  .ForMember(dest => dest.Count, opt => opt.PreCondition(src => src.Count != null));
+```
+
+Without a condition, a null `Nullable<T>` source member is always assigned as `default(T)`. AutoMapper does not decide to map zero -- a name match always produces an assignment, and a non-nullable destination member has no way to represent the absence of a value. Roughly:
```
lbargaoanu, line 67: If you want to keep the destination value when mapping `null`, you can set [`UseDestinationValue`](https://github.com/LuckyPennySoftware/AutoMapper/blob/9b500d50afe327a66d9a0002208aee79b99651d5/src/UnitTests/Bug/UseDestinationValue.cs#L13), no need for conditions.

src/AutoMapper/Execution/TypeMapPlanBuilder.cs
```diff
@@ -454,9 +454,13 @@ private Expression CreatePropertyMapFunc(MemberMap memberMap, Expression destina
             : Assign(destinationMemberAccess, mappedMemberVariable);
         if (memberMap.Condition != null)
         {
+            var conditionMemberType = memberMap.Condition.Parameters[2].Type;
+            var conditionSourceMember = conditionMemberType.IsAssignableFrom(resolvedValueVariable.Type)
+                ? (Expression)resolvedValueVariable
```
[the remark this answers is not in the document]
jbogard: Confirmed and fixed in fec985c4. Reproduced it exactly as described: `CreateMap(typeof(Source), typeof(Destination)).ForMember(..., o => o.Condition(typeof(ICondition<Source, Destination, int>)))` over an `int?` source threw NullReferenceException unboxing the boxed null — passes on main, threw on the branch, so it was a regression this PR introduced.

Root cause is as you diagnosed: the non-generic `MemberConfigurationExpression` derives from `MemberConfigurationExpression<object, object, object>`, so `CreateConditionExpression` stamped the wrapper's member parameters as `object` and the plan builder's assignability check couldn't see the condition's real member type.

Took the first option — the wrapper for `Condition(Type)` is now built from the condition interface's own type arguments, so `Parameters[2].Type` is the actual member type. `ICondition<,,int>` now falls back to the mapped value (same as a typed `ForMember` condition), while `ICondition<,,object>` still sees the null. That also covers the InvalidCastException case you flagged for other differing types, since the check is now against the real type rather than `object`. Two tests added pinning both directions.

docs/source/Conditional-mapping.md
jbogard: You're right, and it's the better answer for this shape — I checked it against `int?` -> `int`: `ForAllMembers(o => o.UseDestinationValue())` keeps 7 for a null source and takes 3 when set, and it works on a single `ForMember` too, where a condition can't see the null at all. Documented it in fec985c4 ahead of the `PreCondition` advice.

One caveat I noted with it, so nobody reads it as a blanket PATCH switch: on a member that is itself a mapped object it maps *into* the existing instance rather than replacing it, so the nested map still overwrites that object's own members (a source with only `A` set left `B` at 0 on the preserved instance). Scalars are where it's a clean substitute for a condition.

lbargaoanu: I'm not sure what you mean :) If the missing value marker is not `null`, you need a condition, but other than that, it seems to me that `UseDestinationValue` works in all cases, inner maps included. Maybe I don't know what PATCH needs. A failing test?

